PR #4, opened 2015-07-14
Remove react from peer dependencies

ide:
Installing react-redux with React Native ends up installing web React because npm automatically installs peer dependencies (until npm 3.0 is here). Likewise it'd be noisy if react-native were listed under the peer dependencies when most people don't need it. One day this won't be an issue anymore React splits into three pieces -- react, react-dom, react-native w/react as an external dependency.

(I understand if you don't want to take this PR especially when React 0.14 is out and parent v owner context becomes an issue...)

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,7 +57,6 @@
     "invariant": "^2.0.0"
   },
   "peerDependencies": {
-    "redux": "^1.0.0 || 1.0.0-alpha || 1.0.0-rc",
-    "react": "^0.13.0 || 0.14.0-beta1"
+    "redux": "^1.0.0 || 1.0.0-alpha || 1.0.0-rc"
   }
 }


gaearon:
OK, that's sensible.
I'll add peerDep *after* 0.14 is released (because we'll depend on parent context & package split).

gaearon:
Out in 0.2.2

Tankpt:
Hi, @gaearon  In npm3  peerDependencies  will not install in node_modules.
And this will  not transform react in react-redux by browserify-shim.

gaearon:
Not sure what you are referring to. Peer dependency has since been added. It won't install React—this is up to you to depend on it, as your app should depend on React anyway. And I don't see how this is related to Browserify.

Tankpt:
Thanks for your reply.

I try to package my code and external React etcs. And every thing was ok except the react-redux. 

browserify-shim Issues is https://github.com/thlorenz/browserify-shim/issues/209.

gaearon:
Thanks, this clarifies it. Likely an issue on their side.
Why do you use externals for React and Redux though? They are compatible with CommonJS just fine. And you can use UMD builds if you really need them as separate files.

PR #13, opened 2015-07-29
Fix tests

nd0ut:


gaearon:
From my understanding, you're porting the tests from Redux 1.0 alpha to 1.0 RC?
Let's hold this off for a while because there's already work on tweaking the API itself.

Issue #1, opened 2015-07-11
Alternative API proposals


gaearon:
This API is taken from Redux `<Provider>`, `<Connector>`, `@provide`, `@connect`.
Right now I don't have examples in this repo but it matches what you've seen in Redux so far.

It's the best we've got now, but I think we can do better!

Common pain points:

* Not intuitive how way to separate smart and dumb components with `<Connector>`, `@connect`
* You have to manually bind action creators with `bindActionCreators` helper which [some don't like](https://github.com/gaearon/redux/pull/86)
* Too much nesting for small examples (`<Provider>`, `<Connector>` both need function children)

Let's go wild here. Post your alternative API suggestions.

They should satisfy the following criteria:

* Some component at the root must hold the `store` instance. (Akin to `<Provider>`)
* It should be possible to connect to state no matter how deep in the tree
* It should be possible to select the state you're interested in with a `select` function
* Smart / dumb components separation needs to be encouraged
* There should be one obvious way to separate smart / dumb components
* It should be obvious how to turn your functions into action creators
* Smart components should probably be able to react to updates to the state in `componentDidUpdate`
* Smart components' `select` function needs to be able to take their props into account
* Smart component should be able to do something before/after dumb component dispatches an action
* We should have `shouldComponentUpdate` wherever we can

gaearon:
For example, we could drop `<Connector>` altogether and encourage people to use `@connect` decorator *as a function in another module*.

### Common case

A dumb component would look exactly as it does now.
A smart component would look like

```js
import React from 'react';
import { connect } from 'react-redux';
import Counter from '../components/Counter';
import * as CounterActions from '../actions/CounterActions';

export default connect(
  state => ({
    counter: state.counter
  }),
  dispatch => ({
    increment: dispatch(CounterActions.increment()),
    decrement: dispatch(CounterActions.decrement())
  })
})(Counter);
```

Note that *the smart component doesn't have to be declared as a component*. Also note that `state => ...` and `dispatch => ...` is all it accepts.

### Case with more control

Want more customization? Want a `componentDidUpdate` hook? Want to select different things depending on the current props? Well, maybe you need to put a component in the middle then:

```js
import React from 'react';
import connect from './connect';
import { bindActionCreators } from 'redux';
import Counter from '../components/Counter';
import * as CounterActions from '../actions/CounterActions';

class CounterContainer {
  componentDidUpdate() {
    ...
  }

  render() {
    const props = somehowSelectChildProps(this.props);
[...]

PR #14, opened 2015-07-30
Update Docs

jlogsdon:
Here's my proposed structure for the documentation. I'm including the original docs from Redux for reference as I move forward. The very top should have a brief "quick start" example, and the sub sections will go into more detail about each.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,101 @@
 react-redux
 =========================
 
-React bindings for Redux.  
-Docs are work in progress.
+Higher-order React components for [Redux](https://github.com/gaearon/redux).
 
-Requires React >= 0.13
+- [Quick Start](#quick-start)
+- [Components](#components)
+  - [`Provider`](#provider)
+  - [`Connector`](#connector)
+- [Decorators](#decorators)
+
+## Quick Start
+
+Hooking up the Counter example into React looks like this:
+
+```js
+import { bindActionCreators } from 'redux';
+import { Provider, Connector } from 'react-redux';
+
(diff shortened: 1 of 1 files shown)


gaearon:
This looks good to me.

jlogsdon:
Cool. I will work on filling out all the sections today/tomorrow.

gaearon:
Thanks, I really appreciate it.

Keats:
@jlogsdon don't forget to look at #1 and #16 so you don't write documentations for something that's going to be removed/changed

gaearon:
@Keats @jlogsdon 

While #1 and #16 are important, it's more important to document the existing APIs right. I'm releasing Redux 1.0 very soon, and I'm going to have examples using React Redux 0.2.1. When we figure #16 out, we'll rewrite the docs, but it might take another week or so. Documenting 0.2.1 is still very important because lots of projects depend on it.

jlogsdon:
Thanks for pointing those out, but I am going to focus on the existing API for the reasons @gaearon said. The new API isn't quite ready, and having docs of the stable release will be very important for the Redux 1.0 release.

I'm working on expanding some of the sections right now and will have something to push up here today.

jlogsdon:
Here we go. The "Child must be a function" section needs to be expanded, but I am not familiar enough with the issue myself and couldn't find any resources about it specifically. If I can get pointed to either the React issue or a fiddle showing the issue I can fill that part out.

gaearon:
Awesome, thanks.

PR #16, opened 2015-07-31
New API for connectDecorator implementation

gnoff:
Alright,

Here's my take on the new connectDecorator API. Normal Args examples should all work from @gaearon 's comment in #1 

https://github.com/gaearon/react-redux/issues/1#issuecomment-120729921

### API
```javascript
decorate = connect(
  (State => StateAny) | null,
  ?((Dispatch => DispatchAny) | actionsObject),
  ?(StateAny, DispatchAny, ParentProps => Props)
)

MyConnectedComponent = decorate(PlainComponent)

//annotation syntax
@connect(
   (State => StateAny) | null,
  ?((Dispatch => DispatchAny) | actionsObject),
  ?(StateAny, DispatchAny, ParentProps => Props))
Component

//component with decoration
Component

//original component
Component.DecoratedComponent
  
```

~~I didn't delete createConnector but I did move createConnectDecorator to createConnectDecoratorDeprecated until we can decide if we should just delete these.~~

the old Connector and connect decorator have been deleted

I have tested through mocha only so no guarantees it is 100%

I'm happy to make changes/work on it etc... as opinions change on best api going forward if it's not ready to merge yet.

State is handled like Connector did on any notification from the store. shouldComponentUpdate will bail if the slice is the same

actions are bound using the provided dispatchBinder (2nd arg) only on component mount. Not sure if this is what you were aiming for in the API but it certainly reduces the number of binds to a lot less than every render and it seemed find since the typical use case for dispatchBinder is to simply bind dispatch to the action creator methods

merge is called on every render, reading stateProps and dispatchBinderProps from component state (not store state) and taking whatever current component props as the 3rd argument.

~~finally the decorated component is rendered with this.props, and mergedProps in that order.~~
If you want choose to provide a custom merge function then only the props returned from merge will be added to the underlying component

I also wrote a util for wrapActionCreators that fits the dispatchBinder argument signature and delegates to redux/bindActionCreators. This is called form connectDecorator if the actionCreators argument is a plain object


gaearon:
This is looking good so far! Can you please remove the old “deprecated” code for cleanness? I don't want to ship the old versions anyway.

src/components/createConnectDecorator.js
```diff
@@ -1,24 +1,121 @@
...
-export default function createConnectDecorator(React, Connector) {
-  const { Component } = React;
+const emptySelector = () => ({});
 
-  return function connect(select) {
-    return DecoratedComponent => class ConnectorDecorator extends Component {
-      static displayName = `Connector(${getDisplayName(DecoratedComponent)})`;
+const emptyBinder = () => ({});
+
+const identityMerge = (slice, actionsCreators, props) => ({...slice, ...actionsCreators, ...props});
```
aaronjensen: We ran into some annoying bugs with a merge like this, specifically the fact that props can override slice. Given that Connect's primary responsibility is getting state from the store and passing it along, I'm not sure if allowing props to override it by default is right.

Some other options:

1. Put props first, allowing slice and actionCreators to override it.
2. Do not merge props in by default, a custom merge can be passed in to do this explicitly if desired. 

For what it's worth, the specific problem we ran into involved nested connectors and the first connector splatting its props to the 2nd. They both had a `loaded` flag, but they meant different things. Nested connectors can happen especially with nested routes.

src/components/createConnectDecorator.js
```diff
@@ -1,24 +1,121 @@
...
+      handleChange(props = this.props) {
+        const nextState = this.selectState(props, this.context);
+        if (!this.isSliceEqual(this.state.slice, nextState.slice)) {
+          this.setState(nextState);
+        }
+      }
+
+      selectState(props = this.props, context = this.context) {
+        const state = context.store.getState();
+        const slice = selectState(state);
```
aaronjensen: We have found use for `selectState` to take `props` as its second argument. It'd be nice if that is supported.

src/components/createConnectDecorator.js
gaearon: I agree `props` should go first.

gaearon: Also, if `props` go first here, they need to be removed from `render`.

src/components/createConnectDecorator.js
aaronjensen: @gaearon in the other thread you said:

> You can make your merge powered by reselect as well.. I think. Am I missing something? Then your third parameter can enjoy the same kind of memoization benefits.

I haven't read it all out, but I think it's just not quite as efficient. Specifically, we could avoid a `setState` here: https://github.com/gaearon/react-redux/pull/16/files#diff-71fbb22950d38c66596e5c31a9bf35eaR74 if the subset we care about (based on props) did not change but the superset did.

gaearon: Is there something preventing you from using `merge` argument for this? That's the intended use case: unlike `select`, it executes on every `render`, so you can access `props` there. Otherwise we'll have to re-evalute `select` on every prop change.

src/components/createConnectDecorator.js
```diff
@@ -1,24 +1,121 @@
...
+        this.state = {
+          ...this.selectState(props, context),
+          ...this.bindActionCreators(context)
+        };
+      }
+
+      componentDidMount() {
+        if (subscribing) {
+          this.subscribed = true;
+          this.unsubscribe = this.context.store.subscribe(::this.handleChange);
```
aaronjensen: Are you planning to stick to ES6 on this repo too?

gaearon: I think it's no big deal here because it's not a public facing example.

src/components/createConnectDecorator.js
gnoff: I'm fine either way we will just need the eventual documentation to indicate that if you set a custom merge you will need to spread props for them to ever reach your underlying component.

either way this gives more control so that is good

gaearon: Yes, I think that would be fine.

src/components/createConnectDecorator.js
aaronjensen: No, the merge argument works. I didn't realize that we could avoid running `select` on props changes.

That said, `select` is probably something that is very fast, and w/ the current `shouldComponentUpdate` above, it'd trigger renders at times when nothing has changed in my scenario, so I'm not sure where the line is. Maybe it's just weird that our container uses props to determine a state subset, but it seems like it may be a common thing for reusable containers.

src/components/createConnectDecorator.js
gnoff: Not sure I understand the original comment? I believe the package builds to es5 for consumption by default. Are you referring to not using es7 features or something?

gaearon: We're removing ES7 features from Redux examples just so people can work with something stable and don't assume Redux requires ES7. But using it in source is fine.

src/components/createConnectDecorator.js
gaearon: It's a common thing, but then it's weird you can't do the same for actions. And you can't because that would be a perf problem (functions being bound every time). At least `merge` is a *single* place to go when you want to do something with props. I think it's OK.

src/components/createConnectDecorator.js
gnoff: oh gotcha, object spread and all that.

gnoff:
@gaearon mind also checking out this additional PR (currently on my fork only since it looks cleaner comparing to my master)

https://github.com/gnoff/react-redux/pull/1

test/components/connect.spec.js
```diff
@@ -118,13 +257,238 @@ describe('React', () => {
...
+            </Provider>
+          );
+        }
+      }
+
+      let tree = TestUtils.renderIntoDocument(<OuterContainer />);
+      let div = TestUtils.findRenderedDOMComponentWithTag(tree, 'div');
+      expect(div.props.children).toBe(42);
+
+      tree.setState({ select: selectB });
```
gnoff: This test is failing. might just be a jsdom issue but I'm really unfamiliar with react unit testing so I'd love some eyes on this to either fix the test or confirm that this functionality is actually broken in my implementation

gaearon: I don't think this test makes sense anymore. `select` isn't a prop now, you must specify it during the class definition.
I'm not sure *why* it breaks but you can remove it.

src/components/createConnectDecorator.js
```diff
@@ -1,24 +1,121 @@
...
+const emptyBinder = () => ({});
+
+const identityMerge = (slice, actionsCreators, props) => ({ ...props, ...slice, ...actionsCreators});
+
+
+export default function createConnectDecorator(React) {
+  const { Component, PropTypes } = React;
+  const storeShape = createStoreShape(PropTypes);
+
+  return function connect(select, dispatchBinder = emptyBinder, mergeHandler = identityMerge) {
```
Keats: should we rename select to selector as mentioned in #15 ?

src/components/createConnectDecorator.js
```diff
@@ -1,24 +1,121 @@
...
-          <Connector select={state => select(state, this.props)}>
-            {stuff => <DecoratedComponent {...stuff} {...this.props} />}
-          </Connector>
+      isSliceEqual(slice, nextSlice) {
+        const isRefEqual = slice === nextSlice;
+        if (isRefEqual) {
+          return true;
+        } else if (typeof slice !== 'object' || typeof nextSlice !== 'object') {
+          return isRefEqual;
+        }
```
Keats:
```js
if (isRefEqual || (typeof slice !== 'object' || typeof nextSlice !== 'object') {
  return isRefEqual;
}
```
would be cleaner

src/components/createConnectDecorator.js
```diff
@@ -1,24 +1,121 @@
...
+export default function createConnectDecorator(React) {
+  const { Component, PropTypes } = React;
+  const storeShape = createStoreShape(PropTypes);
+
+  return function connect(select, dispatchBinder = emptyBinder, mergeHandler = identityMerge) {
+
+    const subscribing = select ? true : false;
+    const selectState = select || emptySelector;
+    const bindActionCreators = isPlainObject(dispatchBinder) ? wrapActionCreators(dispatchBinder) : dispatchBinder;
+    const merge = mergeHandler;
```
Keats: I don't understand the point of that one, it would be actually easier to keep it called mergeHandler to distinguish easily from the Component.merge

src/components/createConnectDecorator.js
```diff
@@ -1,24 +1,121 @@
...
+
+export default function createConnectDecorator(React) {
+  const { Component, PropTypes } = React;
+  const storeShape = createStoreShape(PropTypes);
+
+  return function connect(select, dispatchBinder = emptyBinder, mergeHandler = identityMerge) {
+
+    const subscribing = select ? true : false;
+    const selectState = select || emptySelector;
+    const bindActionCreators = isPlainObject(dispatchBinder) ? wrapActionCreators(dispatchBinder) : dispatchBinder;
```
Keats: Using the same name as a redux function but with a different signature is a bit confusing.
Also @gaearon I don't really see the point of the wrapActionCreators mentioned in #1, why not just pass the actionCreators directly to connect and bind it using redux.bindActionCreators (at line 93 in this file) ?

src/index.js
```diff
@@ -1,4 +1,4 @@
 import React from 'react';
 import createAll from './components/createAll';
 
-export const { Provider, Connector, provide, connect } = createAll(React);
+export const { Provider, Connector, provide, connect, connectDeprecated } = createAll(React);
```
Keats: @gaearon should it still export Connector and connectDeprecated rather than removing it completely ?

quirinpa:
Cool xD When do you think this will be merged? Can't wait to use it hehe

gaearon:
About a week after Redux 1.0 is released, I presume.

quirinpa:
ok, thank you @gaearon :)

src/components/createConnectDecorator.js
gnoff: I'm in favor of arguments being named consistently as nouns or verbs. so select -> selector or change

selector -> select
dispatchBinder -> dispatchBind
mergeHandler -> merge

src/components/createConnectDecorator.js
gnoff: This was lifted straight from Connector

@gaearon is this a micro optimization already?

src/components/createConnectDecorator.js
gnoff: per lint rules I can't reassign the function arguments so it needs a different name if we want an opportunity to do anything with the argument (which with default args we aren't at the moment) then it will need a different name.

I actually don't mind that these const values overload the naming of the local component methods since it implies a delegating type relationship (which is exactly how they are used). I'd prefer to leave these alone but will take additional input

src/components/createConnectDecorator.js
gnoff: yeah, this is really unfortunate, especially since we have an invariant check that reports this method name if you pass in a bad dispatchBind. I would love to change this

Regarding the wrap, I want to keep this since the second arg is not actually a 'definitely wrap action creators' function, it's a general purpose opportunity to do something with dispatch. the wrap is just convenience for the overhwhelming number of times the use case will call for a simple wrap+bind

Keats: but wouldn't it be simpler to just differentiate functions (ie you want to do something with dispatch) and plain object (ie you want to bindActionCreators).
The connector in the second case would be
```js
@connect(
  null,
  myActionCreators
)
```
and in the first case:
```js
@connect(
null,
(dispatch) => (..)
)
```
which will be bound automatically by the connector
Or am I missing something?

gnoff: Yup, this PR supports the two scenarios above. if connect receives a plain object for the dispatchBinder argument it assumes it is plain action creators and internally wraps them in wrapActionCreators (which in turn calls bindActionCreators from redux)

If it is not a plain object it needs to be a function that returns and object (presumably using the dispatch arg it is handed... but you can really do whatever you want here)

gnoff:
Should we consider changing the static member DecoratedComponent on our wrapped component. If you use  ```@connect()(Component)``` you will end up with the Symbol for ```Component``` pointing to the decorated component ```Component.DecoratedComponent``` pointing to the undecorated component. This feels... misleading?

Perhaps we change ```static DecoratedComponent``` in createConnectDecorator.js to ```static UndecoratedComponent```

test/components/connect.spec.js
gnoff: Alright, i'll remove it

src/components/createConnectDecorator.js
gnoff: I changed bindActionCreators internal to createConnectDecorator to instead be bindDispatch.

This maps to the public api more directly and avoids name collision confusion with redux/bindActionCreators

quirinpa:
@Keats Hey Keats, great job! I have a question: If we use the merge method to select inside a store, does it update only when that particular "reselect" gives up different values or does it re-render either way?

rt2zz:
+1 just tried it out and it is working as described

rt2zz:
note: nesting connectors does not work in react 0.13, but does work in 0.14. Something to do with the change to parent contexts.

Keats:
@quirinpa I'm not the one that did the PR!
It's @gnoff

quirinpa:
@Keats oh :P my mistake

PR #18, opened 2015-08-01
Components should extend Component class

sapegin:


gaearon:
Fair enough! (React allows `Component`less components, but they don't have `setState` so people get confused.)

sapegin:
Actually I didn’t know that and it could confuse other people too. Thanks!

PR #29, opened 2015-08-07
Readme: connect() should wrap Counter in example

af:
New to the project, so it's possible I'm reading this wrong, but I think in this example `connect()` is wrapping the wrong component.

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,13 +111,13 @@ function mapDispatchToProps(dispatch) {
 export default connect(
   mapStateToProps,
   mapDispatchToProps
-)(CounterContainer);
+)(Counter);
 
 // You can also pass an object instead of defining `mapDispatchToProps`:
-// export default connect(mapStateToProps, CounterActionCreators)(CounterContainer);
+// export default connect(mapStateToProps, CounterActionCreators)(Counter);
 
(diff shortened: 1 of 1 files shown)


hartzis:
They way it is listed now in the readme seems to be correct.

Since you're "injecting" the state as props into `CounterContainer`. The `CounterContainer` now has access to the state from its `this.props`. Check out this example:

https://github.com/gaearon/redux/blob/master/examples/counter/containers/CounterApp.js

I know this is an old example, but look how connect passes `state.counter` as `this.props.counter` in the render function. `CounterApp` is essentially `CounterContainer` in this example.

Does this help to explain it at all?

af:
Thanks for the feedback @hartzis. Perhaps the problem then is the import statements– `Counter` is imported but never used, while `CounterContainer` is not imported nor defined anywhere, but is then used as a function argument. At the least running the code as-is will give you a ReferenceError :)

gaearon:
@af is correct.

@hartzis Master is a bit outdated, I just released `react-redux@0.5.0` so official examples will be updated later. The intermediate component isn't needed anymore! See also https://github.com/gaearon/react-redux/releases/tag/v0.5.0 release notes.

hartzis:
@gaearon That is awesome! I totally see that now.

@af Great catch!

Issue #28, opened 2015-08-07
Uncaught TypeError: this.unsubscribe is not a function


aaronjensen:
Occasionally, when clicking through our app we get this:

![oliver 2015-08-07 11-01-56](https://cloud.githubusercontent.com/assets/8588/9142321/be9e7822-3cf3-11e5-8008-bf31d5a373bb.png)

Is it possible for will unmount to fire before mount? Should we check to ensure `unsubscribe` is there before calling it or is something else going on?

aaronjensen:
I believe this has to do w/ some incorrect stuff that was happening in the app, though I don't know exactly what yet. This may be something that should never normally happen.

aaronjensen:
Ok, so here's what we were doing... in `componentWillMount` we fired an action via `setTimeout` so that the `react-redux` subscription in `componentDidMount` will have already happened by the time the action is dispatched. It appears that this is a bad idea, though I do not understand why.

gaearon:
OK I introduced this bug while tweaking @gnoff's code. :-(
Will push out a fix soon.

gaearon:
Can you verify the bug is fixed in `0.5.1`?

aaronjensen:
yep, it is. But then I get this: 

```
Uncaught Error: Invariant Violation: Component (with keys: getDOMNode,_handleChange,props,context,state,refs,_reactInternalInstance) contains `render` method but is not mounted in the DOM
```

which makes me think we're doing something very wrong. By removing the `setTimeout` in `componentWillMount` it works out.

What's the right way to fire an action when a component mounts? If the action immediately modifies state, `componentWillMount` is too soon, because react-redux doesn't subscribe until `componentDidMount`...

gaearon:
>contains `render` method but is not mounted in the DOM

I found this: https://github.com/facebook/react/issues/4233.

Sounds like React throws if you use `findDOMNode(this)` when component is already unmounted. You should clear any previously set interval in `componentWillUnmount`.

>If the action immediately modifies state, componentWillMount is too soon, because react-redux doesn't subscribe until componentDidMount.

What do you mean by "too soon"? If React Redux doesn't update in some edge case, it's a bug, no matter where you call `dispatch`! You should file an issue, not just accept it like that. :smile:

aaronjensen:
I just assume that you've thought of everything! I'll make sure it is what I think it is and file an issue when I get a sec. Thanks!

gaearon:
[...]

PR #31, opened 2015-08-08
Dispatches during componentWillMount are not observed

aaronjensen:
Failing spec for issue described in #28


gaearon:
Should be fixed in 0.5.3. Please verify!

This is not a correct fix because it causes subscription to occur on server (server executes `componentWillMount`) but never freed.

aaronjensen:
Ah! I knew there was a reason to not use `componentWillMount`. Thanks.

Issue #32, opened 2015-08-08
on version >= 0.5 shouldComponentUpdate is not working properly!


maxguzenski:
Hi,
I have a redux reducers to keep a hash of all users on my app, and I have some react component connect to only one user of this hash. before version 0.5, my connect was: 

ps.: I use immutable-js

```javascript
@connet( (state, props) => {
  return { user: state.users.get( props.userId ) }
})
``` 

Now, I cant access props from first method, and need to use 3th method to get correctly user: 

```javascript
@connect(
  s => { return {users: s.users}},
  d => { return {actions: bindActionCreators(actions, d)}},
  (s, d, p) { return {...d, ...p, { user: s.users.get( p.userId)}}} 
)
``` 

What happens is that shouldComponentUpdate from connects doesnt see this filter and re render my component all time that ANY users are changed/added/removed

I strong believe that this functions should be binded to connect, and let developer access this.props/this.state from there, or, at least, pass parent props... In my case having this third method is just a overkill

ps.: This issue happens on this example as well: 
https://github.com/gaearon/react-redux#inject-todos-of-a-specific-user-depending-on-props-and-inject-propsuserid-into-the-action

gaearon:
Please see if the issue is fixed in `0.6.0`. We still don't give you `props` in the first function, but if the third function returns props shallowly equal to the previous props, no update should happen.

maxguzenski:
yes! now it is working ... no re-render() :-)

maxguzenski:
Just to know, you plan to give parent props to first function or you don't sure if it is a good thing?

gaearon:
If everyone keeps asking about this, I'll consider it, but I also like that the API is symmetric.

There are some benefits to keeping it a function of single argument:

1. The selector API stays simple. No need for things like https://github.com/faassen/reselect/issues/20 that complicate sibling libraries to Redux.
2. If you need to select state by props, most likely you'll need to inject props into action creators too. And you would be surprised if props were passed to the first function, but not to the second one. And we can't pass it to the second function because we only want to bind actions once for performant most common use case.

PR #34, opened 2015-08-08
Rewrite to improve speed

gaearon:
This should address #32.
@maxguzenski Can you try this branch?


maxguzenski:
I dont know, how I build it? I tried this inside my project: 
npm install github:gaearon/react-redux#fast

but without success

gaearon:
Hmm. Let me just release it as `0.6.0`. If there are any issues we'll just fix them and add more tests. :-)

maxguzenski:
when I tried: npm install github:gaearon/react-redux#fast
there is no error, but there is no lib/ folder (even src/ folder) into node_modules/react-redux
I have babel and webpack installed global.

PR #35, opened 2015-08-08
Hardcode decorated component ref

jhollingworth:
Not setting the ref of the inner component makes it a pain to [test through refs](https://github.com/QubitProducts/react-test-tree#higher-order-components).

Is there any specific reason for setting the ref through `setUnderlyingRef` or could we just set the ref to a string?

### DIFF
--- a/src/components/createConnect.js
+++ b/src/components/createConnect.js
@@ -81,7 +81,6 @@ export default function createConnect(React) {
       constructor(props, context) {
         super(props, context);
         this.version = version;
-        this.setUnderlyingRef = ::this.setUnderlyingRef;
 
         this.stateProps = computeStateProps(context);
         this.dispatchProps = computeDispatchProps(context);
@@ -179,16 +178,12 @@ export default function createConnect(React) {
       }
 
       getUnderlyingRef() {
-        return this.underlyingRef;
-      }
-
-      setUnderlyingRef(instance) {
-        this.underlyingRef = instance;
+        return this.refs.underlyingRef;
       }
 
       render() {
         return (
-          <DecoratedComponent ref={this.setUnderlyingRef}
+          <DecoratedComponent ref='underlyingRef'
                               {...this.state} />
         );
       }


gaearon:
Is calling `getUnderlyingRef()` a burden in this case?

jhollingworth:
Unfortunately yes. We're using [react-test-tree](https://github.com/QubitProducts/react-test-tree) which wraps a component and any children in a testable interface (Essentially a page object). Not having a ref present in the component means it cannot find any children.

Also, the fact that you have to call `getUnderlyingRef()` at all is a bit of a smell. The higher order component is really just an implementation detail and you rarely want to test it, rather the inner component. You can [tell react-test-tree to make HoC's invisible](https://github.com/QubitProducts/react-test-tree#higher-order-components) making testing much more pleasant. Again, not having a `ref` there makes this impossible.

Out of all the issues I've just raised this is the only one that is completely blocking us from moving to redux 1.0. We used to use the `<Connector .../>` which allowed us to support our desired syntax but thats now gone...

What's your motivation behind not making the decorated component available via `element.refs.underlyingRef` considering the existing `getUnderlyingRef()` API is still supported?

gaearon:
>Out of all the issues I've just raised this is the only one that is completely blocking us from moving to redux 1.0. 

Wait wait. Redux is in no way related to React Redux. You can use `redux@1.0.0-rc` with `react-redux@0.2.1` (or whatever worked for you) just fine while we're figuring it out..

gaearon:
>Also, the fact that you have to call getUnderlyingRef() at all is a bit of a smell. The higher order component is really just an implementation detail and you rarely want to test it directly, rather the impact it has on the inner component.

Why not export the vanilla component then and test it directly without the HOC? I'm just trying to understand the issue better.

Are you testing the whole tree instead of particular components?

gaearon:
I'm OK with merging this, but let's rename `getUnderlyingRef()` to `getInstance()` and `refs.underlyingRef` to `refs.instance`?

gaearon:
I changed names a bit and put it in https://github.com/gaearon/react-redux/commit/b7ef1789d14761d189be4cc4857608c5c82233e3, will release soon. Thanks for the tip.

jhollingworth:
Awesome thanks :smile:

PR #38, opened 2015-08-08
Don't throw error if the store isn't in the context

jhollingworth:
I've written a smaller wrapper around the connect component which, amongst other things, allows you to skip all the redux stuff and just render the inner component. This can be really helpful when you don't want to mess with setting up contexts, etc and test the render.

```js
@dataful
class User extends React.Component {
  static data = {
    user: ({users}, {userId}) => users[userId]
  }

  static actions = {
    updateUser
  }

  render () {
    const {
      data: {
        user
      }
    } = this.props

    return <div className='User'>{user.name}</div>
  }

  onClick () {
    this.props.actions.updateUser()
  }
}


const actions = {
  updateUser: sinon.spy()
}

const data = {
  user: {
    id: 123,
    name: 'Foo'
  }
}

<User actions={actions} data={data} />
```
The issue is `connect` requires that `context.store` is there. Something I'd ideally like to avoid if possible for ease of testing. Would it be OK if we just did nothing if the store isn't in the context?

### DIFF
--- a/src/components/createConnect.js
+++ b/src/components/createConnect.js
@@ -16,6 +16,12 @@ function getDisplayName(Component) {
   return Component.displayName || Component.name || 'Component';
 }
 
+const nullStore = {
+  getState() {},
+  subscribe() {},
+  dispatch() {}
+};
+
 // Helps track hot reloading.
 let nextVersion = 0;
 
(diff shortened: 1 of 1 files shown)


gaearon:
Let's do it another way: allow `store` to be passed as a prop to any `connect()`ed component. If it is passed, it will be used instead of `context.store`. Would you update your PR to implement this?

jhollingworth:
Closing in favour of #44

PR #40, opened 2015-08-08
Fixes breaking change for React Native

### LINKED ISSUE #39: React Native support is broken
https://github.com/gaearon/react-redux/blob/7458a923769e43dc647c270a7102d55236fa3b79/src/components/createConnect.js#L149

`process` is not available in React Native. React Native does however have a `__DEV__` global.

lwansbrough:
Fixes #39

### DIFF
--- a/src/components/createConnect.js
+++ b/src/components/createConnect.js
@@ -146,7 +146,7 @@ export default function createConnect(React) {
       }
 
       componentWillUpdate() {
-        if (process.env.NODE_ENV !== 'production') {
+        if (__DEV__ || (process && process.env.NODE_ENV !== 'production')) {
           if (this.version === version) {
             return;
           }


gaearon:
This breaks the web version: `ReferenceError: __DEV__ is not defined`.
You can run `npm test` to check.

gaearon:
Another problem with this condition is that Uglify won't know what to cut out at the build time because people's Webpack/Browserify configs have no idea what `__DEV__` is.

gaearon:
For now I think I fixed it with https://github.com/gaearon/react-redux/commit/238e7d90202ea51495e5b217e74619a3150ca93e. Will be out in `0.7.0` shortly, I'll deprecate `0.6.0`.

gaearon:
Please let me know if `0.7.0` helps.

lwansbrough:
Thanks, I'm out right now but I'll give it a shot when I get back. Cheers!

PR #44, opened 2015-08-09
Support passing store in via props for testing

### LINKED ISSUE #33: Make context-less version of connect
Suggestion: make component wrapped by `connect()` optionally accept `store` as a prop.
This may help integration to codebases that don't use React for everything, and can't rely on context.

jhollingworth:
Allow you to pass the redux store in via props to make it easier to unit test (Instead of having to mess with `Provider`'s or contexts)

See also #38

Closes #33

### DIFF
--- a/src/components/createConnect.js
+++ b/src/components/createConnect.js
@@ -93,7 +93,7 @@ export default function createConnect(React) {
             `props of "${this.constructor.displayName}". ` +
             `Either wrap the root component in a <Provider>, ` +
             `or explicitly pass "store" as a prop to "${this.constructor.displayName}".`
-          )
+          );
 
           this.stateProps = computeStateProps(this.store);
           this.dispatchProps = computeDispatchProps(this.store);
--- a/src/components/createConnect.js
+++ b/src/components/createConnect.js
@@ -34,8 +34,8 @@ export default function createConnect(React) {
     // Helps track hot reloading.
     const version = nextVersion++;
 
-    function computeStateProps(context) {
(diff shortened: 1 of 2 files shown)


src/components/createConnect.js
```diff
@@ -72,7 +72,7 @@ export default function createConnect(React) {
         static WrappedComponent = WrappedComponent;
 
         static contextTypes = {
-          store: storeShape.isRequired
+          store: storeShape
         };
```
gaearon: Let's add similar `propTypes` then?

jhollingworth: done

src/components/createConnect.js
```diff
@@ -82,13 +86,17 @@ export default function createConnect(React) {
         constructor(props, context) {
           super(props, context);
           this.version = version;
-          this.stateProps = computeStateProps(context);
-          this.dispatchProps = computeDispatchProps(context);
+          this.store = props.store || context.store;
+
+          invariant(this.store, '`store` must be passed in via the context or props');
```
gaearon: Let's change it to

Could not find "store" in either context or props of "${this.constructor.displayName}". Either wrap the root component in a <Provider>, or explicitly pass "store" as a prop to "${this.constructor.displayName}".

and also verify that the error message is correct in a unit test.

jhollingworth: updated with test.

gaearon:
Thank you for your work on this. Released in `0.8.0`.

PR #50, opened 2015-08-11
Fix undefined children does not update children in state

istarkov:
If at first render you have children props, and at second not 
(this.props does not contain children element), 
this.state contains previous version of children.

### DIFF
--- a/src/components/createConnect.js
+++ b/src/components/createConnect.js
@@ -131,7 +131,13 @@ export default function createConnect(React) {
         recomputeState(props = this.props) {
           const nextState = this.computeNextState(props);
           if (!shallowEqual(nextState, this.state)) {
-            this.setState(nextState);
+            this.setState({
+              ...Object.keys(this.state)
+                .reduce((r, k) => ({
+                  [k]: undefined
+                }), {}),
+              ...nextState,
+            });
           }
         }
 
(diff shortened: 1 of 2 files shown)


istarkov:
And it look's like the problem is deeper 
for any component props which is gone 
you can't use setState because of it's merge nature,
it's need something like replaceState

gaearon:
Can you add a failing test case instead? I don't understand the problem yet.

istarkov:
Ok i'll add, the problem is that setState merges `newState` with old state.
And if at first rendering step you have some prop, and at next rendering step this prop is not exists,
connector state still contains prop from first rendering.
As example
```javascript
render () {
  const obj = this.props.a ? {someProp: 1} : {};
  <ReduxConnectedComponent {...obj} />
}
```
if `this.props.a` was `true`, and becomes `false`, `ReduxConnectedComponent` still contains `this.props.someProp` equal to `1`

istarkov:
So recomputeState must be like
```javascript
        recomputeState(props = this.props) {
          const nextState = this.computeNextState(props);
          if (!shallowEqual(nextState, this.state)) {
            this.setState({
              ...Object.keys(this.state)
                .reduce((r, k) => ({
                  [k]: undefined
                }), {}),
              ...nextState,
            });
          }
        }
```

istarkov:
I've add failing test, (without @connect it works as expected)

gaearon:
I got it now, thanks a lot!

gaearon:
This is out in 0.8.1, if you see any use cases that aren't covered yet, feel free to add more tests ;-)

PR #51, opened 2015-08-11
Add failing test

istarkov:
8.1 failing test
prehistory [here](https://github.com/gaearon/react-redux/pull/50#issuecomment-129953612)

Test in first commit is new.

Changes in second commit is about that such keys also need to be deleted from this.state

### DIFF
--- a/src/components/createConnect.js
+++ b/src/components/createConnect.js
@@ -12,8 +12,6 @@ const defaultMergeProps = (stateProps, dispatchProps, parentProps) => ({
   ...dispatchProps
 });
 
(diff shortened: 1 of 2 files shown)


istarkov:
2 fix first failing test, it's need to change 
```javascript
const resetValue = (_, key) => ({[key]: undefined });
```
on this
```javascript
const resetValue = (r, key) => ({ ...r, [key]: undefined });
```

istarkov:
Third commit fixes both tests, 
to emulate replaceState i've moved current `state` to `state.props`

Issue #52, opened 2015-08-11
Add ownProps to mapStateToProps and mapDispatchToProps as second parameter


gaearon:
It should only be passed if `fn.length > 1` as per https://github.com/gaearon/react-redux/pull/36#issuecomment-129945686.

If you want to make a PR, please write here so we don't duplicate effort.

gaearon:
Important note: if `mapDispatchToProps.length < 2`, `mapDispatchToProps` should not be called on prop change.

Similarly, if `mapStateToProps.length < 2` or `mergeProps.length < 3`, `componentWillReceiveProps` should not result in their recalculation, respectively.

gaearon:
cc @jhollingworth, @danharper, @ForbesLindesay who wanted to make this happen

jhollingworth:
Happy to implement this if no one else is

gaearon:
Feel free to!

PSpSynedra:
@gaeron What is the current state of this issue? 
We were just updating to redux 1.0.x and have serious problems with props not being available to mapStateToProps. The reason is that we pass memoized selector functions via props that select the desired childProperties from the state. 
As a workaround we could select the whole state and thus have it passed to the mergeProps callback. The problem using this workaround is, that the potentionally expensive mergeProps becomes the mapStateToProps and is callled much to often (the recomputeState is actually called for every state change in the redux store, no matter what part of the state was influenced by the change).

theaqua:
@PSpSynedra https://github.com/gaearon/react-redux/pull/59#issuecomment-131764832

PSpSynedra:
@theaqua, Yeah I just saw it :-) Looking forward to it already.

gaearon:
Released in 0.9.0, please verify.

PR #55, opened 2015-08-14
Pass parent props into mapStateToProps

ide:
Addresses part of #52 by passing the parent props into mapStateWithProps.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ React Redux
 Official React bindings for [Redux](https://github.com/gaearon/redux).  
 Performant and flexible.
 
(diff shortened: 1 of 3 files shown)


jhollingworth:
This doesn't actually check whether we are passing in `props`, see #36

gaearon:
Thanks for the effort! I'm closing in favor of #59 which is a more complete solution.

gaearon:
A version of this is released in 0.9.0, please verify.

ide:
Sweet, thank you guys! Will test it out sometime today.

PR #59, opened 2015-08-16
Add ownProps to mapStateToProps and mapDispatchToProps as second parameter

### LINKED ISSUE #52: Add ownProps to mapStateToProps and mapDispatchToProps as second parameter

jhollingworth:
If you pass in more than one argument to `mapStateToProps` and `mapDispatchToProps` then we assume you are passing in props and so pass it in. Whenever the component receives new props we will re-invoke the function.

Fixes #52


jhollingworth:
@gaearon

gaearon:
Good stuff. I'll merge today. :heart:

gaearon:
Released as 0.9.0, please verify.

PR #67, opened 2015-08-21
Fix some typos and inconsistencies in documentation

agraboso:


README.md
```diff
@@ -366,7 +366,7 @@ function mapStateToProps(state) {
 }
 
 function mapDispatchToProps(dispatch) {
-  return bindActionCreators(Object.assign({}, todoActionCreators, counterActionCreators), dispatch);
+  return bindActionCreators({ ...todoActionCreators, ...counterActionCreators}, dispatch);
```
gaearon: We'd rather not use ES7 experimental code in examples to avoid confusing beginners.

agraboso: There are several other instances of object spread notation in the README, both in the text and in the examples. If you want, I can revert them all back to Object.assign.

agraboso: Hmm, in fact there were only the two that I introduced and two more.

gaearon: Yeah, I'd prefer to use `assign` everywhere.

gaearon:
Thanks.

PR #70, opened 2015-08-24
Use parent context for <Provider> in React 0.14

gaearon:
This implements https://github.com/rackt/react-redux/issues/64 in a backward-compatible by checking `React.version`. Because I updated the tests to use 0.14 beta, I also manually ran the new source against the previous test suite, and it passed.

There's a single breaking change here: `<Provider>{() => {})</Provider>` will no longer work, as a single child (either as a function child or a vanilla child) is enforced.

When you try to use function as a child in 0.14, a warning is printed to the console. Conversely, React Redux warns you if you use a vanilla element in 0.13 where function as a child is needed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,6 @@
     "mocha-jsdom": "~0.4.0",
     "react": "^0.14.0-beta3",
     "react-addons-test-utils": "^0.14.0-beta3",
-    "react-dom": "^0.14.0-beta3",
     "redux": "^1.0.1",
     "rimraf": "^2.3.4",
     "webpack": "^1.11.0"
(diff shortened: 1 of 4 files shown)


gaearon:
>does react-redux state anywhere that it only supports React 13? I thought context was available in 12 and earlier too.

It doesn't state this, but it won't work with 0.12 and nobody complained so far. We don't enforce a `peerDependency` because until everybody uses NPM3 it's often more trouble than it's worth.

We should add a note to README saying we only support 0.13+, or else we need to migrate to `createClass` and support 0.12 (but I don't see the point).

gaearon:
>I would also additional minor versions

?

gaearon:
Noted 0.13+ in the README: https://github.com/rackt/react-redux/commit/949c3992cadd64c1f517d6fb8f54952d622700db

gnoff:
minor versions meaning 11 or 12 but since you put the note on requiring react 13 or higher I agree that's not necessary.

PR #71, opened 2015-08-24
Drop support for React 0.13

gaearon:
We won't do this yet, but React 0.14 makes everything simpler for us.
When it's out, releasing this will be a breaking change.

Changes:

* No more support for function inside `<Provider>`, as we have parent context in 0.14
* No more jumping through hoops to support React Native: we can still `import React from 'react'`
* We can put `react` in `peerDependencies` now because it will work for React Native, and also won't cause “duplicate React” issues
* With React Native 0.10, we can rely on `process.env.NODE_ENV` polyfill being there

Checklist:

- [x] Drop support for React 0.13 in a branch
- [x] Wait for React 0.14
- [ ] <s>Wait for React Native 0.10 (will it use 0.14?)</s>
- [x] Update the docs
- [ ] Merge and release as a breaking change


gnoff:
What is the impetus from switching for a creator pattern to directly exporting the two components?

gaearon:
>What is the impetus from switching for a creator pattern to directly exporting the two components?

The reason a factory was added in the first place is because we either needed `React` from `react` or from `react-native`, depending on what user wanted, and couldn't express this statically. With 0.14 we can just use `react` because `react-native` will just be a renderer like `react-dom`, and won't contain `Component`, `PropTypes`, etc.

gaearon:
Updated this.

cwbutler:
Still waiting on React Native?

gaearon:
Not really—we'll just ask RN users to stay on previous major. I'll release this soon.

PR #72, opened 2015-08-25
Improve handling of React Native version warnings

udfalkso:
Using React Native, a disruptive error appears to warn about function/no-function for Provider in app root.  It seems that the version check is failing as React.version is undefined in my React Native environment.  

As these are just warnings, perhaps it's best to use console.warn rather than console.error?  Or perhaps there is a better way to fix the version check?  

{
    "react": "^0.13.3",
    "react-native": "^0.8.0",
    "react-redux": "^1.0.0",
    "redux": "^1.0.1",
    "redux-thunk": "^0.1.0"
}
ios simulator screen shot aug 25 2015 11 52 03 am

### DIFF
--- a/src/components/createProvider.js
+++ b/src/components/createProvider.js
@@ -3,7 +3,7 @@ import createStoreShape from '../utils/createStoreShape';
 function isUsingOwnerContext(React) {
   const { version } = React;
   if (typeof version !== 'string') {
-    return false;
+    return true;
   }
 
   const sections = version.split('.');


gnoff:
@udfalkso since React Native is not using React 0.14 the correct fix would be to change https://github.com/udfalkso/react-redux/blob/console-warn-instead-of-error-for-function-wrap/src/components/createProvider.js#L6 to `return true;`

In fact i believe it is a genuine bug that we are returning false here since the only version of React today not using Parent Context is 0.14 beta.

gnoff:
mind recreating the PR with that one change? Or i can do so in a minute

udfalkso:
Ok, how's that look?

gnoff:
perfect, can you confirm this fixes the bug on react-native?

udfalkso:
Yup, the error display goes away.

gaearon:
Interesting. I returned `false` there because I assumed that today any React distribution has a version, so if it is missing, we are running a much more modern React..

gaearon:
There's something I don't understand then.

`fbjs` [uses `console.error` for `warning`](https://github.com/facebook/fbjs/blob/master/src/__forks__/warning.js#L42).
[`react` uses `warning`](https://github.com/facebook/react/blob/master/src/isomorphic/classic/element/ReactElementValidator.js#L252) for PropTypes validation.

How is React Native different? I think we should use the same warning kind as `PropTypes` validation.

gaearon:
Should be fixed in 1.0.1.

gnoff:
Yeah, i figured that was the intention of the original `return false;` but since we are going to drop all of this logic with R14 and RN10 (hopefully) and just use a peer Dependency i figured being more conservative with the version check is the proper choice right now

udfalkso:
Thanks guys!

gnoff:
RE: warnings for PropTypes I actually think the red screen is what you get too.

PR #77, opened 2015-08-28
fix eslint error and simplify webpack external option.

ironhee:


gaearon:
Thanks, but I'd rather keep configuration explicit.

ironhee:
Ok thank you :D

PR #80, opened 2015-09-01
Remove support for changing store on the fly

gaearon:
This removes the magical reducer replacement in favor of explicit `replaceReducer()` call where you create your store. See https://github.com/rackt/redux/pull/667 for details.


aFarkas:
From you relaease notes about hot module reload:
> See https://github.com/rackt/redux/pull/667 for a migration example.

The link provided isn't really helpful.

PR #90, opened 2015-09-03
Impure option

dallonf:
Added a new `options` argument to `connect()`, with just one option so far: `pure`. If set to `false` (it defaults to `true`), it will bypass `shouldComponentUpdate`. This resolves https://github.com/rackt/react-redux/issues/88.

Let me know if it needs any better explanation in the README.

It might be slightly more performant to do something like `if (!finalOptions.pure) { delete Connect.prototype.shouldComponentUpdate }`, but that's a little bit too hackish for me to really be proud of...

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ Performant and flexible.
 - [Quick Start](#quick-start)
 - [API](#api)
   - [`<Provider store>`](#provider-store)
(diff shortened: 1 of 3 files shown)


gaearon:
This looks good to me.
@gnoff what do you think?

gnoff:
Yeah looks good to me.

Issue #86, opened 2015-09-02
connect(mapStateToProps) is executed with component props based on the previous state


epeli:
Ok, this is really hard to explain so I've set up an example here:

https://github.com/epeli/redux-connect-demo

Build it and play with it with devtools console opened. You should see following errors:

    Uncaught TypeError: Cannot read property 'foo' of undefined

This happens because the [`mapStateToProps` of the `Item` component][1] is called with component props based on the previous store state. The two states are not compatible anymore and it causes all kinds of errors.

I think the parent should render and update the props for the `Item` component before the `mapStateToProps` of the `Item` component can be executed.

[1]: https://github.com/epeli/redux-connect-demo/blob/master/index.js#L40-L43

gaearon:
>This happens because the mapStateToProps of the Item component is called with component props based on the previous store state.

Why do you say the *previous* state? I think they are called with the most *recent* state—and because the `Item`s have not yet unmounted, they query state that doesn't exist anymore.

I don't think this is something we can fix. React state changes are asynchronous and React may (or may not) batch them. Therefore, the moment you press “Remove”, the Redux store updates, and both `Item` and `App` receive the new state. Even if the `App` state change results in unmounting of `Item`s, that will happen later than `mapStateToProps` is called for `Item`.

Unless I'm mistaken, there is nothing we can do. You have two options:

1. Request all required state at `App` (or a lower, e.g. `ItemList`) level and pass it down to “dumb” `Item`s.
2. Add safeguards to `mapStateToProps` for “currently unmounting” state. For example, you may return `null` from `render` in this case.

*Potentially* we could have the component generated by `connect()` return `null` from its `render` if `mapStateToProps` returned `null`. Does this make any sense? Is this too surprising?

epeli:
> Why do you say the previous state?

Because the store has been updated and the props `App` has passed to the `Item` are based on an old store state when the store query is run.

> Therefore, the moment you press “Remove”, the Redux store updates, and both Item and App receive the new state. 

Hmm. Is it actually required for the both to be subscribed to the store? Wouldn't it be enough if only the one most closest to the component root is subscribed? When that component renders it will trigger a render (and a store query) in all child components.

Seems like it would solve this. Although I have no idea how to detect the one closest to the root.

epeli:
Darn. Just realized it would break apps using pure components. Because the props between the parent and child components does not necessary change it might prevent some children from updating because React thinks there's nothing to the update but there might be because the store has changed.

[...]

PR #99, opened 2015-09-08
Fix issues with stale props

epeli:
Fix for #86

The idea in this change is to delay executing `mapStateToProps` until the render call. The render call is forced for every state change by setting the store state to the component state for every connect wrapper. This works because state changes are batched in React. Which means that when the render and there for the `mapStateToProps` is executed the state and any props based on it are consistent across the full component tree. Which is awesome!

Only concern that I have with this is performance. That's why in the second commit I add a pure render wrapper component (`PureWrap`) which makes sure that the original component is only rendered when the `mapStateToProps` actually produces a change. Not sure whether this is better or worse than the original. But even if this is not that performant I'd prefer it because it avoids whole class of weird edge cases.

----

Anyone wanting to help and test this I pushed a compiled version to `epeli/react-redux#fix86`. Put that in your package.json as the `react-redux` version number.

### DIFF
--- a/src/components/createConnect.js
+++ b/src/components/createConnect.js
@@ -89,7 +89,30 @@ export default function createConnect(React) {
         };
 
         shouldComponentUpdate(nextProps, nextState) {
-          return !pure || !shallowEqual(this.state.props, nextState.props);
+          if (!pure) {
+            this.updateState(nextProps);
+            return true;
+          }
+
+          const storeChanged = nextState.storeState !== this.state.storeState;
+          const propsChanged = !shallowEqual(nextProps, this.props);
+          let mapStateProducedChange = false;
+          let dispatchPropsChanged = false;
+
+          if (storeChanged || (propsChanged && shouldUpdateStateProps)) {
+            mapStateProducedChange = this.updateStateProps(nextProps);
+          }
+
+          if (propsChanged && shouldUpdateDispatchProps) {
+            dispatchPropsChanged = this.updateDispatchProps(nextProps);
+          }
+
+          if (propsChanged || mapStateProducedChange || dispatchPropsChanged) {
+            this.updateState(nextProps);
+            return true;
+          }
+
+          return false;
         }
 
         constructor(props, context) {
@@ -106,9 +129,8 @@ export default function createConnect(React) {
 
           this.stateProps = computeStateProps(this.store, props);
           this.dispatchProps = computeDispatchProps(this.store, props);
-          this.state = {
-            props: this.computeNextState()
-          };
+          this.state = { storeState: null };
+          this.updateState();
         }
 
         computeNextState(props = this.props) {
@@ -140,12 +162,7 @@ export default function createConnect(React) {
         }
 
         updateState(props = this.props) {
-          const nextState = this.computeNextState(props);
-          if (!shallowEqual(nextState, this.state.props)) {
-            this.setState({
-              props: nextState
-            });
-          }
+          this.nextState = this.computeNextState(props);
         }
 
         isSubscribed() {
@@ -170,20 +187,6 @@ export default function createConnect(React) {
           this.trySubscribe();
         }
 
-        componentWillReceiveProps(nextProps) {
-          if (!shallowEqual(nextProps, this.props)) {
-            if (shouldUpdateStateProps) {
-              this.updateStateProps(nextProps);
-            }
-
-            if (shouldUpdateDispatchProps) {
-              this.updateDispatchProps(nextProps);
-            }
-
-            this.updateState(nextProps);
-          }
-        }
-
         componentWillUnmount() {
           this.tryUnsubscribe();
         }
@@ -193,9 +196,7 @@ export default function createConnect(React) {
             return;
           }
 
-          if (this.updateStateProps()) {
-            this.updateState();
-          }
+          this.setState({storeState: this.store.getState()});
         }
 
         getWrappedInstance() {
@@ -205,7 +206,7 @@ export default function createConnect(React) {
         render() {
           return (
             <WrappedComponent ref='wrappedInstance'
-                              {...this.state.props} />
+                              {...this.nextState} />
           );
         }
       }
--- a/test/components/connect.spec.js
+++ b/test/components/connect.spec.js
@@ -1212,5 +1212,38 @@ describe('React', () => {
       // store.dispatch({ type: 'APPEND', body: 'd'});
       // expect(childMapStateInvokes).toBe(5);
     });
+
+    it('should not render the wrapped component when mapState does not produce change', () => {
+      const store = createStore(stringBuilder);
+      let renderCalls = 0;
+      let mapStateCalls = 0;
+
+      @connect(() => {
+        mapStateCalls++;
+        return {}; // no change!
+      })
+      class Container extends Component {
+        render() {
+          renderCalls++;
+          return <Passthrough {...this.props} />;
+        }
+      }
+
+      TestUtils.renderIntoDocument(
+        <ProviderMock store={store}>
+          <Container />
+        </ProviderMock>
+      );
+
+      expect(renderCalls).toBe(1);
+      expect(mapStateCalls).toBe(2);
+
+      store.dispatch({ type: 'APPEND', body: 'a'});
+
+      // After store a change mapState has been called
+      expect(mapStateCalls).toBe(3);
+      // But render is not because it did not make any actual changes
+      expect(renderCalls).toBe(1);
+    });
   });
 });
--- a/test/components/connect.spec.js
+++ b/test/components/connect.spec.js
@@ -1147,5 +1147,70 @@ describe('React', () => {
       wrapper.setState({ value: 1 });
       expect(target.props.statefulValue).toEqual(1);
     });
+
+    it('should pass state consistently to mapState', () => {
+      const store = createStore(stringBuilder);
+
+      store.dispatch({ type: 'APPEND', body: 'a'});
+      let childMapStateInvokes = 0;
+
+      @connect(state => ({ state }))
+      class Container extends Component {
+
+        emitChange() {
+          store.dispatch({ type: 'APPEND', body: 'b'});
+        }
+
+        render() {
+          return (
+            <div>
+              <button ref="button" onClick={this.emitChange.bind(this)}>change</button>
+              <ChildContainer parentState={this.props.state} />
+            </div>
+          );
+        }
+      }
+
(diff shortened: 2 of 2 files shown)


epeli:
Just realized that iojs is required to run the tests. Fixed all tests but one. Not quite sure yet how to fix that one...

I committed test changes as separate commits with some explanations why I changed those.

gnoff:
@epeli this brings us back to the way things were before #1 which as @gaearon says makes the worst case performance the default. I'm inclined to leave the existing behavior as is. If we did adopt this form I would hope that it be made optional and opt-in much like the 2nd arg props parameter to the first two connect arguments.

epeli:
> this brings us back to the way things were before #1 which as @gaearon says makes the worst case performance the default

Not able apprehend right away every point made in #1 but I'd like hear why this would be the __worst__ case compared to the current implementation because in it `mapState` is also executed always when the store changes. Also note the PureWrap optimization which prevents any useless renders in the wrapped components

epeli:
Also there's another optimization which is not currently implemented in the PR: It's not required to execute mapState at all on the prop changes if it does not take the second parameter (ownProps) because we know that the store hasn't changed as the connect wrappers track themselves store changes.

gnoff:
I could be wrong because I haven't tried running it with your PR but I believe that this change will make even users of the `mapStateToProps` with a single arg call that selector on every re-render of the component, not just when state actually changes. It is true that if you use the props arg for `mapStateToProps` then this behavior already happens but is opt-in and was only added at the request of certain library users, it isn't intended to be the default mode. If your PR is merged then we adopt this more eager mode automatically (I think)

Is that not true?

epeli:
All tests now pass.

> I believe that this change will make even users of the mapStateToProps with a single arg call that selector on every re-render of the component, not just when state actually changes

If I understand you correctly the test `should not invoke mapState when props change if it only has one argument` tests for that and it passes. Funny enough I had to change the `invocationCount`  assert from 2 to 1 which would indicate that this does less work. Also that was not the only place I had to assert for less mapState invokes.

I will do some real world tests later today to see if I missed some regressions. I'd love see other test this as well!


I also added a [new test][1] which makes sure that the wrapped component does not re-render on every state state. Only when the mapState creates a new state.

[1]: https://github.com/rackt/react-redux/commit/02259996ea94888c31b75dbf4a885cab4a753247

epeli:
Wrote a [failing test][1] for the issue at hand and learnt something new about React. React only batches setState calls during event handlers and because this fix relies on the state batching it means that the original issue can still occur when the store is updated from somewhere else. Any ideas how common that is? Could it be possible to somehow force React in to the batching mode when store updates?

In the other news I'm begining to be fairly confident about the performance here. Waiting for feedback.

[1]: https://github.com/epeli/react-redux/commit/683ba87626482381bc5a6c000139c1274b8ac0e8

epeli:
> Could it be possible to somehow force React in to the batching mode when store updates?

To answer myself: Yes with `React.unstable_batchedUpdates(cb)`. Previous discussion: https://github.com/rackt/redux/issues/125

In that issue sebmarkbage [mentions][1] that React has a plan to move batching by default. Can you @sebmarkbage confirm that that's still the case?

If so then that issue will be resolved by itself with React update. Until then [redux-batched-updates][2] middleware can be used as a workaround.

[1]: https://github.com/rackt/redux/issues/125#issuecomment-119381649
[2]: https://github.com/acdlite/redux-batched-updates

gnoff:
@epeli took a better look and you I retract my earlier statements about regressing performance-wise. However regarding the lack of consistent batched updating is it true then that for this PR to not regress in any case one needs to use either the redux-batched-updates middleware or does the lack of that simply return us to the edge-case broken state that we are in with the reported issue #86 ?

epeli:
> is it true then that for this PR to not regress in any case one needs to use either the redux-batched-updates middleware or does the lack of that simply return us to the edge-case broken state that we are in with the reported issue #86 ?

It simply returns to the edge-case broken state. So currently this PR just makes the edge-case less likely and goes completely away with redux-batched-updates middleware.

Incorporating batched updates into Redux was discussed in https://github.com/rackt/redux/issues/125 and it seems that the main reason not to do that was the fact that it was available only as React addon. Could it be considered for react-redux as it's now available in the React object?

epeli:
> as it's now available in the React object

I'll take this back. It was available only in the 0.14 beta. It has been moved to react-dom in 0.14 RC. I guess we don't want to depend on that because react-redux can be used with React native and it makes no sense there.

Nevertheless I don't see any downsides in merging this PR. Without it it is impossible to workaround the issue.

gaearon:
Is it absolutely necessary to have the intermediate component here?

epeli:
Not absolutely necessary. I think the same optimization could be implemented manually in the connector component by invoking the mapState function in shouldComponentUpdate.

gnoff:
My opinion is leave out the PureWrap and let that be a call site optimization. Lots of people connect pure render mixin'd components so I don't think it grants a whole lot in terms of benefit for the extra complexity of code (unwrap function for instance) plus the variable way the connected component would be represented in the component tree (normal connected components will now have 3 tiers intead of just 2 in the devtools explorer)

Neither of these downsides are huge but then neither is the upside so simplicity wins?

gnoff:
> I think the same optimization could be implemented manually in the connector component by invoking the mapState function in shouldComponentUpdate.

most mapState functions will be fast but i wonder if adding the extra call would be worth the mild improvement you get by avoiding some updates. I can imagine that this might improve the already decently fast cases but make the minority of slower cases even slower. just food for thought

epeli:
The devtools argument is really good argument against the PureWrap. I'll remove it on Monday. 

Seems that the batching will become default in React at some point:https://mobile.twitter.com/sebmarkbage/status/642366976824864768?refsrc=email&cn=cmVwbHk%3D

gaearon:
To clarify: I don't promise I'll merge this PR yet :-).
I'm a bit sick now and will review later when I get better.

I definitely want to avoid:

* any performance regressions from current implementation
* adding a component layer (we actually worked to get rid of it before)

epeli:
Currently I don't see any reason why both would not be avoidable.  Perf should be OK already.  I'm now busy for couple days but will work on this after that.

epeli:
I rewrote the entire pull request and force pushed it to this PR branch.

- The change is now much simpler. Should be easier to understand
- All existing tests are now untouched and passing
- Extra component layer (PureWrap) is now gone

As far as I understand the performance should be as good as in the master. Although because there is no performance test suite I cannot be 100% sure.

A new compiled version is again available in `epeli/react-redux#fix86` for easy testing in your apps.

gaearon:
Good job! I've been trying to avoid `setState()` unless needed because, even with `shouldComponentUpdate` returning `false`, it's somewhat of a perf hit in a tight spot. However I agree correctness is more important than performance.

Do you have any idea how we can test perf regression, if any? Maybe we can put this into https://github.com/evancz/todomvc-perf-comparison and hope the benchmark isn't too skewed?

epeli:
> Good job!

Thanks!

> I've been trying to avoid setState() unless needed

Sadly it's exactly `setState()` (and batching) which makes the state consistent.

> Do you have any idea how we can test perf regression, if any?

I think http://benchmarkjs.com/ by the Lo-Dash creator jdalton  is the tool for the job. We've used it for `underscore.string`.

gaearon:
Do you have the time capacity to try to benchmark before/after?

epeli:
So I created #104 for that. Just one simple test to see how store changes perform.

Results of three runs on the current master:

```
dispatch x 86.78 ops/sec ±0.95% (73 runs sampled)
dispatch batched x 135 ops/sec ±0.76% (77 runs sampled)

dispatch x 76.05 ops/sec ±2.02% (64 runs sampled)
dispatch batched x 114 ops/sec ±1.62% (70 runs sampled)

dispatch x 81.47 ops/sec ±2.01% (68 runs sampled)
dispatch batched x 126 ops/sec ±1.32% (78 runs sampled)
```

and with this PR

```
dispatch x 84.80 ops/sec ±1.83% (70 runs sampled)
dispatch batched x 128 ops/sec ±1.63% (77 runs sampled)

dispatch x 84.90 ops/sec ±2.11% (72 runs sampled)
dispatch batched x 132 ops/sec ±1.76% (75 runs sampled)

dispatch x 84.02 ops/sec ±1.72% (69 runs sampled)
dispatch batched x 123 ops/sec ±1.49% (75 runs sampled)
```

Not seeing any significant differences. But this is just a one test.

epeli:
Something still blocking this?

gaearon:
Can you please write release notes and upgrade instructions for this change?
See [Releases](https://github.com/rackt/react-redux/releases) for examples.

From my understanding, this *potentially* can be a breaking change, so we'll jump to 3.0, but we need to tell people *which exactly* patterns would fail which worked before. And of course we need to explain which patterns, previously failing, would work now with this change.

I'll release it as `3.0.0-alpha` today, but it will be labeled as `react-redux@next` on NPM until people give it some usage.

gaearon:
Out as `3.0.0-alpha`, let's wait for feedback.

epeli:
One correction to the release notes:

> This is only relevant if you have nested connect()ed components

It's only relevant if you have nested connect()ed components ***and you use the `ownProps` param in `mapStateToProps` in the nested ones***.

Also I would like to recommend usage of `redux-batched-updates` middleware until React starts defaulting to batching.

gaearon:
I released 3.0.0 but I think I'm misrepresenting the change again.
Can you please look at the release notes: https://github.com/rackt/react-redux/releases/tag/v3.0.0

?

epeli:
The `ownProps` and props from from the parent component are the same thing. So the current description doesn't make much sense. Cannot complain, this change is really hard to explain simply.

I would put it like this:

> Now the map functions (`mapStateToProps`, `mapDispatchToProps` and `mergeProps`) are not called until React starts to render the `connect()`ed components. Previously the map functions where called immediately when store changed which could cause weird edge case bugs when the `ownProps` parameter was a derivative of the state. The state from which it was derivative of was a different version than what was passed as the `state`parameter. In some cases the states can be incompatible with each other and cause very confusing bugs in user code.
>
>Unfortunately the states stay consistent only when store dispatches are called in batches ie. from DOM handlers or manually from `ReactDOM.unstable_batchedUpdates(fn)`. Luckily `redux-batched-updates` middleware can be used to force batching for all dispatches.

epeli:
Oh, a collaborator hat. Updated it.

gaearon:
Thanks!

PR #104, opened 2015-09-14
Add benchmark test suite

epeli:
Sending this mainly as a starting point for discussion on how and what should be benchmarked in Redux & react-redux.

Few points to begin with:

- The production builds should be used for benchmarking
- It should not be required to open devtools when running the bench

### DIFF
--- a/bench.html
+++ b/bench.html
@@ -1,0 +1,18 @@
+<!doctype html>
+<html>
+    <head>
+        <meta charset="utf-8">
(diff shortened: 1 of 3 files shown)


gaearon:
Wow, thank you! It's a great start.

timdorr:
@epeli I know it's been a while, but any progress on this?

epeli:
Unfortunately no. Better close this for now.

PR #105, opened 2015-09-14
Static methods using @connect

callum:
This isn't currently possible:

```js
@connect()
class Foo {
  static bar () {
    return 'baz'
  }

  quux () {
    return Foo.bar() // doesn't work
  }
}
```

```js
// current behaviour
import WrappedFoo from './Foo'
WrappedFoo.WrappedComponent.bar() // 'baz'

// desired behaviour
import Foo from './Foo'
Foo.bar() // 'baz'
```

It's worth noting that there are potential side effects to this change. A given component might define lifecycle methods that aren't present on the connect component, thus altering its behaviour. Consider this by no means ready to go - just want to open up the discussion to explore a solution.

### DIFF
--- a/src/components/createConnect.js
+++ b/src/components/createConnect.js
@@ -76,7 +76,7 @@ export default function createConnect(React) {
     }
 
     return function wrapWithConnect(WrappedComponent) {
-      class Connect extends Component {
+      class Connect extends WrappedComponent {
         static displayName = `Connect(${getDisplayName(WrappedComponent)})`;
         static WrappedComponent = WrappedComponent;
 


gaearon:
I'd rather prefer to do nothing about statics. Start copying methods, and people will want us to copy properties, then we'll have to blacklist *some* properties (e.g. `contextTypes`), then it's just crazy amount of code and edge cases to maintain.

If you'd like to call statics, don't use decorator:

```js
class Foo {
  static bar () {
    return 'baz'
  }

  quux () {
    return Foo.bar() // works
  }
}

const ConnectedFoo = connect()(Foo);
export default ConnectedFoo;
```

cesarandreu:
+1 to keeping things simpler

Perhaps it'd be a good idea to explicitly mention that in the Support for Decorators section? @gaearon

gaearon:
>Perhaps it'd be a good idea to explicitly mention that in the Support for Decorators section?

I'd say it's not even related to decorators. Maybe "My static and instance methods don't work!" in Troubleshooting?

namuol:
Shouldn't decorator functions follow the [canonical definition](https://en.wikipedia.org/wiki/Decorator_pattern) in the sense that they merely add new behavior?

Maybe we should call these something other than "decorator functions", because this is a point of confusion I've seen a lot in the React community, lately.

blakeembrey:
Using higher order components as a decorator function is confusing - especially to new people in JavaScript or from other languages. It's easy enough for anyone else to say, well, just work around it by doing x or y. To say this though, is to ruining the use of decorators in the future by breaking the implicit contract defined by decorators. In a strongly typed environment, this would not work. We're just using it as a single sugar over `x(class {})` here because "it looks better".

gaearon:
Well, just don't use decorators for HOCs like `connect()` then. :-)
We don't encourage this in the docs.

blakeembrey:
@gaearon You're right and it was nothing personal. I could have sworn all your examples used to use it, but that was a few months back - at least I remember going through something which had it. Good to see it's not the case anymore :+1:

namuol:
True, it's not *explicitly* encouraged, but perhaps [this section](https://github.com/rackt/react-redux#support-for-decorators) in the readme might caution that `connect()` is not a canonical decorator function.

I'm not trying to pick on Redux, specifically -- [I see this anti-pattern *everywhere* right now](https://twitter.com/louroboros/status/643200044213137410) -- but since Redux is becoming widely adopted by the community, this could be an opportunity to establish a better conventional understanding of the decorator pattern, which should not be confused with HOCs.

gaearon:
Doc PR welcome!

PR #106, opened 2015-09-15
Add troubleshooting entry about static properties

cesarandreu:
Not sure about the wording... But it's a start :D. 

Relates to #105

### DIFF
--- a/README.md
+++ b/README.md
@@ -502,6 +502,44 @@ If you’re using React for web, this usually means you have a [duplicate React]
 
 If you’re using React Native, make sure you’re importing `react-redux/native` both for `<Provider>` and any `connect()` call. Importing from `react-redux` will not work on React Native.
 
+### My static and instance methods don't work!
+
+When you use `connect()`, it wraps your component and returns a new component. Upon doing this, the new component will be unable to access the original component's static properties or methods.
+
+You may avoid this by not using decorators and using separate variables for the component and the connected component.
+
+**Good**
+
+```jsx
+class Foo {
+  static bar () {
+    return 'baz';
+  }
+
+  quux () {
+    return Foo.bar(); // works
+  }
+}
+
+const ConnectedFoo = connect()(Foo);
+export default ConnectedFoo;
+```
+
+**Bad**
+
+```
+@connect()
+export default class Foo {
+  static bar () {
+    return 'baz';
+  }
+
+  quux () {
+    return Foo.bar(); // doesn't work
+  }
+}
+```
+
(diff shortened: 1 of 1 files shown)


patrick-jones:
You might want to mention some of the "magic" statics for React components also. I found this issue because I needed to add something to the child context, and I was losing `childContextTypes` when using the decorator version of `connect`. My solution was:

```javascript
class _Application extends React.Component {
    getChildContext() {
        return {
            muiTheme: this.props.theme
        };
    }

    render () {
        return (
            <Provider store={ this.props.store }>
                { () => <PageRoot /> }
            </Provider>
        )
    }
}

// must do this on the original class or you get invariant errors
_Application.childContextTypes = {
    muiTheme: React.PropTypes.object
};

let Application = connect()(_Application);

export default Application;
```

gaearon:
I wouldn't call these "good" and "bad", these are just situations where particular things work or don't work. Also, a bigger issue that needs to be clarified is that *the consumer outside the file* will need to call `Foo.WrappedComponent.someStaticMethod()`, or you'd need to put `someStaticMethod` onto the wrapper:

```js
@connect()
class Stuff { }

Stuff.someStaticMethod = ... // can be called from outside directly
```

namuol:
You may also want to mention that public "instance" methods must be copied onto the `prototype` of `connect()`-wrapped components:

```js
class Stuff {
  foo() { return 'bar'; }
}

const WrappedStuff = connect()(Stuff);
WrappedStuff.prototype.foo = Stuff.prototype.foo;

export default WrappedStuff;
```

Similarly, public methods/properties that *aren't* defined on `Foo.prototype` will become inaccessible:

```js
class Stuff {
  constructor() {
    this.foo = () => { return 'bar'; };
  }
}

const WrappedStuff = connect()(Stuff);

let stuff = new WrappedStuff();
stuff.bar; // undefined
```

I'm not sure if these are common gotchas, but I figured it was worth making a note here. :beers:

epeli:
FYI Fluxible uses [hoist-non-react-statics](https://www.npmjs.com/package/hoist-non-react-statics) to  copy static methods to the [wrapper](https://github.com/yahoo/fluxible-addons-react/blob/16cc68846f96a7e0ae70a7d1ba293bee5b347ea2/connectToStores.js#L49). Not sure that's a good idea but something to consider.

gaearon:
I'm closing as there's been no response to my comments above, and we actually *do* proxy static methods now thanks to `hoist-non-react-statics`.

PR #107, opened 2015-09-15
Make main readme/docs less intimidating 

danmartinez101:


danmartinez101:
I can do more here if you would like

gaearon:
Been meaning to do this.. Thanks!

PR #117, opened 2015-09-24
Clarification of ownProps value in connect callback

grrowl:
> If `ownProps` is specified as a second argument, then `mapStateToProps` will be re-invoked whenever the component receives new props.

changed to:

> If `ownProps` is specified as a second argument, its value will be the properties passed to your component, and `mapStateToProps` will be re-invoked whenever the component receives new props.

Happy for alternatives, just figured on first glance that ownProps changed the behaviour of the callback, but it doesn't mention (so I didn't think) what the value would be (also non-obvious who's "own" props they are), so thought it could use a visit.


gaearon:
That's better, thanks!

PR #127, opened 2015-09-28
Allow connect() to hoist non-react statics from wrapped component

### LINKED ISSUE #53: Maintain static functions
Could `connect()` check its wrapped component for static functions and add them to the returned component? For things like `fetchData()` and `onEnter()`. Seems like it would be pretty easy to go through the keys and check if `typeof WrappedComponent[key] === 'function'`.

Just a thought.

erikras:
* Adds dependency on [`hoist-non-react-statics`](https://github.com/mridgway/hoist-non-react-statics)
* Fixes #53

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,6 +58,7 @@
     "webpack": "^1.11.0"
   },
   "dependencies": {
+    "hoist-non-react-statics": "^1.0.3",
     "invariant": "^2.0.0"
   },
   "peerDependencies": {
--- a/src/components/createConnect.js
+++ b/src/components/createConnect.js
@@ -2,6 +2,7 @@ import createStoreShape from '../utils/createStoreShape';
 import shallowEqual from '../utils/shallowEqual';
 import isPlainObject from '../utils/isPlainObject';
 import wrapActionCreators from '../utils/wrapActionCreators';
+import hoistStatics from 'hoist-non-react-statics';
 import invariant from 'invariant';
 
 const defaultMapStateToProps = () => ({});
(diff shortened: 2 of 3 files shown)


gaearon:
Out in 3.1.0.

erikras:
Hmm... The way greenkeeper references each PR with a hyperlink might get old. Do we really need to know about every public repo that merged this?

gaearon:
I don't really mind. The references are because it mentions release notes which link to this PR.

erikras:
Yes, that's what I meant. It uses the commit text, which should reference issues and PRs it fixes/merges.

I guess as long as it doesn't generate emails. :smile:

CrocoDillon:
This is sweet, but how do you get redux' state and dispatch in the static methods?

gaearon:
It's better to create a new issue than post in a not quite related one. Static methods don't have dispatch because it only exists in a specific rendering context. For example in a server rendered app you would render components with a different store instance on every request. This is why it doesn't make sense for static methods to have access to the dispatch function.

What I think you can do is define `createRoutes(store)`. It would use the store to provide `dispatch` to the hook: `onEnter={(...args) => MyComponent.onEnter(...args, store.dispatch)}`.

Then you would call `createRoutes(store)` when rendering with your store instance. The component would receive `dispatch` as the last argument.

CrocoDillon:
Sorry about that, found this PR after some searching and thought it was related enough since the hoisting is exactly what I was looking for. And yeah it did feel dirty abusing a static method like that, anyway I thought the method was hoisted to the connected component instance (and as such no longer static). I looked wrong though, it's still static so you're right... what I wanted doesn't make any sense.

The `createRoutes(store)` seems like the way to go! And it's really awesome that you respond so quickly :D Thanks a lot!

PR #138, opened 2015-10-07
Externalize getDisplayName since its a common concern for most higher order components?

jurassix:
PR to externalize `getDisplayName` into a sharable utility. This is a common function found in many Higher Order Components in the wild. 

[This external dependency is fully tested, etc. You can see the minimal source here.](https://github.com/jurassix/react-display-name)

Not sure if react-redux interested in an extra dependency; issuing PR for optional consideration.


jurassix:
Or another alternative would be to incorporate [recompose](https://github.com/acdlite/recompose/blob/master/src/getDisplayName.js) utilities here as well?

acdlite:
I'm cool with using Recompose's `wrapDisplayName()` — true story, I copied the code for `getDisplayName()` from this repo — though I know Dan likes to keep the dependencies to a minimum. 

@gaearon?

gaearon:
I'm OK with using `recompose` if it doesn't sufficiently increase UMD build size.
Mind checking?

```
npm i
npm run build
gzip -c dist/react-redux.min.js | wc -c
```

jurassix:
Updated PR to use recompose.

> gzip -c dist/react-redux.min.js | wc -c
> 11283

gaearon:
This way you're bundling whole recompose for one method.
The build size increased 4x, it was 2867 before the change.

@acdlite Do you support importing individual files?

acdlite:
`import recompose/wrapDisplayName from 'recompose'`

jurassix:
updated import:

```js
import wrapDisplayName from 'recompose/wrapDisplayName';
```

Output seems reasonable now:

> gzip -c dist/react-redux.min.js | wc -c
> 2917

acdlite:
Hrm, still seems a bit high for such a simple util. Probably the Babel helpers... I'll see if I can get it down a bit.

acdlite:
^ also lol at my comment above, obviously that import is wrong

gaearon:
Meh, let's keep it internal for now. If we depend on a few more helpers, might be worth revisiting.

PR #142, opened 2015-10-08
Add docs for WrappedComponent and getWrappedInstance

BinaryMuse:
This information was included on the web site, but is missing from the API docs.


gaearon:
Good point. I'll hold off for now because we might have problems with stateless components and `getWrappedInstance()` and the API might change there.

gaearon:
I'll close this out because I'm working on similar updates to the docs concerning changes for 0.14.

Issue #141, opened 2015-10-08
Connect's wrapped component can't be a stateless function


dariocravero:
If the component wrapped with `connect` is a stateless function component, React warns about the [internal ref](https://github.com/rackt/react-redux/blob/master/src/components/createConnect.js#L203) that the wrapper holds with the following message `Warning: Stateless function components cannot be given refs (See ref "wrappedInstance" in App created by Connect(App)). Attempts to access this ref will fail.`

```
import { connect } from 'react-redux';
import React from 'react';

export default connect(mapStateToProps)(props => <div />);
```

Would it be useful to check whether the component is stateless or not to decide when to apply the ref or do we always need it?

troch:
You can suppress that warning by using [troch/react-stateless](/troch/react-stateless) as a workaround for now.

What is the wrappedInstance ref used for?

gaearon:
People use it for tests and side effects (i.e. calling instance methods) I presume.
Personally I don't have use for it, but it was repeatedly asked.

hex13:
So is it okay to use stateless functions with React-Redux and I can ignore this warning? (I do not use ref or wrappedInstance).

gaearon:
Yes, you can ignore the warning for now.

hex13:
Ok, thx :)

gaearon:
This is fixed via https://github.com/rackt/react-redux/commit/2d3d0beade55477b3af65534ceb793db18b25705.
We're making `withRef` an opt-in parameter, with `false` by default.

dariocravero:
Thanks! :)

mini-eggs:
I ran into something that may be related.

While using connect with a stateless component that had a child component with state (extends React.Component): the child component's `componentWillReceiveProps` would never fire on prop change. Instead the child component with state would rerender and it's `componentDidMount` would fire again. I ran into this within a create-react-native-app project.

If this is indeed an issue I would be happy to look into it and create a PR, I could also start a branch on the project to easily reproduce this (I've since removed the stateless component with connect).

markerikson:
@mini-eggs : hmm.  If it's something you can reproduce, please go ahead and file a new issue.

rafalradomski:
I using react-redux 5.0.6 and I still getting this warning: 

``Warning: Stateless function components cannot be given refs. Attempts to access this ref will fail.``

PR #143, opened 2015-10-08
connect doesn't apply ref to stateless component

bspaulding:
Took a stab at #141 

Probably would be better to create the render function beforehand, instead of hitting that conditional every render. Just getting this out there for discussion.

Worth noting that the detection isn't perfect, but solves the case mentioned here, and keeps the current test for wrappedInstance ref passing. We'd have to instantiate the component to know for sure, but this might be good enough.

Pretty much stole the check from ReactCompositeComponent

### DIFF
--- a/src/components/createConnect.js
+++ b/src/components/createConnect.js
@@ -77,6 +77,8 @@ export default function createConnect(React) {
     }
 
     return function wrapWithConnect(WrappedComponent) {
+      let isStateless = !(WrappedComponent.prototype && WrappedComponent.prototype.isReactComponent) || !('prototype' in WrappedComponent);
+
       class Connect extends Component {
 
         shouldComponentUpdate(nextProps, nextState) {
@@ -199,6 +201,10 @@ export default function createConnect(React) {
         }
 
         render() {
+          if (isStateless) {
+            return <WrappedComponent {...this.nextState} />;
+          }
+
           return (
             <WrappedComponent ref='wrappedInstance'
                               {...this.nextState} />
--- a/test/components/connect.spec.js
+++ b/test/components/connect.spec.js
@@ -1117,6 +1117,21 @@ describe('React', () => {
       expect(decorated.refs.wrappedInstance.someInstanceMethod()).toBe(someData);
     });
 
+    it('should not add refs to wrapped instance if stateless', () => {
+      const store = createStore(() => ({}));
+      const decorator = connect(state => state);
+      const Decorated = decorator(() => <div/>);
+
+      const tree = TestUtils.renderIntoDocument(
+        <ProviderMock store={store}>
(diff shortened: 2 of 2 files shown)


src/components/createConnect.js
```diff
@@ -199,6 +201,10 @@ export default function createConnect(React) {
         }
 
         render() {
+          if (isStateless) {
+            return <WrappedComponent {...this.nextState} />;
+          }
+
```
bspaulding: This should probably not be here. If we decide to go ahead with this, we should probably create render above and apply it to this class.

gaearon:
Wouldn't this give false positives for module pattern of non-Component-inheriting classes which are still valid in 0.14?

bspaulding:
Yes, indeed. I'm not sure at this point how you'd figure that out before mounting the component. I'll keep hacking on it.

test/components/connect.spec.js
```diff
@@ -1117,6 +1117,45 @@ describe('React', () => {
...
+
+    it('should consider module pattern of non-Component-inheriting classes stateful', () => {
+      function MyComponent(initialProps) {
+        return {
+          state: { value: initialProps.initialValue },
+          render: function render() {
+            return <span className={this.state.value} />;
+          }
+        };
+      }
```
bspaulding: I assume this is the case you're talking about?

bspaulding: Hmm, interestingly, this turns out to not be a false positive. At least, React thinks this is stateless, and shows the warning.

bspaulding: Sorry, I got confused. This spec is bad, wrapping it in a function below.

src/components/createConnect.js
```diff
@@ -77,6 +77,8 @@ export default function createConnect(React) {
     }
 
     return function wrapWithConnect(WrappedComponent) {
+      let isStateless = !(WrappedComponent.prototype && WrappedComponent.prototype.isReactComponent) || !('prototype' in WrappedComponent);
```
lukewestby: I'm not sure if this is a major concern, but this method of detecting stateful components isn't compatible with React < 0.14. Stateful components created in React 0.13 will be flagged as stateless as well. We had a similar situation at https://github.com/gajus/react-css-modules/issues/40

gaearon: Yes, we won't get this in before going 0.14+ either way.

lukewestby: Sounds good, thanks @gaearon

gaearon:
Thanks for the PR.
In the end I decided on a simpler approach: refs are now opt-in at the `connect()` call site.

Closed via https://github.com/rackt/react-redux/commit/2d3d0beade55477b3af65534ceb793db18b25705.

PR #146, opened 2015-10-09
Switched from React to ReactDOM when calling 'findDOMNode'

toxickg:
It applies to a single test case only. :smirk:


gaearon:
Actually just

```
const node = container.getWrappedInstance().refs.button;
```

should be enough. DOM refs *are* nodes in 0.14.

toxickg:
Sure! Changed it.

PR #149, opened 2015-10-09
Make shouldComponentUpdate less aggressive

### LINKED ISSUE #144: Connected component prevents all context propagations
Because connected components have aggressive `shouldComponentUpdate`, and `shouldComponentUpdate` doesn't support context, connected components basically stop any other context variables changes to propagate to their children.

Is there a recommended solution to this problem? I found `pure` option (`connect(..., {pure: false})`), which helps, but is there a better way?

rpominov:
This should fix #144 

One test still failing, but I can see how to fix that if this approach make sense.

### DIFF
--- a/src/components/createConnect.js
+++ b/src/components/createConnect.js
@@ -87,8 +87,11 @@ export default function createConnect(React) {
             return true;
           }
 
+          const stateChanged = nextState !== this.state;
           const propsChanged = !shallowEqual(nextProps, this.props);
-          if (this.state === nextState && !propsChanged) {
+          if (!stateChanged && !propsChanged) {
+            // No reason to reconcile but we're in shouldComponentUpdate():
+            // This is probably due to a context change so we let the update happen.
             return true;
           }
 
--- a/src/components/createConnect.js
+++ b/src/components/createConnect.js
@@ -87,8 +87,12 @@ export default function createConnect(React) {
             return true;
           }
 
-          const storeChanged = nextState.storeState !== this.state.storeState;
           const propsChanged = !shallowEqual(nextProps, this.props);
+          if (this.state === nextState && !propsChanged) {
+            return true;
+          }
+
+          const storeChanged = nextState.storeState !== this.state.storeState;
           let mapStateProducedChange = false;
           let dispatchPropsChanged = false;
 
--- a/test/components/connect.spec.js
+++ b/test/components/connect.spec.js
@@ -40,7 +40,7 @@ describe('React', () => {
 
       const store = createStore(() => ({}));
       const contextTypes = {
-        foo: PropTypes.string.isRequred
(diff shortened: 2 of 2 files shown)


gaearon:
How does this work? It isn't obvious to me how state comparison solves the context issue. Can you elaborate?

rpominov:
It basically checks what caused the update. If the update was caused by a context change or some .forceUpdate() call, `state` and `props` will remain the same. So the logic is: if React wants to make and update, and we don't know the reason, just make the update happen.

gaearon:
This makes sense.
I'd just change it to be written a bit differently:

```js
          const stateChanged = nextState !== this.state;
          const propsChanged = !shallowEqual(nextProps, this.props);
          if (!stateChanged && !propsChanged) {
            // No reason to reconcile but we're in shouldComponentUpdate():
            // This is probably due to a context change so we let the update happen.
            return true;
          }

          const storeStateChanged = ...
```

gaearon:
Hmm.. Wouldn't non-essential prop changes then cause a re-render? e.g. `setState()` in parent component passing the same values as props.

rpominov:
> Hmm.. Wouldn't non-essential prop changes then cause a re-render? e.g. setState() in parent component passing the same values as props.

Probably yes. Although this can be handled by the parent component's `shouldComponentUpdate`. This is a trade-off anyway, but I think it better than too aggressive shouldComponentUpdate.

rpominov:
> Hmm.. Wouldn't non-essential prop changes then cause a re-render? e.g. setState() in parent component passing the same values as props.

And this is why the remaining test fails :)

gaearon:
>Although this can be handled by the parent component's shouldComponentUpdate. This is a trade-off anyway, but I think this is better than a too aggressive shouldComponentUpdate.

This was the point of having `shouldComponentUpdate()` on `connect()`. Otherwise it doesn't really give you anything, as prop changes will be common, especially with nested `connect()`s.

If there's no easy fix, I think `{ pure: false}` is the correct solution until React context problem is fixed.

rpominov:
Ok, maybe we could put the `if (!stateChanged && !propsChanged)` check inside `if (!pure) {...}` then? So when `{pure: false}` mode is used we still do less unnecessary work. Or maybe we can have 3 modes?

gaearon:
`{ pure: false }` should disable the check completely. It should update even if you deeply mutate the props—just like React's default `shouldComponentUpdate()` works out of the box.

I don't really want to support three different modes. :-( If you really need this I think you're better off writing your own version of `connect()` that satisfies your needs.

rpominov:
Okay. All this made me think that I probably should not use context the way I use it anyway. Otherwise I'll have have the same issue with my own `shouldCompoentUpdate` methods.

PR #151, opened 2015-10-11
Add option to turn connect refs on and off

troch:
An alternative way to deal with #141

Add prop `connectOptions` to Provider (put in context). Default connect options are `{useRefs: true}`, and options can be overwritten on each connect instance (so refs can be turned on or off globally and locally). If refs are used for testing purposes only, then it gives the ability to not have them in production.

`useRefs` is by default true to avoid introducing a breaking change, but IMO it would be best if it is false by default (and turned on for testing and debugging purposes).

### DIFF
--- a/src/components/createConnect.js
+++ b/src/components/createConnect.js
@@ -112,6 +112,7 @@ export default function createConnect(React) {
           super(props, context);
           this.version = version;
           this.store = props.store || context.store;
+          this.connectOptions = {...context.connectOptions, ...props.connectOptions};
 
           invariant(this.store,
             `Could not find "store" in either the context or ` +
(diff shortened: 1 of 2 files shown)


gaearon:
This still doesn't feel like a right solution.

troch:
I cannot think of another way right now if detecting stateless functional components is not bullet proof :persevere:

gaearon:
I don't think we need to detect them. Maybe we should instead offer some other API for accessing refs. For example, we could accept `instanceRef` prop that would act as a callback ref. If user doesn't specify it, it would be undefined.

gaearon:
Thanks for the PR.
In the end I decided on a simpler approach: refs are now opt-in at the `connect()` call site.

Closed via https://github.com/rackt/react-redux/commit/2d3d0beade55477b3af65534ceb793db18b25705.

troch:
:+1:

PR #158, opened 2015-10-15
Drop dependency on mocha-jsdom in favor of explicit solution

gaearon:
Instead of doing something mysterious with `mocha-jsdom` we now do as suggested in https://github.com/facebook/react/issues/5046#issuecomment-146222515.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "clean": "rimraf lib dist coverage",
     "lint": "eslint src test",
     "prepublish": "npm run clean && npm run build",
-    "test": "mocha --compilers js:babel/register --recursive",
+    "test": "mocha --compilers js:babel/register --recursive --require ./test/setup.js",
     "test:watch": "npm test -- --watch",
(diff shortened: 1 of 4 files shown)


zpratt:
@gaearon: I have been using a similar solution to this for my tests for quite some time. Recently we have switched to using the [shallow renderer](https://facebook.github.io/react/docs/test-utils.html#shallow-rendering), which has allowed us to remove jsdom from our tests completely. This has given our tests a significant speed boost. Shallow rendering also forces you to keep each component relatively simple since it prevents you from testing heavily nested components. 

Just thought I'd share a tip on something that we've found to be valuable.

gaearon:
We are using shallow rendering too, take a closer look :-) The only reason we use jsdom is to prevent React from blowing up with errors when using `setState` with shallow renderer. (#4019) I have heard people say that actually shimming `global.document = {}` is enough for React to work properly. If that's really so, we can drop jsdom from instructions.

PR #175, opened 2015-11-05
Use stateless functional component in dumb example

sirreal:
"Dumb" components like this example seem to be an obvious place to use React 0.14 stateless functional components. This PR changes the example to use one.

I don't know if a stateful component was a concious choice, but I would prefer stateless in this situation and I think it should be considered for the example.


omnidan:
@sirreal thank you :grin: It probably just hasn't been updated to 0.14 yet.

PR #180, opened 2015-11-09
Adds memoize option, defaulting to no memoization

tgriesser:
So ultimate goal of #179 was to be able to add a layer of memoization on the selector - this PR does just that, if there is a function provided as a `memoize` option, the selector is passed through that function at instantiation and attached to the component instance. If `memoize` option is not provided ~~the default is identity~~ the function is attached directly to the component.

I moved a bit of the construction of the class to `componentWillMount` so we have a method to call in `componentWillUpdate` to re-attach the functions on hot reloading.


tgriesser:
Actually, this approach doesn't work either because it uses the same function for both mapState and mapDispatch, reopening #179.

PR #182, opened 2015-11-11
createConnect, with configurable shallowEqual

tgriesser:
Similar to `createSelectorCreator` in reselect, provides `createConnect` to optionally wrap `connect` providing a custom implementation of `shallowEqual`. 

@ellbee thoughts? Are there any other functions that might be worth allowing customization of?


gaearon:
Seems like an overkill to add new top-level API when we have `options`.
Why not let user specify `options.arePropsEqual`?

tgriesser:
Ah yep that'd work too, I thought maybe we were looking to avoid new options :)

Will reopen with that.

gaearon:
New options are better than new top-level APIs :wink:

PR #185, opened 2015-11-11
connectComponent, attaching prop configuration to component

tgriesser:
Alternate approach to #179 for consideration, adds a ~~`connectFactory`~~ `connectComponent` which `connect` will wrap,  allowing for unique selectors per-component instance. Existing use of connect will work the same as it does currently.


gaearon:
What if instead we checked the return type of `mapStateTo*` functions when they are invoked for the first time. If the return type is a function rather than an object, assume we are dealing with a factory. This would be backwards compatible and wouldn't require a new top-level export. Am I missing something?

ellbee:
Oh, good shout. I don't think you are missing anything. I just hacked together a quick test for mapStateToProps, and it seems like it would be a reasonable solution.

gaearon:
@tgriesser Would you be interested in implementing https://github.com/rackt/react-redux/pull/185#issuecomment-175832321 as an alternative approach? Would it solve your problem?

gaearon:
I’ll close this PR because I don’t plan to merge it as is but I’m happy to consider merging https://github.com/rackt/react-redux/pull/185#issuecomment-175832321 which should solve the same problem.

PR #189, opened 2015-11-13
fix length comparisons on xxxToProps functions

planetcohen:
Testing finalMap{State|Dispatch}ToProps.length > 1 breaks when these functions are wrapped in generic decorator functions. Instead, test for !== 1.

For example, let's say I wanted to write a decorator function that logs function invocation like so:
```
function traceExecution(fn) {
  const name = fn.name;
  return function() {
    console.log('ENTER: ', name);
    const result = fn.apply(this, arguments);
    console.log('LEAVE: ', name, ' --> ', result);
    return result;
  }
}
```
Then, I want to wrap my mapStateToProps and mapDispatchToProps functions with traceExecution and pass it to connect:
```
function mapStateToProps(state, props) { /* stuff here */}
function mapDispatchToProps(dispatch, props) { /* stuff here */}
const connector = connect(traceExecution(mapStateToProps), traceExecution(mapDispatchToProps))
```
This would not work without these fixes, since the functions returned by the decorator have arity 0 and would fail the "> 1" length test.

By changing the arity test to be "!== 1" instead of "> 1", you enable this scenario to work as expected.

### DIFF
--- a/src/components/connect.js
+++ b/src/components/connect.js
@@ -28,8 +28,8 @@ export default function connect(mapStateToProps, mapDispatchToProps, mergeProps,
     wrapActionCreators(mapDispatchToProps) :
     mapDispatchToProps || defaultMapDispatchToProps
   const finalMergeProps = mergeProps || defaultMergeProps
-  const shouldUpdateStateProps = finalMapStateToProps.length > 1
-  const shouldUpdateDispatchProps = finalMapDispatchToProps.length > 1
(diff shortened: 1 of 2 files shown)


gaearon:
Fair enough. Mind adding tests for this?

planetcohen:
Done - I added tests checking for expected behavior when arity is zero.

gaearon:
Thanks!

PR #197, opened 2015-11-20
Empty commit to test Travis failing

ryanbrainard:
This is an empty commit based on `master` to test if `eslint` is failing on `master`. I have seen it failing on branches with other code changes, but I believe it is failing due to an external change and want to see if it also failing on `master`. I'm not a contributor, so can't simply click the "Restart Build" button in Travis for `rackt/react-redux`, so kicking off a build this way.


ryanbrainard:
Yes, it is failing. Strangely also failing on this even though I didn't change anything:

```
1) React connect should throw an error if the store is not in the props or context:
     Error: Expected [Function] to throw 'Invariant Violation: Could not find "store" in either the context or props of "Connect(Container)". Either wrap the root component in a <Provider>, or explicitly pass "store" as a prop to "Connect(Container)".'
      at Object.assert [as default] (node_modules/expect/lib/assert.js:20:9)
      at Expectation.toThrow (node_modules/expect/lib/Expectation.js:90:24)
      at Context.<anonymous> (test/components/connect.spec.js:1103:9)
```

Going to close this pull and open an issue.

PR #204, opened 2015-11-27
Fix CI build errors

ellbee:
Bumped node version to 5 to fix lint error.
Made failing test 'should throw an error if the store is not in the props or context' not have to match the exact error string as suggested [here](https://github.com/rackt/react-redux/pull/196#issuecomment-160009368) (thanks @schmeedy!)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-  - "iojs-2"
+  - "5"
 script:
   - npm run lint
   - npm test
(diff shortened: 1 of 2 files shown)


gaearon:
React probably got rid of `Invariant Violation: ` recently as pretext for the message. It was moved to be `error.name` if I recall correctly. This may be the reason test is failing.

ellbee:
Yep, I don't think its good to be relying on the *exact* string.

conorhastings:
@gaearon , yeah they did, pretty recently, about a month ago - https://github.com/facebook/react/pull/5185

PR #208, opened 2015-12-03
Return a promise from handleChange

mhodgson:
Resolves when React rendering is complete. See https://github.com/rackt/redux/pull/1096 for related change to Redux core.

Enables callers of dispatch to run a callback when all React rendering has been completed.

Thoughts?

### DIFF
--- a/src/components/connect.js
+++ b/src/components/connect.js
@@ -185,7 +185,7 @@ export default function connect(mapStateToProps, mapDispatchToProps, mergeProps,
           return
         }
 
-        return new Promise((resolve, reject) => {
+        return new Promise((resolve) => {
           this.setState({
             storeState: this.store.getState()
           }, resolve)
--- a/src/components/connect.js
+++ b/src/components/connect.js
@@ -185,8 +185,10 @@ export default function connect(mapStateToProps, mapDispatchToProps, mergeProps,
           return
         }
 
-        this.setState({
-          storeState: this.store.getState()
+        return new Promise((resolve, reject) => {
+          this.setState({
+            storeState: this.store.getState()
+          }, resolve)
         })
       }
 


mhodgson:
Obviously can update tests and docs if no one is opposed to this.

gaearon:
Two thoughts:

1. We can't rely on `Promise` being available in older browsers and we don't want to rely on polyfill.
2. Can you show how you would use this?

mhodgson:
@gaearon The macro problem I am trying to solve is simply knowing when react is finished flushing changes to the DOM. Unfortunately the only affordance to do that is through a callback in ever call to `setState`. I was hoping it would be ok to try to thread that through redux, but given that you rejected to the necessary changes to Redux core I don’t think this PR will help at all.

mhodgson:
For the purpose of discussion, here’s a more detailed explanation of why this is necessary in some use cases: 

Assume that you are trying to create a WYSIWYG text editor for the browser (fairly common and painful endeavor). The state at any given moment is primarily composed of two things: 

1. The document itself (the thing the user is creating).
2. The current text selection within the document.

At first glance you might wonder why it is important to keep the text selection in state. The answer is simple: undo/redo. When a user triggers an undo or redo action, we not only need to restore the document to the previous state, we also need to restore the text selection to the previous state. 

It is important in this case that we fully restore the document state to the DOM __before__ trying to restore the text selection. It should be obvious that trying to restore a selection of something that doesn’t yet exist on the page is problematic. Hence the need to know when the DOM has been fully rendered within the normal action emit/listener/render loop provided by Redux. After the DOM has been fully rendered, we can do the manual task of restoring the selection state on top of the DOM.

As far as I can tell, the current Redux and `connect` implementations do not support this use case. I can’t think of another use case that would require this off the top of my head, but I assume that the callback in React’s `setState` method was added for these types of rare but important edge cases. 

@gaearon do you see a different approach to this that would still allow for using redux and react-redux? Are there other scenarios in React or React-Native or even angular which might require knowing when changes to state have been fully flushed to the UI?

gaearon:
I understand your use case. I don't want to complicate `connect()` with this but I'm up for extracting the heavy lifting from `connect()` into a separate function that would make it much easier to build your custom `connect()`. If you're interested in proposing an API, let's discuss in https://github.com/rackt/react-redux/issues/269#issuecomment-175920073.

Issue #206, opened 2015-12-02
[REQ] Babel update


NullDivision:
Would it be possible to update Babel to version 6? Version 5 interferes with newly created projects.

duro:
I'm running into this as well in a react-native 0.16 project which uses Babel 6 under the hood. Looks like the .babelrc in react-redux throws an `unknown option` error.

ajwhite:
I'm unable to build with the changes made to `stage` in 6, I believe

```
{
  "message": "TransformError: /Users/ajwhite/Development/OneMightyRoar/Robin/robin-rooms-react-0.16/node_modules/react-redux/native.js: [BABEL] /Users/ajwhite/Development/OneMightyRoar/Robin/robin-rooms-react-0.16/node_modules/react-redux/native.js: Unknown option: /Users/ajwhite/Development/OneMightyRoar/Robin/robin-rooms-react-0.16/node_modules/react-redux/.babelrc.stage",
  "type": "TransformError",
  "errors": [{}]
}
```

NullDivision:
As a workaround for the `unknown option` issue was to install babel-core with the presets in my project and babel-cli globally. As long as there's no babel executable in the `./node_modules/.bin/` directory it installs its own.

ajwhite:
Interesting, @NullDivision -- so you basically kill the local installations and rely on the global? In theory, it sounds like this might also let you compile with a different version of Babel? I'm currently using the `@connect` decorators that are no longer supported in Babel 6 (and also not resolved with the current workaround in [babel-plugin-transform-decorators-legacy](https://github.com/loganfsmyth/babel-plugin-transform-decorators-legacy))

NullDivision:
I'm not sure about the `@connect` decorators but all my dependencies work just fine with the global installation of babel-cli. For reference I'm using Webpack as my main builder and everything works as it should.

I need `babel-core/register` for runtime transpiling on Karma and Gulp but everything else uses the global binary.

eadz:
I believe this is also causing an error with the react-native compatible branch, as the latest RN is using version 6 of babel.

eadz:
Just an idea for another workaround - would using a precompiled version of react-redux would work? I see there is a precompiled version on the CDN, but this is not for the native version.

NullDivision:
It's always a posibility but then what's the use of NPM and versioning? It also won't allow you to import it using ES6 unless you download the js file.

martincik:
[...]

PR #212, opened 2015-12-08
Running react-redux with Babel 6

martincik:
Refs #206


.babelrc
```diff
@@ -1,4 +1,21 @@
 {
-  "stage": 0,
-  "loose": "all"
+  "presets": ["es2015", "stage-0", "react"],
+  "plugins": [
```
chentsulin: Why we need to add those plugins? They should already included in presets.

martincik: Refactored the code. Thanks. :+1:

test/components/connect.spec.js
```diff
@@ -985,13 +985,14 @@ describe('React', () => {
 
       function imitateHotReloading(TargetClass, SourceClass) {
         // Crude imitation of hot reloading that does the job
-        Object.keys(SourceClass.prototype).filter(key =>
+        Object.getOwnPropertyNames(SourceClass.prototype).filter(key =>
           typeof SourceClass.prototype[key] === 'function'
         ).forEach(key => {
-          if (key !== 'render') {
+          if (key !== 'render' && key !== 'constructor') {
```
chentsulin: Is modifications in this file relative to babel6?

martincik: Yes.

satazor:
+1

ellbee:
@gaearon Do we want loose mode for IE8? React-Redux previously used it but this PR does not.

gaearon:
How is loose mode related to IE8? I thought it was about improving perf.

gaearon:
Either way I'd prefer adding loose mode back :-). I don't know how to do it though.

ellbee:
When not in loose mode Babel uses Object.defineProperty. You are right about the perf too though.

There is a preset for loose mode. The Babel 6 PR in redux has it.

gaearon:
Can you please fix it here?

ellbee:
Yep, give me a minute.

gaearon:
@ellbee Any other work there, or can I release 4.0.1 that doesn't have `.babelrc`?

ellbee:
@gaearon Go for it.

gaearon:
Out with 4.0.1 :+1:

PR #213, opened 2015-12-11
Use files in package.json

pietropizzi:
Copying the approach in https://github.com/rackt/redux/pull/1039. Need to get rid of the `.babelrc` file for React Native 0.16.0

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,1 +1,0 @@
-examples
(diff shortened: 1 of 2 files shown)


pietropizzi:
Just saw https://github.com/rackt/react-redux/pull/212. But I guess this still does not hurt

PR #219, opened 2015-12-15
Consistent usage of props instead of properties in docs

Strajk:
Great docs btw!

react-redux 2015-12-15 09-07-25


gaearon:
Thanks for the diff screenshot. :+1:

PR #222, opened 2015-12-18
Alter test case for dispatching in componentWillMount

jedborovik:
Test case for issue https://github.com/rackt/react-redux/issues/210.

I chose to edit the existing test rather than create a new test. The original test case in question looks like it tests dispatching in `componentWillMount`, but it doesn't actually check if the props make it into the component the first time it is rendered. So rather than keep it as it was, I changed it so it now properly tests (and fails).

### DIFF
--- a/test/components/connect.spec.js
+++ b/test/components/connect.spec.js
@@ -142,12 +142,12 @@ describe('React', () => {
         }
 
         render() {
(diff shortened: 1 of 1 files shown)


gaearon:
Closing as wontfix because https://github.com/rackt/react-redux/issues/210#issuecomment-166055644.

PR #225, opened 2015-12-22
Move the side effects out of shouldComponentUpdate()

### LINKED ISSUE #224: Hot reloading via react-transform/babel6 is not working when rendering a connected component from a non-connected component
I think this is an issue with how react-redux handles the hot reloading of connected components. If you have a non-connected (dumb) component render a connected component as a wrapper component, any content within that wrapper will no longer be re-rendered on change.

***Connected component:***
````
class Form extends Component {
  constructor(props) {
    super(props);
  }

  render() {
    return (
      <form>
        {this.props.children}
      </form>
    );
  }
}

export default connect(...)(Form);
````

***Dumb component rendering connected component:***
````
class LoginForm extends Component {
  constructor(props, context) {
    super(props, context);
  }

  render() {
    return (
      <Form id="loginForm">
        this is a test
      </Form>
    );
  }
}
````

Changing the text `this is a test` will not cause the component to re-render.  Replace `<Form>` with `<div>` and everything works as expected.  Replace `connect()(Form)` with just `Form` and everything works as expected.

Repository demonstrating issue is available at https://github.com/bunkat/counter.  It is the react-redux counter example with babel6 hot loading and a new connected wrapper component. 

To reproduce the issue:

1. git clone https://github.com/bunkat/counter.git
2. cd counter
3.  npm install
4.  npm start
5. Open browser to http://localhost:3000
6. Edit components/Counter.js
7. Modify the render method, change Clicked to Not Clicked

The module will be updated via HMR, but the Counter component will not be rendered with the new text.

gaearon:
This attempts to completely remove any internal side effects from `shouldComponentUpdate()` because is not in line with how React tells us to treat it and [may not even be called in certain cases, e.g. during hot reloading](https://github.com/rackt/react-redux/issues/224).

I didn't touch any tests, and that they pass is a good sign because it means we didn't regress over #99 which is where those side effects were moved to `shouldComponentUpdate()`. Here, instead, I moved them to `render()`, and applied a different optimization (namely, returning a constant element) to achieve the same effect as `shouldComponentUpdate()` did while always calling `mapStateToProps` and `mapDispatchToProps` with up-to-date props.

I confirm this PR fixes #224. It should not regress on performance.

@epeli Would you like to review?

### DIFF
--- a/src/components/connect.js
+++ b/src/components/connect.js
@@ -28,8 +28,8 @@ export default function connect(mapStateToProps, mapDispatchToProps, mergeProps,
     wrapActionCreators(mapDispatchToProps) :
     mapDispatchToProps || defaultMapDispatchToProps
   const finalMergeProps = mergeProps || defaultMergeProps
-  const shouldUpdateStateProps = finalMapStateToProps.length !== 1
-  const shouldUpdateDispatchProps = finalMapDispatchToProps.length !== 1
+  const doStatePropsDependOnOwnProps = finalMapStateToProps.length !== 1
+  const doDispatchPropsDependOnOwnProps = finalMapDispatchToProps.length !== 1
   const { pure = true, withRef = false } = options
 
   // Helps track hot reloading.
@@ -37,7 +37,7 @@ export default function connect(mapStateToProps, mapDispatchToProps, mergeProps,
 
   function computeStateProps(store, props) {
     const state = store.getState()
-    const stateProps = shouldUpdateStateProps ?
+    const stateProps = doStatePropsDependOnOwnProps ?
       finalMapStateToProps(state, props) :
       finalMapStateToProps(state)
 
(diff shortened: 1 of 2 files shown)


gaearon:
Actually I'm much happier with this code now :-). I feel like we're not working around but with React now.

epeli:
Yeah, calling map state functions in `shouldComponentUpdate` felt bit like a hack to me too but I just wanted to make a smallest possible change in order to fix the issue at hand and didn't think it would be an issue because the map state functions should be pure too.

But I recall that in somewhere in React docs they advise against keeping references to created React elements so I wonder whether it would be better to just cache the data?

gaearon:
No, caching React element is an advanced perf pattern. (People usually do this for wrong reasons so it's not recommended.) The thing is, it works like `shouldComponentUpdate()`. React will bail out of reconciliation if we return the same element. That's why my change is not a perf regression.

epeli:
Nice. This should make perf even a bit better.

PR #233, opened 2015-12-28
Use CommonJS modules until ES3 transforms are fixed

gaearon:
Our current presets didn't include ES3 transforms by default, so IE8 support is broken. Normally we would have fixed it by adding a few plugins:

* babel-plugin-transform-es3-member-expression-literals
* babel-plugin-transform-es3-property-literals

Unfortunately there is a [bug in Babel](https://phabricator.babeljs.io/T2817) that prevents them from working correctly in certain cases with `export default` declarations. Until this is resolved, we will go back to using CommonJS modules internally.

This fixes https://github.com/rackt/react-redux/issues/232.


gaearon:
Fixed via https://github.com/rackt/react-redux/commit/65a80f89dbad5e1d3d4ff8ec11472ce0c966aad5.

PR #238, opened 2016-01-04
Update troubleshooting.md

sausagejohnson:
Module import tip that could be left over before adding react-redux. Easy to overlook.


gaearon:
Can you reword this in a way that doesn't imply a specific convention? You're using default exports for connected components and named exports for the wrapped components, but not everybody does it this way. It's better to rephrase to explain that you need to make sure that whatever you import (no matter how you do it) should matter whatever way you export (no matter which exactly) your connected component.

gaearon:
Closing for inactivity but please feel free to reword as I suggested above.

PR #245, opened 2016-01-09
Clearified the "Don’t do this"

schwolf:
I am a Redux beginner and I stumbled over the "Don't do this!" because I thought that injecting dispatch + every field in the global state should generally not be done - never.
After doing some research my understanding is that this is perfectly ok for applications with only one container component connected to Redux.
So my addition should clearify this.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -104,7 +104,7 @@ export default connect()(TodoApp)
 
 ##### Inject `dispatch` and every field in the global state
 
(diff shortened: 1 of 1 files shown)


gaearon:
Well, technically having just one component in a big app is still going to be a performance problem so we don't recommend that.

gaearon:
As said above we no longer recommend having just one connected component so I believe the warning still holds.

PR #256, opened 2016-01-17
Create a redistributable bundle

### LINKED ISSUE #243: jsmain:next should point to a “redistributable ES2015 bundle”
what that means is “syntax the platform (browser/node) can read except for the ES2015 import/export syntax”

funnily my test project doesn’t work with specifically react-redux:

* <code>rollup([babel](https://github.com/rollup/rollup-plugin-babel)(), [commonjs](https://github.com/rollup/rollup-plugin-commonjs)(), [npm](https://github.com/rollup/rollup-plugin-npm)({'jsnext:main': true}))</code> → rollup’s acon can’t parse the object spread syntax [here](https://github.com/rackt/react-redux/blob/master/src/components/connect.js#L12)
* `'jsnext:main': false` → https://github.com/rollup/rollup-plugin-commonjs/issues/29

cc @Rich-Harris @TrySound @Victorystick

flying-sheep:
and point to it from jsnext:main.

relatively inelegant build script as i have to basically replicate the babel-cli/dir script, but i can’t see a better solution.

fixes #243

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,6 +1,12 @@
 {
-  "presets": ["es2015-loose", "stage-0", "react"],
-  "plugins": [
-    "transform-decorators-legacy"
-  ]
+  "env": {
+    "development": {
+      "presets": ["es2015-loose", "stage-0", "react"],
+      "plugins": ["transform-decorators-legacy"]
+    },
+    "buildmodule": {
+      "presets": ["es2015-loose-rollup", "stage-0", "react"],
+      "plugins": ["transform-decorators-legacy"]
+    }
+  }
 }
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,5 @@ npm-debug.log
 .DS_Store
 dist
 lib
+mod
 coverage
--- a/package.json
+++ b/package.json
@@ -3,13 +3,14 @@
   "version": "4.0.6",
   "description": "Official React bindings for Redux",
   "main": "./lib/index.js",
-  "jsnext:main": "./src/index.js",
+  "jsnext:main": "./mod/index.js",
   "scripts": {
     "build:lib": "babel src --out-dir lib",
+    "build:mod": "BABEL_ENV=buildmodule babel src --out-dir mod",
     "build:umd": "webpack src/index.js dist/react-redux.js --config webpack.config.development.js",
     "build:umd:min": "webpack src/index.js dist/react-redux.min.js --config webpack.config.production.js",
-    "build": "npm run build:lib && npm run build:umd && npm run build:umd:min",
-    "clean": "rimraf lib dist coverage",
+    "build": "npm run build:lib && npm run build:mod && npm run build:umd && npm run build:umd:min",
+    "clean": "rimraf lib mod dist coverage",
     "lint": "eslint src test",
     "prepublish": "npm run clean && npm run build",
     "test": "mocha --compilers js:babel-core/register --recursive --require ./test/setup.js",
@@ -23,7 +24,7 @@
   "files": [
     "dist",
     "lib",
-    "src"
+    "mod"
   ],
   "keywords": [
     "react",
@@ -49,6 +50,7 @@
     "babel-loader": "^6.2.0",
     "babel-plugin-transform-decorators-legacy": "^1.2.0",
     "babel-preset-es2015-loose": "^6.1.4",
+    "babel-preset-es2015-loose-rollup": "^7.0.0",
     "babel-preset-react": "^6.3.13",
     "babel-preset-stage-0": "^6.3.13",
     "eslint": "^1.7.1",
--- a/src/components/Provider.js
+++ b/src/components/Provider.js
@@ -1,5 +1,5 @@
-const { Component, PropTypes, Children } = require('react')
-const storeShape = require('../utils/storeShape')
+import { Component, PropTypes, Children } from 'react'
+import storeShape from '../utils/storeShape'
 
 let didWarnAboutReceivingStore = false
 function warnAboutReceivingStore() {
@@ -17,7 +17,7 @@ function warnAboutReceivingStore() {
   )
 }
 
-class Provider extends Component {
+export default class Provider extends Component {
   getChildContext() {
     return { store: this.store }
   }
@@ -49,5 +49,3 @@ Provider.propTypes = {
 Provider.childContextTypes = {
   store: storeShape.isRequired
 }
-
-module.exports = Provider
--- a/src/components/connect.js
+++ b/src/components/connect.js
@@ -1,10 +1,10 @@
-const { Component, createElement } = require('react')
-const storeShape = require('../utils/storeShape')
-const shallowEqual = require('../utils/shallowEqual')
-const isPlainObject = require('../utils/isPlainObject')
-const wrapActionCreators = require('../utils/wrapActionCreators')
-const hoistStatics = require('hoist-non-react-statics')
-const invariant = require('invariant')
+import { Component, createElement } from 'react'
+import storeShape from '../utils/storeShape'
+import shallowEqual from '../utils/shallowEqual'
+import isPlainObject from '../utils/isPlainObject'
+import wrapActionCreators from '../utils/wrapActionCreators'
+import hoistStatics from 'hoist-non-react-statics'
+import invariant from 'invariant'
 
 const defaultMapStateToProps = state => ({}) // eslint-disable-line no-unused-vars
 const defaultMapDispatchToProps = dispatch => ({ dispatch })
@@ -21,7 +21,7 @@ function getDisplayName(WrappedComponent) {
 // Helps track hot reloading.
 let nextVersion = 0
 
-function connect(mapStateToProps, mapDispatchToProps, mergeProps, options = {}) {
+export default function connect(mapStateToProps, mapDispatchToProps, mergeProps, options = {}) {
   const shouldSubscribe = Boolean(mapStateToProps)
   const finalMapStateToProps = mapStateToProps || defaultMapStateToProps
   const finalMapDispatchToProps = isPlainObject(mapDispatchToProps) ?
@@ -273,5 +273,3 @@ function connect(mapStateToProps, mapDispatchToProps, mergeProps, options = {}) 
     return hoistStatics(Connect, WrappedComponent)
   }
 }
-
-module.exports = connect
--- a/src/index.js
+++ b/src/index.js
@@ -1,4 +1,4 @@
-const Provider = require('./components/Provider')
-const connect = require('./components/connect')
+import Provider from './components/Provider'
+import connect from './components/connect'
 
-module.exports = { Provider, connect }
+export { Provider, connect }
--- a/src/utils/isPlainObject.js
+++ b/src/utils/isPlainObject.js
@@ -4,7 +4,7 @@ const fnToString = (fn) => Function.prototype.toString.call(fn)
  * @param {any} obj The object to inspect.
  * @returns {boolean} True if the argument appears to be a plain object.
  */
-function isPlainObject(obj) {
+export default function isPlainObject(obj) {
   if (!obj || typeof obj !== 'object') {
     return false
   }
@@ -23,5 +23,3 @@ function isPlainObject(obj) {
     && constructor instanceof constructor
     && fnToString(constructor) === fnToString(Object)
 }
-
-module.exports = isPlainObject
--- a/src/utils/shallowEqual.js
+++ b/src/utils/shallowEqual.js
@@ -1,4 +1,4 @@
-function shallowEqual(objA, objB) {
+export default function shallowEqual(objA, objB) {
   if (objA === objB) {
     return true
   }
@@ -21,5 +21,3 @@ function shallowEqual(objA, objB) {
 
   return true
 }
-
-module.exports = shallowEqual
--- a/src/utils/storeShape.js
+++ b/src/utils/storeShape.js
@@ -1,9 +1,7 @@
-const { PropTypes } = require('react')
+import { PropTypes } from 'react'
 
-const storeShape = PropTypes.shape({
+export default PropTypes.shape({
   subscribe: PropTypes.func.isRequired,
   dispatch: PropTypes.func.isRequired,
   getState: PropTypes.func.isRequired
 })
-
-module.exports = storeShape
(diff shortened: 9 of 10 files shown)


flying-sheep:
eh, why the fuck does it not work on travis?

> sh: 1: scripts/build-module.js: not found

it’s right there! pls help!

gaearon:
Any way we can avoid duplicating Babel options? I'm pretty sure we'll forget to update them at some point unless they're all in `.babelrc`.

scripts/build-module.js
```diff
@@ -0,0 +1,24 @@
...
+'use strict'
+
+const path = require('path')
+const glob = require('glob')
+const fs = require('fs')
+
+const babel = require('babel-core')
+const outputFileSync = require('output-file-sync')
+
+let options = JSON.parse(require('fs').readFileSync('.babelrc', 'utf8'))
```
flying-sheep: look here, i don’t duplicate them!

instead i read the .babelrc and replace one preset with another programmatically!

gaearon: Ah right. I was confused by the explicit mention of preset later. Can we use babelrc "env" option instead to activate it?

flying-sheep: oh, didn’t know about that. sure!

so it would be:

```json
{
  "env": {
    "development": {"presets": ["es2015-loose"]},
    "build-module": {"presets": ["es2015-loose-rollup"]}
  },
  "presets": ["stage-0", "react"],
  "plugins": ["transform-decorators-legacy"]
}
```

because "development" is the default.

only remaining question: does this work or do we need to do this?

```json
{
  "env": {
    "development": {
      "presets": ["es2015-loose", "stage-0", "react"]
    },
    "build-module": {
      "presets": ["es2015-loose-rollup", "stage-0", "react"]
    }
  },
  "plugins": ["transform-decorators-legacy"]
}
```

gaearon: I don't know whether the first option works. It probably should.

flying-sheep: done, looks good!

flying-sheep:
OK this works and is relatively elegant. it is all in .babelrc right now!

---

my only gripe: sadly either the tests fail or the module syntax gets modified in the bundle unless i duplicate everything inside of .babelrc.

```json
{
  "env": {
    "development": {
      "presets": ["es2015-loose", "stage-0", "react"],
      "plugins": ["transform-decorators-legacy"]
    },
    "buildmodule": {
      "presets": ["es2015-loose-rollup", "stage-0", "react"],
      "plugins": ["transform-decorators-legacy"]
    }
  }
}
```

⇒ works!

```json
{
  "env": {
    "development": {
      "presets": ["es2015-loose", "stage-0", "react"]
    },
    "buildmodule": {
      "presets": ["es2015-loose-rollup", "stage-0", "react"]
    }
  },
  "plugins": ["transform-decorators-legacy"]
}
```

⇒ SyntaxError: Decorators are not supported yet in 6.x pending proposal update.

```json
{
  "env": {
    "development": {
      "presets": ["es2015-loose"]
    },
    "buildmodule": {
      "presets": ["es2015-loose-rollup"]
    }
  },
  "presets": ["stage-0", "react"],
  "plugins": ["transform-decorators-legacy"]
}
```

⇒ tests fail

```json
{
  "env": {
    "buildmodule": {
      "presets": ["es2015-loose-rollup"]
    }
  },
  "presets": ["es2015-loose", "stage-0", "react"],
  "plugins": ["transform-decorators-legacy"]
}
```

⇒ module syntax gets compiled

flying-sheep:
i filed [this babel bug](https://phabricator.babeljs.io/T7005) for my troubles.

but i think this one is ripe for merging once you’re OK with dropping IE8 support. but it actually hasn’t been supported anymore for a week, so why wait :wink:

taion:
A few comments:

- `mod/` is not an ideal name; a number of packages use `modules/` for their actual source (e.g. all the @rackt/routing packages). I think `es6/` or `es2015/` would be better.
- You can't just switch to ES2015 modules with Babel 6 unless you add https://www.npmjs.com/package/babel-plugin-add-module-exports; otherwise you break CJS users.
- It's really odd to me that the ES2015 build uses the external-helpers plugin but the CJS build doesn't; IMO the ES2015 build should be identical to the CJS build, save only for the lack of module transpliation.

gaearon:
>mod/ is not an ideal name; a number of packages use modules/ for their actual source (e.g. all the @rackt/routing packages). I think es6/ or es2015/ would be better.

:+1: on `es6`

>You can't just switch to ES2015 modules with Babel 6 unless you add https://www.npmjs.com/package/babel-plugin-add-module-exports; otherwise you break CJS users.

This is not entirely accurate because we don't have a default export. The only reason we switched from ES6 modules to CommonJS a few releases ago is to work around Babel outputting `.default` rather than `['default']` which breaks in IE8. However we can fix this problem by putting https://github.com/sorrycc/es3ify-loader in our Webpack config.

>It's really odd to me that the ES2015 build uses the external-helpers plugin but the CJS build doesn't; IMO the ES2015 build should be identical to the CJS build, save only for the lack of module transpliation.

I think I agree with this, curious to hear the reasons.

taion:
Ah – I got confused; looks like there wouldn't be any point in doing e.g. `rackt-redux/lib/<whatever>` imports, given that in practice people are going to want both `Provider` and `connect` anyway.

But if that's the case, what does adding a proper ES2015 module build add? If you always want everything in the library, the tree-shaking is going to be a no-op.

I set this up for React Router and for history because in those cases tree-shaking would actually accomplish something, since often users wouldn't want every module in those packages.

gaearon:
>But if that's the case, what does adding a proper ES2015 module build add? If you always want everything in the library, the tree-shaking is going to be a no-op.

I don't know enough about Rollup to answer this. I thought it doesn't work with CommonJS builds at all, does it?

taion:
That doesn't sound right to me. React doesn't offer a ES2015 module build, does it?

gaearon:
Hm, good point. So then dropping `jsnext:main` should help here?

flying-sheep:
no, babel-created code is pretty hard to statically analyze. much `if (package.default != null) ...` and `__esmodule` stuff going on.

flying-sheep:
and i prefer `modules` because “es6” is confusing: that directory won’t contain ES6 code, just the module syntax, right?

taion:
Having `jsnext:main` pointing at untranspiled code will actively break Rollup users (and anybody else looking at that field).

I'll have to defer to @flying-sheep as to whether there are any benefits to having an ES2015 module build aside from the possibility of tree-shaking. I think tree-shaking matters a lot for allowing e.g.

```js
import { createBrowserHistory, useBasename, useQueries } from 'history';
```

over

```js
import createBrowserHistory from 'history/lib/createBrowserHistory';
import useBasename from 'history/lib/useBasename';
import useQueries from 'history/lib/useQueries';
```

for minimizing bundle size.

Separately, I'd say `mod/` might make sense in a vacuum, but unfortunately `modules/` already is used for non-transpiled source on a number of projects – `es6` at least conveys that you're using ES6 modules rather than CJS ones.

gaearon:
>no, babel-created code is pretty hard to statically analyze. 

Why would it need to be analyzed? We can't benefit from tree shaking in this project.

flying-sheep:
> unfortunately `modules/` already is used for non-transpiled source on a number of projects

ah, ok. others simply ship `src/`.

> Why would it need to be analyzed? We can't benefit from tree shaking in this project.

irrelevant. the important part is that the code babel creates is too complicated for rollup-plugin-commonjs to understand. (rollup-plugin-commonjs basically converts commonjs `require()` calls to ES2015 module syntax)

so if we want to be static-analysis-friendly, we need to ship simple commonjs (`const foo = require('foo')`) or ES2015 module syntax.

i prefer the latter, because the former leads to guesswork if `const foo = require('foo')` means `import foo from 'foo'` or `import * as foo from 'foo'`.

taion:
Sure, but there are plenty of packages that do use babel-compiled module syntax, and nevertheless they are used by rollup users.

What do you get from static analysis other than tree-shaking?

flying-sheep:
> Sure, but there are plenty of packages that do use babel-compiled module syntax, and nevertheless they are used by rollup users.

well, it works if you’re lucky. else you’ll end up with [this](https://gist.github.com/flying-sheep/94477c48c69a65d410d2) ([discussion](https://github.com/rollup/rollup-plugin-commonjs/issues/29))

---

btw i’m really discouraged right now. rollup seems really promising, but this discussion dragged on far too long already to have any hope in rollup becoming practical.

i really don’t want to have endless discussion with every package maintainer about this

taion:
Well, no, for some packages having an ES2015 module build is obviously beneficial for both Rollup and webpack 2 users. I've personally added ES2015 module builds to both React Router and history.

That said, there is some non-zero cost to the module maintainer to maintain an additional build. From my perspective, if there are meaningful benefits available through tree-shaking, as there are for my other packages like React Router, history, and React-Bootstrap, then I'm happy to add the ES2015 module build.

For something where there's really only one export, like e.g. react-router-relay, or possibly react-redux? I think there needs to be a concrete benefit to adding another build target.

If nothing else I hope this discussion will demonstrate good general guidelines for other projects in considering whether to add an ES2015 module build.

flying-sheep:
> If nothing else I hope this discussion will demonstrate good general guidelines for other projects in considering whether to add an ES2015 module build.

you’re right, that occurred to me as well. we’re treading unexplored ground right here, so it’s probably not bad to have some discussion now and a “best practices” thing in the end.

---

> I think there needs to be a concrete benefit to adding another build target.

IMHO there is. it’s the difference between a default export and grab-bag-of-stuff exports. `import * as name` vs `import name`.

rollup handles those as strict as the real deal in the future. you won’t be able to do `import React, { Component } from 'react'` if there’s not both individual and default exports. babel however allows that.

currently i have to go to `node_modules` or even github and look into the ES6 source code to see which export style is intended by the authors.

TrySound:
Commonjs requires some extra code in bundle. Even without tree shaking result will be smaller.

taion:
@flying-sheep 

I don't think that's correct. Looking at the implementation of `_interopRequireDefault`, you wouldn't be able to do e.g.

```js
import ReactRedux from 'react-redux';
```

At this point in time, I also don't think the extra bundle size purely from using CJS modules rather than ES6 ones (when tree shaking is not part of the picture) is really all that meaningful.

flying-sheep:
forget rollup. forget bundling. static analysis of dependencies is important for tooling.

`_interopRequireDefault` is a function call, not static syntax. therefore we need ES2105 module syntax in order to advance the JS ecosystem.

taion:
We do, yes, but there's no accounting for timing. Eventually Node is going to have actual support for ES2015 modules, and webpack 2 will have support for the same – at which point it'd be possible to distribute just an ES2015 module build, and not have both that and a CJS build.

flying-sheep:
don’t forget people who’ll distribute node 0.12 stuff for backwards compatibility

flying-sheep:
so back to topic: when renaming the module directory to es2015 or es6, is everyone confident that this is a good way to go?

if not, please speak up and voice your concerns, and ideally propose alternatives.

taion:
If you have a hard dependency on a specific Node version for your React code, you're doing something very odd/wrong.

I just feel like, in the absence of a compelling bundle size benefit from tree shaking, that it's not really worthwhile to maintain two transpiled builds – ideally at some point in the future it will be possible to just swap out the CJS module build for an ES2015 module build.

It's not "advancing the JS ecosystem" to just set up some awkward shims that most people can't take advantage of – Rollup is not an option in general for packaging a full-fledged application, and webpack 2 isn't ready yet.

flying-sheep:
> Rollup is not an option in general for packaging a full-fledged application

because not enough npm modules support jsnext:main, right?

taion:
No, because real applications tend to need things like CSS and other assets.

flying-sheep:
how is this relevant?

taion:
What would one use webpack, Browserify, or Rollup for, if not to generate client-side bundles?

flying-sheep:
sure, but either you use [JSS](https://github.com/jsstyles/jss) and so on or bundle assets independently

taion:
Either way, most users are currently simply not going to be able to take advantage of an ES2015 module build. Maybe when webpack 2 leaves beta, assuming it doesn't hit the same teething issues as Babel 6, and/or when Node cuts a stable release that supports this module syntax.

I'd happily encourage and even push people toward using proper ES2015 modules over the transpiled CJS stuff once it's something that has the potential for broad adoption, but we're not at that point yet, and "it feels cleaner" is not IMO enough of a benefit to make the build pipeline more complicated on all the libraries.

We should save it for the ones where tree shaking gives a concrete bundle size reduction.

taion:
I'd also prefer we use https://github.com/gajus/babel-preset-es2015-webpack (perhaps a loose version?) over the rollup preset, since the webpack one doesn't introduce the extra babel external transform, and more of our users will likely be using webpack than rollup.

timdorr:
This is recommended by @sokra too: https://gist.github.com/sokra/27b24881210b56bbaff7#babel-and-webpack

However, we aren't using presets on the project, so it's just a matter of [removing this line](https://github.com/rackt/react-redux/blob/master/.babelrc#L21).

flying-sheep:
the rollup preset’s transform externalizes the babel-helpers, that’s all

gaearon:
I don’t really understand what the deal with Babel external helpers is, so I can’t move forward here.
Please help me understand the issue.
For now, I released React Redux that removes broken `jsnext:main` because it didn’t help anybody.

flying-sheep:
> For now, I released React Redux that removes broken jsnext:main because it didn’t help anybody.

thank you!

> I don’t really understand what the deal with Babel external helpers is, so I can’t move forward here.
> 
> Please help me understand the issue.

[here](https://babeljs.algolia.com/docs/advanced/external-helpers/) and [here](https://github.com/rollup/rollup-plugin-babel/blob/master/src/index.js#L26-L28) is the explanation:

the plugin makes the individual compiled files rely on a “babelHelpers” variable. rollup-plugin-babel detects if this is the case (but also handles other ways to include babel helpers, [warning](https://github.com/rollup/rollup-plugin-babel/blob/master/src/index.js#L90) if necessary).

taion:
Can we pull in the tooling from the main Redux library, then figure out across-the-board what to do with external helpers?

I don't think it'd be good to have an ES2015 module build that only works with Rollup and not with webpack 2.

kidwm:
FYI, babel 6.6.0 has fixed the exports default problem in IE8.
https://github.com/babel/babel/blob/master/CHANGELOG.md#exportsdefault-fix

gaearon:
I’m closing as this is rather outdated. We can revisit later if there is enough desire to do so.

PR #257, opened 2016-01-18
removed redundant `package manager`

axdg:
Just reading through the readme and picked up on this. [npm](https://www.npmjs.com/) is an acronym for Node Package Manager, so the phrase `npm package manager` essentially evaluates to `node package manager package manager`.

No sure that anyone else will care, but it was bugging me. It's like when people say 'ATM machine', or 'as ASAP as possible'.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ React Redux requires **React 0.14 or later.**
 npm install --save react-redux
 ```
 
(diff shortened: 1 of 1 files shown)


gaearon:
I agree but some people don't know what npm. I'm happy to accept a rewording like "a package manager like npm" though!

Issue #236, opened 2016-01-03
Update React Native installation instructions 


Emilios1995:
Seems that React Native works on top of react already. the issue https://github.com/facebook/react-native/issues/2985 is already closed.
Can we already use React-Redux 4?

epeli:
That change is not in the released version yet. In few days I guess...

zenlambda:
The change is included in the newly released 0.18-rc (I haven't tested it with redux 4 yet), so hopefully not long to go!

alinz:
OK, I got `react-redux 4.0.6` to work with `react-native`, it will be available in [example-react-native-redux](https://github.com/alinz/example-react-native-redux).

There are couple of steps need to be done which I hide them at the moment in `package.json`'s `postscript`. Here's the details:

1. `react` module in `react-native/node_module` must be deleted.
2. `fbjs` module in `react-native/node_modules` must be deleted.
3. `fbjs` module in `react/node_modules` must be deleted.
4. `fbjs` module must be added as module to project in main `package.json`
5. `all .babelrc` files must be deleted under node_module folder

There is a discussion here https://github.com/facebook/react-native/issues/2985#issuecomment-168412030 which @spicyj talks about changing fbjs.

Also @gaearon I still need to remove all .babelrc files under node_modules before I can run the [example-react-native-redux](https://github.com/alinz/example-react-native-redux). What is your suggestion?

gaearon:
I don't have any suggestions here. Feedback from people more knowledgeable in RN is welcome. 

Perhaps @ide or @spicyj could clarify the situation.

Purii:
Before checking anything I just replaced `import { connect } from 'react-redux/native'` with `import { connect } from 'react-redux'`. Also updated the following:
```javascript
import { Provider } from 'react-redux/native';
...
<Provider store={store}>	
  {() => <App/>}
</Provider>

TO:

import { Provider } from 'react-redux';
...
<Provider store={store}>	
 <App/>
</Provider>
```

I use the most recent version of `react-redux`:
```
"react-native": "^v0.18.0-rc",
"react-redux": "^4.0.6",
"redux": "^3.0.5",
```

I didn't need to change anything in the `react-redux` package. It just works, as expected by a tool from @gaearon ;-)

alinz:
[...]

PR #258, opened 2016-01-20
Remove react native warning

zackify:
Now that 0.18 is out, this works without the old version.


gaearon:
I've heard that without npm 3 there is a risk of having duplicate React and weird bugs.
Is this the case?

gaearon:
See also #236

zackify:
I am on npm 3 so I can't confirm.

Issue #268, opened 2016-01-25
Context store name


nickdima:
I see that the store is set on on the context with the name `store` which is fine in most cases but since the context is global there are case when you need it to be more specific.
I have an existent app which I want to port to Redux incrementally but it already has something called `store` on the context.
The solutions I see here are either choose a more specific name eg. `reduxStore` or have it passed as a prop to `Provider`.
I could make a pull request if any of these two are acceptable.

danmartinez101:
Taking an example from another Rackt project: https://github.com/rackt/react-router/blob/449fefa4d55025ea459d6b9e2e3aa7169e00b0f5/examples/master-detail/app.js#L66

react-router => context.router
react-redux => context.redux

nickdima:
So, what are you suggesting?

danmartinez101:
I'm suggesting the context property for react-redux be "redux" in the same way that react-router's is "router".

nickdima:
Yeah, that's definitely the way to go. Should we CC someone that has merge privileges before making a pull request?

danmartinez101:
I'd say just go ahead and make a PR for it and reference this issue in the descriptions. When they have time, they will review and comment / decide; no need to cc imo.

gnoff:
> I'm suggesting the context property for react-redux be "redux" in the same way that react-router's is "router".

I thought the same thing at first but react-router is still claiming a implementation non-specific namespace on context. they technically have the same problem we have with `store`. i've commented on the PR regarding supporting as `as` prop which optionally places the store in another namespace on context. I don't think this change is worth breaking everyone who has depended on the store outside of Provider and connect personally

gaearon:
Please see my response in https://github.com/rackt/react-redux/pull/272#issuecomment-175215719.

PR #272, opened 2016-01-26
Rename 'store' context to 'redux' in order to support #268

nickdima:
The store is set on on the context with the name `store` which is fine in most cases but since the context is global there are cases when you need it to be more specific.
I have an existent app which I want to port to Redux incrementally but it already has something called store on the context. 
The proposed solution here is to rename `store` to `redux` as per #268

### DIFF
--- a/src/components/Provider.js
+++ b/src/components/Provider.js
@@ -19,7 +19,7 @@ function warnAboutReceivingStore() {
 
 class Provider extends Component {
   getChildContext() {
-    return { store: this.store }
+    return { redux: this.store }
   }
 
   constructor(props, context) {
@@ -47,7 +47,7 @@ Provider.propTypes = {
   children: PropTypes.element.isRequired
 }
 Provider.childContextTypes = {
-  store: storeShape.isRequired
+  redux: storeShape.isRequired
 }
 
 module.exports = Provider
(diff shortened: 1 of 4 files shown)


gnoff:
this is technically an internal api change only but I believe lots of people have depended on the store name of the context property so this would be a breaking change in practice for quite a few people.

A backwards compatible way would be to add an optional prop that describes where the store lives on context.

```
//default case
<Provider store={theStore} />

//context prop override, would put store on context as 'redux'
<Provider store={theStore} as="redux" />
```

This is opt in and backwards compat

gaearon:
This would a breaking change so we're not going to take this.
Context API will eventually change to avoid string keys so it will be resolved naturally at some point.

In the meantime you can either pass `store` manually as a prop to all connected components or create a custom `<Provider>` that puts the store into `redux` context field. Then you can write a custom `connect()` that uses `connect()` from React Redux but also provides a store as a prop to the wrapped component from `redux` context field.

nickdima:
OK, thanks.

gaearon:
Yeah, sorry about that. It’s just not worth breaking everyone because React will fix it eventually. And people might already be using `redux` context name in their projects (e.g. for bound action creators or utilities).

nickdima:
Yeah, totally agree. I had doubts about it myself also :)

PR #279, opened 2016-02-03
Overloading connect with factory

tgriesser:
I had actually considered this before doing the `connectFactory` approach in #185. The issue was that it felt wrong to overload the arguments and require first calling the function to determine its return type.

That said, this is the cleanest implementation I could come up with. It checks if the argument in question has `fn.length` of 0 and also returns a function.

This breaks a few tests, but nothing critical - only on some spies or guarantees an arity-0 function is called a certain amount of times. I can clean those up if this looks alright otherwise.

### DIFF
--- a/src/components/connect.js
+++ b/src/components/connect.js
@@ -18,50 +18,32 @@ function getDisplayName(WrappedComponent) {
   return WrappedComponent.displayName || WrappedComponent.name || 'Component'
 }
 
+function checkStateShape(stateProps, dispatch) {
+  invariant(
+    isPlainObject(stateProps),
+    '`%sToProps` must return an object. Instead received %s.',
+    dispatch ? 'mapDispatch' : 'mapState',
+    stateProps
+  )
+  return stateProps
+}
+
 // Helps track hot reloading.
 let nextVersion = 0
 
 export default function connect(mapStateToProps, mapDispatchToProps, mergeProps, options = {}) {
   const shouldSubscribe = Boolean(mapStateToProps)
-  const finalMapStateToProps = mapStateToProps || defaultMapStateToProps
-  const finalMapDispatchToProps = isPlainObject(mapDispatchToProps) ?
+  const mapState = mapStateToProps || defaultMapStateToProps
+  const mapDispatch = isPlainObject(mapDispatchToProps) ?
     wrapActionCreators(mapDispatchToProps) :
     mapDispatchToProps || defaultMapDispatchToProps
+
   const finalMergeProps = mergeProps || defaultMergeProps
-  const doStatePropsDependOnOwnProps = finalMapStateToProps.length !== 1
-  const doDispatchPropsDependOnOwnProps = finalMapDispatchToProps.length !== 1
+  const checkMergedEquals = finalMergeProps !== defaultMergeProps
   const { pure = true, withRef = false } = options
 
   // Helps track hot reloading.
   const version = nextVersion++
-
-  function computeStateProps(store, props) {
-    const state = store.getState()
-    const stateProps = doStatePropsDependOnOwnProps ?
-      finalMapStateToProps(state, props) :
-      finalMapStateToProps(state)
-
-    invariant(
-      isPlainObject(stateProps),
-      '`mapStateToProps` must return an object. Instead received %s.',
-      stateProps
-    )
-    return stateProps
-  }
-
-  function computeDispatchProps(store, props) {
-    const { dispatch } = store
-    const dispatchProps = doDispatchPropsDependOnOwnProps ?
-      finalMapDispatchToProps(dispatch, props) :
-      finalMapDispatchToProps(dispatch)
-
-    invariant(
-      isPlainObject(dispatchProps),
-      '`mapDispatchToProps` must return an object. Instead received %s.',
-      dispatchProps
-    )
-    return dispatchProps
-  }
 
   function computeMergedProps(stateProps, dispatchProps, parentProps) {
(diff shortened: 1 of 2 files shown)


gaearon:
Rather than look at arg count, can we determine whether they returned functions instead of objects lazily first time we call them from the constructor?

tgriesser:
Took another pass at it, it's a little more verbose / complicated, but does things lazily as you mentioned, all tests are passing and it avoids calling the mapping functions any more times than necessary. 

If this looks alright I can add some tests.

src/components/connect.js
```diff
@@ -14,65 +14,40 @@ const defaultMergeProps = (stateProps, dispatchProps, parentProps) => ({
   ...dispatchProps
 })
 
+function isFunction(fn) {
```
gaearon: I would leave those checks inline.

gaearon:
Yes, this looks good. Please do add tests.

tgriesser:
Updated, rebased, tests added. Also noticed when writing the tests that a final shallow equality check on the final merged props would save a render if there is a custom `mergeProps` supplied and nothing has changed, so I added that as well.

test/components/connect.spec.js
```diff
@@ -1509,5 +1509,108 @@ describe('React', () => {
...
+        <ProviderMock store={store}>
+          <Container />
+        </ProviderMock>
+      )
+
+      store.dispatch({ type: 'test' })
+      expect(updatedCount).toBe(0)
+      expect(memoizedReturnCount).toBe(2)
+    })
+
```
gaearon: Nit: mind killing the newline here?

src/components/connect.js
```diff
@@ -18,51 +18,33 @@ function getDisplayName(WrappedComponent) {
...
-  const finalMapDispatchToProps = isPlainObject(mapDispatchToProps) ?
+  const mapState = mapStateToProps || defaultMapStateToProps
+  const mapDispatch = isPlainObject(mapDispatchToProps) ?
     wrapActionCreators(mapDispatchToProps) :
     mapDispatchToProps || defaultMapDispatchToProps
+
   const finalMergeProps = mergeProps || defaultMergeProps
-  const doStatePropsDependOnOwnProps = finalMapStateToProps.length !== 1
-  const doDispatchPropsDependOnOwnProps = finalMapDispatchToProps.length !== 1
+  const checkMergedEquals = finalMergeProps !== defaultMergeProps
```
gaearon: Can we add a new test for this optimization so we don't regress?

src/components/connect.js
```diff
@@ -96,8 +78,47 @@ export default function connect(mapStateToProps, mapDispatchToProps, mergeProps,
...
+        const isFactory = typeof mappedState === 'function'
+        this.finalMapStateToProps = isFactory ? mappedState : mapState
+        this.doStatePropsDependOnOwnProps = this.finalMapStateToProps.length !== 1
+        return isFactory ? this.computeStateProps(store, props) : checkStateShape(mappedState)
+      }
+
+      computeDispatchProps(store, props) {
+        if (!this.finalMapDispatchToProps) {
+          return this.configureFinalMapDispatch(store, props)
+        }
```
gaearon: Nit: please add a newline here

src/components/connect.js
```diff
@@ -96,8 +78,47 @@ export default function connect(mapStateToProps, mapDispatchToProps, mergeProps,
...
+        const { dispatch } = store
+        const dispatchProps = this.doDispatchPropsDependOnOwnProps ?
+          this.finalMapDispatchToProps(dispatch, props) :
+          this.finalMapDispatchToProps(dispatch)
+        return checkStateShape(dispatchProps, true)
```
gaearon: Nit: please, in general, format more sparsely.

```js
if (!this.finalMapDispatchToProps) {
  return this.configureFinalMapDispatch(store, props)
}

const { dispatch } = store
const dispatchProps = this.doDispatchPropsDependOnOwnProps ?
  this.finalMapDispatchToProps(dispatch, props) :
  this.finalMapDispatchToProps(dispatch)

return checkStateShape(dispatchProps, true)
```

src/components/connect.js
```diff
@@ -96,8 +78,47 @@ export default function connect(mapStateToProps, mapDispatchToProps, mergeProps,
...
+      }
+
+      configureFinalMapDispatch(store, props) {
+        const mappedDispatch = mapDispatch(store.dispatch, props)
+        const isFactory = typeof mappedDispatch === 'function'
+        this.finalMapDispatchToProps = isFactory ? mappedDispatch : mapDispatch
+        this.doDispatchPropsDependOnOwnProps = this.finalMapDispatchToProps.length !== 1
+        return isFactory ? this.computeDispatchProps(store, props) : checkStateShape(mappedDispatch, true)
```
gaearon: Nit: Would be great to break this ternary into three lines like the rest of them

gaearon:
That’s amazing, thank you so much for your work on this. A few more nits and it’s good to go.

tgriesser:
All set with a test and a bit of formatting polish

gaearon:
Would you be up to contributing a section to [Computing Derived Data](https://github.com/rackt/redux/blob/master/docs/recipes/ComputingDerivedData.md) recipe documenting that? I think a lot of people won’t learn about this useful feature otherwise.

tgriesser:
Sure thing, I'll take a look at doing this a little later.

ellbee:
@tgriesser I have started to update the Reselect docs to reflect this change. If you haven't started on this already, would you like me to carry on and document it?

PR #283, opened 2016-02-10
Add jsnext:main property to the package.json

salztorte:
This help some bundler

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "version": "4.4.0",
   "description": "Official React bindings for Redux",
   "main": "./lib/index.js",
+  "jsnext:main" : "src/index.js",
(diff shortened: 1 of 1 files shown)


TrySound:
@salztorte This will not help. In this case you should process with babel this code. It's not fast.

TrySound:
@salztorte See here
https://github.com/rackt/redux/pull/1369

gaearon:
We had it before, but this is not what `jsnext:main` should point to.

See discussions in https://github.com/rackt/redux/pull/1369, https://github.com/rackt/redux/issues/1042, https://github.com/rackt/redux/pull/1327, https://github.com/rackt/react-redux/pull/256.

Issue #269, opened 2016-01-25
connect could be used with a custom store.subscribe method provided as option


slorber:
# Context

Today there was some discussions about the new redux tree view example and this SO question:
http://stackoverflow.com/questions/34981924/performance-issues-with-a-tree-structure-and-shouldcomponentupdate-in-react-re

As I pointed out the solution of Redux examples is fine enough for most usecases but does not scale it the number of nodes is very big. We agree that it's not a good idea in the first place to render a huge list or a huge tree, but it seems the author of the question is still interested in a solution. I don't know if the usecase is valid or an antipattern but I guess it's worth giving the opportunity to render a huge number of items in an efficient way

# Solution

As described in my answer if the number of connected items grows too big, then every state change triggers all connected HOC's subscriptions so it does not scale well.

If we want to make it scale better, the HOC subscriptions should only be called when necessary.

If the state slice of node with id=1 is updated, it produces overhead to actually trigger the subscription of the HOC that connects data to node with id=2, because obviously the connected component is not interested at all in this change.

Redux store has a `store.subscribe(listener)` method, and it is called inside `connect`

One can easily create a store enhancer that exposes a method like `store.subscribeNode(nodeId,listener)`, which is only triggered when that node, with the provided, id is called. There are multiple ways of triggering this listener efficiently.

The problem is that once we have built this custom optimized subscription system, there's no way to reuse `connect` to listen to it because the call to `store.subscribe` is hardcoded here:

```
    trySubscribe() {
        if (shouldSubscribe && !this.unsubscribe) {
          this.unsubscribe = this.store.subscribe(::this.handleChange)
          this.handleChange()
        }
      }
````

A flexible solution would be to let the user control hw to subscribe to the store himself, by using connect options for example:

```javascript
const mapStateToProps = (state,props) => {node: selectNodeById(state,props.nodeId)}

const connectOptions = {
   doSubscribe: (store,props) => store.subscribeNode(props.nodeId)
}

connect(mapStateToProps, undefined,connectOptions)(ComponentToConnect)
```

The default `doSubscribe` implementation would be: `doSubscribe: (store,props) => store.subscribe()` which keeps the current behavior unchanged.

There's one corner case I don't know yet how to solve is when the nodeId props changes over time. It would mean the HOC should then unsubscribe and then resubscribe for the new nodeId. In practice this seems farfetched to do so and I'm not sure it's really worth trying to solve this but if anyone has an idea...

[...]

PR #285, opened 2016-02-11
Update connect to make it more extensible

mhodgson:
I've attempted to refactor `connect` to provide additional (optional) exports that would allow the user to easily create a custom `Connect` class. This allows the user to override specific parts of `Connect` while maintaining all the existing caching and update logic. Here is an example of usage:

```javascript
import { Connect, createConnect } from 'react-redux/components/connect'

class MyConnect extends Connect {
  handleChange() {
    if (!this.unsubscribe) {
      return
    }

    const prevStoreState = this.state.storeState
    const storeState = this.store.getState()

    if (!this.pure || prevStoreState !== storeState) {
      this.hasStoreStateChanged = true
      return new Promise((resolve) => { this.setState({ storeState }, resolve) })
    }
  }
}

export default createConnect(MyConnect)
```

In this example I've overridden the `handleChange` method to return a Promise that resolves when the changes have been flushed to the DOM.

You then just use this custom `connect` method to decorate your components instead of the default one from react-redux.

All but one of the tests are passing, but I'm interested in getting feedback on this before working on it further.

### DIFF
--- a/src/components/connect.js
+++ b/src/components/connect.js
@@ -6,6 +6,8 @@ import isPlainObject from 'lodash/isPlainObject'
 import hoistStatics from 'hoist-non-react-statics'
 import invariant from 'invariant'
 
+
+
 const defaultMapStateToProps = state => ({}) // eslint-disable-line no-unused-vars
 const defaultMapDispatchToProps = dispatch => ({ dispatch })
 const defaultMergeProps = (stateProps, dispatchProps, parentProps) => ({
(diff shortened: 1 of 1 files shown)


gaearon:
Inheritance is generally problematic, and especially so when mixed with a functional paradigm. React components should not create class hierarchies, and even for this use case we don’t want to do that.

To solve #269, I would much prefer to *extract* the core logic from `connect()` so it can become a simple wrapper that is easy to write on your own with custom logic.

mhodgson:
@gaearon thanks for the quick feedback. I'm really interested in getting something that works for this, but it seems you have pretty specific ideas about how it should work. Would you mind giving a little bit more guidance on how you'd like to see this structured? Happy to put the work in to get to something useful. I'm struggling to see how to get a clean API here with out inheritance when so many lifecycle methods are involved. We could double wrap the component but that seems even worse to me.

gaearon:
Something like `subscribeComponentToStateChanges(store, callback, mapStateToProps, mapDispatchToProps, mergeProps, options)`?

slorber:
@mhodgson yes i'd also prefer to avoid inheritence but rather use composition / strategy pattern, where the default strategy remains compatible with current behavior.

PR #293, opened 2016-02-18
[WIP] Proof of concept of batching updates

gaearon:
This is a proof of concept of batching updates in `<Provider>`.
I’m not sure we want to get this in, but it can serve as a starting point.

Solves https://github.com/reactjs/react-redux/issues/292.

Remaining issues:

* Should this be opt-in?
* We need to verify this works with RN correctly.
* Currently it forces you to use `react-dom` if you don’t use `react-native` which is a bummer for other renderers.

### DIFF
--- a/src/components/Provider.js
+++ b/src/components/Provider.js
@@ -1,5 +1,6 @@
 import { Component, PropTypes, Children } from 'react'
 import storeShape from '../utils/storeShape'
+import batchedUpdates from '../utils/batchedUpdates'
 
 let didWarnAboutReceivingStore = false
 function warnAboutReceivingStore() {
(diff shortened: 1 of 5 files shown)


epeli:
> We need to verify this works with RN correctly.

I can verify that. I ported [the test to for the real React Native environment](https://gist.github.com/epeli/4ab332f7570c2e39f483) it passes without manual `unstable_batchedUpdates` when using this branch.


> Currently it forces you to use react-dom if you don’t use react-native which is a bummer for other renderers.

I see this as a blocker. Doing it as an opt-in would solve it but what would it look like?

Maybe?

```js
<Provider store={store} updateBatcher={ReactDOM.unstable_batchedUpdates}>
  <MyApp />
<Provier>
```

timdorr:
Superseded by #416

PR #297, opened 2016-02-20
Bind `handleChange` in constructor

prashaantt:
Did you not mean to do this here?


gaearon:
In general this code executes once so there is no big difference.
However we want to avoid extra `bind()` calls on the server where `handleChange()` is not used.
So I’d rather leave it as is.

PR #317, opened 2016-03-08
Added support to react@15.0.0-rc.1

chentsulin:


istarkov:
Isn't the same must be done for `devDependencies` except `||`

jslauthor:
I would love to see this happen immediately. :smile:

gaearon:
For the future reference, you can use `npm@3.x` in which peer dependency mismatch is a warning rather than a hard error.

>Isn't the same must be done for devDependencies except ||

When 15.0 comes out, we’ll upgrade them. For now let’s run tests on the stable version.

gaearon:
Out in 4.4.1, thanks!

Issue #292, opened 2016-02-17
Can we avoid inconsistencies on non-batched dispatches?


gaearon:
Prompted by https://github.com/reactjs/redux/issues/1415, please read the discussion there. We currently subscribe in `componentDidMount` but it runs from children first. This has a potential of introducing inconsistencies when a child receives some update state earlier than its parent that passes a state-dependent prop to it.

Would subscribing the parents first fix the inconsistencies? Can we do that somehow (e.g. by passing subscribers up via context)?

Alternatively, can/should we wrap `dispatch` into `ReactDOM.unstable_batchedUpdates()` by default like Relay does?

cc @epeli @chandlerprall @tappleby @acdlite

gaearon:
>Alternatively, can/should we wrap dispatch into ReactDOM.unstable_batchedUpdates() by default like Relay does?

Obvs this wouldn’t work for React Native. :grin:

epeli:
If I understand correctly https://github.com/reactjs/redux/issues/1415 is basically dup of #99?

> Alternatively, can/should we wrap dispatch into ReactDOM.unstable_batchedUpdates() by default like Relay does?

I think so. I've been using `redux-batched-updates` in all of my projects and there have been zero issues. More often my apps tend to break if I'm not using it... It will also boost performance in some cases.

But the biggest argument for enabling it by default in my option is that it makes `dispatch()` consistent. Currently by default `dispatch()` can surprise you because it works differently depending on when you happen to call it. Ex. `onClick` vs. `setTimeout`.

Is there actually any downsides if it's made default?

> Obvs this wouldn’t work for React Native. 

Any ideas why it is in ReactDOM and not in React itself?

I haven't confirmed that the issue is actually present in React Native but I'm now working on a React Native app so I can easily test if it's not confirmed yet?

epeli:
> Is there actually any downsides if it's made default?

Hah. I just realized I might have answered it myself:

> Any ideas why it is in ReactDOM and not in React itself?

Redux nor react-redux  does not depend on react-dom and for a good reason because it's not wanted in React Native or any other non-dom React projects...

Any other reasons?

epeli:
> Would subscribing the parents first fix the inconsistencies? Can we do that somehow (e.g. by passing subscribers up via context)?

This is an interesting idea!

[...]

PR #320, opened 2016-03-10
Use componentWillMount instead of componentDidMount for subscriptions

jdeal:
Even if this PR doesn't get accepted, I'm hoping to find out the reasoning behind the current implementation and find out what patterns to use to avoid the problem I found. The problem being...

Because component store subscriptions are created in `componentDidMount` (which fires child before parent), you can get into situations like this:

1. Parents make a decision about what components to render, based on store state.
2. Children get current store state which is fine.
3. Store state is changed, requiring a new component tree.
4. Children get new store state before the parents, which is bad, because the new state may not make any sense for the new components.
5. Parents get new store state and switch out components. But it's too late, because children have already received bad state.

Here's a simple example showing the problem:

https://jsfiddle.net/justindeal/myqyxpcw/6/

There's a value in the store 'a' or 'b' which is used to switch components `A` and `B`. I have some side effect rendering to show what values are actually rendered by `A` and `B`. You can see that `A` incorrectly gets a 'b' value on the first toggle.

By switching to `componentWillMount`, subscriptions are always in parent-child order, so there's never any inconsistent state.

If you want to consider a simple real-world case, think about the filter string from the canonical todo example. Now consider that you can switch between todos and recipes. The filter for todos may make no sense for filtering recipes. So sending the global filter for todos to the child recipe components makes no sense and can cause errors to be thrown if unexpected props/values show up.

This change had one side effect for existing tests: because we haven't yet rendered when subscribing, the impure case avoids a duplicate initial call to its state mapping function. As far as I can tell, this has no effect on what's actually rendered though.

I also saw no negative outcome when I made the change for my app. (Just a fix to the surprising behavior I saw.) But I'm curious to get some feedback.

Thanks!

### DIFF
--- a/src/components/connect.js
+++ b/src/components/connect.js
@@ -176,7 +176,7 @@ export default function connect(mapStateToProps, mapDispatchToProps, mergeProps,
         }
       }
 
-      componentDidMount() {
+      componentWillMount() {
         this.trySubscribe()
       }
 
--- a/test/components/connect.spec.js
+++ b/test/components/connect.spec.js
@@ -1398,8 +1398,8 @@ describe('React', () => {
       const target = TestUtils.findRenderedComponentWithType(tree, Passthrough)
       const wrapper = TestUtils.findRenderedComponentWithType(tree, StatefulWrapper)
 
-      expect(mapStateSpy.calls.length).toBe(2)
-      expect(mapDispatchSpy.calls.length).toBe(2)
+      expect(mapStateSpy.calls.length).toBe(1)
+      expect(mapDispatchSpy.calls.length).toBe(1)
       expect(target.props.statefulValue).toEqual('foo')
 
       // Impure update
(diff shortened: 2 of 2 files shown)


gaearon:
Please see the discussion in https://github.com/reactjs/react-redux/issues/292.

We can’t use `componentWillMount` because it runs on the server, which would result in memory leaks. The solution here is to batch updates but React doesn’t yet provide a stable API to do this.

Check out #293 for a possible solution. In the meantime you can implement this in userland.

jdeal:
Ah, sorry I missed the open issue. Thanks for the heads up.

FWIW, I don't think userland really works in the case of trying to build a redux-compatible lib that provides state for multiple components. The lib would have to caveat its usage. (Of course, with this problem, usage has to be caveated anyway.)

I'll keep an eye on #292 and #293.

gaearon:
Such library could wrap its dispatches into `ReactDOM.unstable_batchedUpdates()`, could it not?

jdeal:
I guess my concern was:

The library would have to do _all_ dispatching. It can't export a raw action creator and let the consumer dispatch that. At least not without the consumer knowing that it also has to wrap those dispatches with `unstable_batchedUpdates`. Essentially, an implementation detail (about inconsistent state updating) is leaking out to the consumer.

Having said that, I think it's pretty workable for now. `unstable_batchedUpdates` seems to cure the problem, and it's pretty easy to slide in. It's more or less an aesthetically odd thing to say "caveat: use this unstable API for these edge cases."

BTW, still a big fan of Redux! Just picking this nit!

gaearon:
>It's more or less an aesthetically odd thing to say "caveat: use this unstable API for these edge cases."

Yeah but it’s the best we have until React either enables batching by default or provides a stable API for this. :-)

PR #331, opened 2016-03-30
Add example to API docs for no-subscribe actions

joshwcomeau:
In a recent project, I needed a container component that had access to a few actions, but it didn't need to subscribe to the store. The component was just a button that performed an action; it didn't care about what the state of the app was.

The docs for `connect` don't have an example for how to omit the `mapStateToProps` argument, while still providing a value for `mapDispatchToProps`.

I'm not entirely sure if this is the best way to accomplish that goal, so please do let me know if there's a better/simpler way! I tried simply passing a single `mapDispatchToProps` argument, but it was interpreted as the `mapStateToProps` param.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -106,6 +106,14 @@ Returns the wrapped component instance. Only available if you pass `{ withRef: t
 export default connect()(TodoApp)
 ```
 
+##### Inject all action creators  (`addTodo`, `completeTodo`, ...) without subscribing to the store
+
+```js
+import * as actionCreators from './actionCreators'
+
+export default connect(null, actionCreators)(TodoApp)
+```
+
 ##### Inject `dispatch` and every field in the global state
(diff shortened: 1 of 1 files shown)


gaearon:
We already have https://github.com/joshwcomeau/react-redux/blob/feature/injectActionsDocs/docs/api.md#inject-just-dispatch-and-dont-listen-to-store so maybe it’s worth putting the example next to it? Also let’s use `null` instead of `false`. Somehow that feels.. better.

joshwcomeau:
Yeah, I didn't wanna be presumptuous by putting my example so close to the top, but I agree it makes more sense there.

Since it then became the first example to use `import * as actionCreators`, I moved the parentheses explaining that bit to it.

gaearon:
Even better. Thanks!

PR #335, opened 2016-04-06
Close #309, Improve  invariant validation error message

### LINKED ISSUE #309: Improve `mapStateToProps` invariant validation error message
During refactoring of my stores I more than once hit an invariant violation like this:

```
Uncaught Invariant Violation: `mapStateToProps` must return an object. Instead received "A string".
Uncaught Invariant Violation: `mapStateToProps` must return an object. Instead received .
```

It would be nice of the name of the connected component would also be part of the error message

The latter error message above is from an "undefined" return I guess, so it should print that in the error message:
`Uncaught Invariant Violation: 'ComponentName mapStateToProps' must return an object. Instead received undefined`

chentsulin:
This should fix #309

### DIFF
--- a/src/components/connect.js
+++ b/src/components/connect.js
@@ -18,15 +18,15 @@ function getDisplayName(WrappedComponent) {
   return WrappedComponent.displayName || WrappedComponent.name || 'Component'
 }
 
-function checkStateShape(componentDisplayName, stateProps, dispatch) {
+function checkStateShape(componentDisplayName, props, methodName) {
   invariant(
-    isPlainObject(stateProps),
-    '`%s %sToProps` must return an object. Instead received %s.',
+    isPlainObject(props),
+    '`%s %s` must return an object. Instead received %s.',
     componentDisplayName,
-    dispatch ? 'mapDispatch' : 'mapState',
-    stateProps
+    methodName,
+    props
   )
-  return stateProps
+  return props
 }
 
 // Helps track hot reloading.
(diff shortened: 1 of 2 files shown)


src/components/connect.js
```diff
@@ -45,11 +46,12 @@ export default function connect(mapStateToProps, mapDispatchToProps, mergeProps,
   // Helps track hot reloading.
   const version = nextVersion++
 
-  function computeMergedProps(stateProps, dispatchProps, parentProps) {
+  function computeMergedProps(componentDisplayName, stateProps, dispatchProps, parentProps) {
     const mergedProps = finalMergeProps(stateProps, dispatchProps, parentProps)
     invariant(
```
gaearon: We should refactor so that this check also occurs inside `checkStateShape()`. I would pass `methodName` (e.g. `'mapStateToProps'`, `mapDispatchToProps'` or `'mergeProps'`) instead of the existing boolean `dispatch` argument.

src/components/connect.js
```diff
@@ -18,10 +18,11 @@ function getDisplayName(WrappedComponent) {
...
 
-function checkStateShape(stateProps, dispatch) {
+function checkStateShape(componentDisplayName, stateProps, dispatch) {
   invariant(
     isPlainObject(stateProps),
-    '`%sToProps` must return an object. Instead received %s.',
+    '`%s %sToProps` must return an object. Instead received %s.',
+    componentDisplayName,
     dispatch ? 'mapDispatch' : 'mapState',
     stateProps
```
gaearon: Let’s also just call this argument `props`.

src/components/connect.js
```diff
@@ -88,7 +90,7 @@ export default function connect(mapStateToProps, mapDispatchToProps, mergeProps,
           this.finalMapStateToProps(state, props) :
           this.finalMapStateToProps(state)
 
-        return checkStateShape(stateProps)
+        return checkStateShape(this.constructor.displayName, stateProps)
```
gaearon: Rather than pass `this.constructor.displayName` in many places, let’s just calculate `var connectDisplayName = ...` right away and close over it everywhere we used `this.constructor.displayName` before this PR. Then we would be able to close over it in `checkStateShape` as well without even passing it.

chentsulin: Should we use var before `wrapWithConnect`? Or move `checkStateShape` into `wrapWithConnect`? `checkStateShape` can't use `connectDisplayName` directly because they live in the different scope.

gaearon: I'd move `checkStateShape` into `wrapWithConnect`.

chentsulin: Done. I move `checkStateShape` and also `computeMergedProps` into `wrapWithConnect`.

gaearon:
Thanks!

hogganbeck:
Thanks for this. Knee deep in containers. Very helpful.

sorenhoyer:
I guess this problem is back. At least I just got that error message with react 15.6.1 :/

PR #338, opened 2016-04-06
Fix #337, update impure component with custom mergeProps

### LINKED ISSUE #337: Specifying mergeProps causes impure connected components to not update
When an impure connected component is re-rendered due to its parent's state changing, its `shouldComponentUpdate` returns true, but if a non-default `mergeProps` is supplied then `checkMergedEquals` will also be true, causing the component to only update if its `nextMergedProps` have changed from the previous `mergedProps`.

In this [JSFiddle example](http://jsfiddle.net/mz2ajLko/1/), I would expect the root component's `setState` to trigger a re-render of the `PartComponent` and both of the impure components, but since the second impure component has a custom `mergeProps`, it doesn't update.

```
let store = Redux.createStore(() => ({}))

class PartComponent extends React.Component {
  render() {
    return <span>{ new Date().getTime() }</span>
  }
}

const Impure_DefaultMerge_Connected_PartComponent = ReactRedux.connect(
	null, null, null, { pure: false }
)(PartComponent)

const Impure_CustomMerge_ConnectedPartComponent = ReactRedux.connect(
	null, null, () => ({}), { pure: false }
)(PartComponent)

const Pure_DefaultMerge_ConnectedPartComponent = ReactRedux.connect(
	null, null, null, { pure: true }
)(PartComponent)

const Pure_CustomMerge_ConnectedPartComponent = ReactRedux.connect(
	null, null, () => ({}), { pure: true }
)(PartComponent)

class RootComponent extends React.Component {
  render() {
    return <div>
        <PartComponent /> PartComponent <br />
        <Impure_DefaultMerge_Connected_PartComponent /> Impure_DefaultMerge_Connected_PartComponent <br />
        <Impure_CustomMerge_ConnectedPartComponent /> Impure_CustomMerge_ConnectedPartComponent <br />
        <Pure_CustomMerge_ConnectedPartComponent /> Pure_CustomMerge_ConnectedPartComponent <br />
        <Pure_DefaultMerge_ConnectedPartComponent /> Pure_DefaultMerge_ConnectedPartComponent <br />
      </div>
  }
  componentDidMount() {
    setInterval(() => this.setState({}), 777)
  }
}

ReactDOM.render(
  <ReactRedux.Provider store={store}>
    <RootComponent />
  </ReactRedux.Provider>,
  document.getElementById('container')
)
```

tgriesser:
Needed to check if the component is pure to determine whether we can rely on `checkMergedEquals` behavior in setting the updated merged props. Fixes #337, test included.

### DIFF
--- a/src/components/connect.js
+++ b/src/components/connect.js
@@ -39,8 +39,8 @@ export default function connect(mapStateToProps, mapDispatchToProps, mergeProps,
     mapDispatchToProps || defaultMapDispatchToProps
 
   const finalMergeProps = mergeProps || defaultMergeProps
-  const checkMergedEquals = finalMergeProps !== defaultMergeProps
   const { pure = true, withRef = false } = options
+  const checkMergedEquals = pure && finalMergeProps !== defaultMergeProps
 
   // Helps track hot reloading.
   const version = nextVersion++
--- a/test/components/connect.spec.js
+++ b/test/components/connect.spec.js
@@ -1639,5 +1639,38 @@ describe('React', () => {
       expect(mapStateCalls).toBe(2)
       expect(renderCalls).toBe(1)
     })
+
+    it('should update impure components with custom mergeProps', () => {
+      let store = createStore(() => ({}))
+      let renderCount = 0
+
+      @connect(null, null, () => ({ a: 1 }), { pure: false })
+      class Container extends React.Component {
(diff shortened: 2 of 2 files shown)


gaearon:
Big thanks for fixing this so quickly.

PR #343, opened 2016-04-08
Support React 15 stable

trevordmiller:


gaearon:
Thanks but the current range is sufficient. Please see my comment in https://github.com/reactjs/react-redux/pull/342#issuecomment-207642767.

PR #345, opened 2016-04-10
Use callback ref instead of string

jisaacks:
Not sure if you care about this, but I was browsing the source code and noticed you were still using a string ref. I just updated it to a callback ref.


gaearon:
Good call. I know some people may be relying on exact ref name in their test cases so I’ll probably hold this off until the next major.

timdorr:
Superseded by #416. (Sorry, JD!)

jisaacks:
lol @timdorr, no worries :)

Issue #300, opened 2016-02-23
Potential connect() optimization


gaearon:
As described in https://github.com/reactjs/redux/issues/1437#issuecomment-187763637, we can try introducing a separate code path for the fast case where neither state props nor dispatch props depend on the own props. In this case the problem from #99 seems irrelevant and we can do the fast check *before* calling `setState()`.

I am not convinced that this is the best solution however. It basically forces computations to happen in the `handleChange()` hook which React has no control over. If in the future React learns to prioritize updates, it will not be able to schedule this work for later, but if we keep things as they are today, it will be able to do this just by postponing `render()`.

matianfu:
Will another callback as connect()'s parameter do the job? But I haven't yet figure out why you choose a connect() function, rather than a class to be sub-classed and mixed in to construct the container class. In this case, adding feature will be more smoothly without changing any existing code.

gaearon:
>But I haven't yet figure out why you choose a connect() function, rather than a class to be sub-classed and mixed in to construct the container class. In this case, adding feature will be more smoothly without changing any existing code.

Inheritance is not really a thing in React ecosystem, and for good reasons. It makes it too easy to break consuming packages, introduces naming conflicts between methods, is not composable, and in general exposes too many implementation details.

gaearon:
>Will another callback as connect()'s parameter do the job?

Why would we need it? I meant changing the internal implementation without touching the API. But as I said, I have concerns even about that.

matianfu:
> Why would we need it? I meant changing the internal implementation without touching the API. But as I said, I have concerns even about that.

I understand that you'd like to keep connect() simple. But it really does a lot of thing and hardly be considered as "do one thing, do it well"

Existing argument: mapState is for calculating derived value from store state. mapDispatch is providing actions.

The most distinct nature of React, imho, is not the much talked asynchronous. It is essentially an Inversion of Control Pattern. All life cycle method and render() is called by React, not by app code. The only (almost) thing app code can do, is to call the setState() method.

This is something quite similar to Mark & Sweep garbage collector. There are two stages, in the first stage all referenced objects are marked. In the second stage all unmarked objects are recycled.

From viewpoint of app code, React render can also be considered as two stages. The first stage can also be called "mark", which means, mark the components which should be refreshed. The second stage is the React's reconciliation.

In current status of code. There is no selective mark at all. All container/stateful component are marked. This is pointless.

[...]

PR #348, opened 2016-04-12
Bail out if stateProps can be calculated early and did not change

### LINKED ISSUE #1437: Dispatch more than once in non-batched update cause incorrect result with `useSelector`

Report a bug.

I'v created a reproducing repo here https://github.com/malash/react-redux-issue-ref/blob/master/reduxBug.js

This repo use both `react-redux` and `react-native`. 
```jsx
<Button
  title="[BUG] Click Me (setTimeout)"
  onPress={() => {
    setTimeout(() => {
      dispatch({type: 'NOOP'});
      dispatch({type: 'TOGGLE'});
    }, 0);
  }}
/>
```

When using `setTimeout`'s callback to dispatch multi times, the `useSelector` return incorrect value. The reason why I use `setTimeout` is to ensure these dispatch calling are non-batched.

```jsx
const selector = state => ({
	bool: state.bool,
});

const ReduxBugParent = () => {
  const {bool} = useSelector(selector);
  // ...
  <Text>bool from useSelector is {JSON.stringify(bool)}</Text>
  <Text>bool from store.getState is {JSON.stringify(boolFromStore)}</Text>
}
```

After click the button different selector result are shown. the `bool` are always different with `boolFromStore`.

Screen Recording 2019-10-29 at 11 05 55 PM

1. Clone the repo https://github.com/malash/react-redux-issue-ref
2. [Install React Native CLI](https://facebook.github.io/react-native/docs/getting-started)
3. Run `react-native run-ios`
4. Click buttons and see rendered values.

This bug can only reproduce on non-web environment, maybe it is related to https://github.com/reduxjs/react-redux/issues/1436

I believe it should works on React Native as well as React DOM.

react@16.9.0
react-native@0.61.2
react-redux@7.1.1
redux@4.0.4

**cc**

@byunicorn

gaearon:
This should fix https://github.com/reactjs/redux/issues/1437 and #300.

If the component doesn’t care about `ownProps`, we can bail out of calling `setState()`.
This PR implements this optimization.

As a reminder, we can’t do this for components that *do* care about `ownProps` due to #99.

----------------

Above, I said:

>As a reminder, we can’t do this for components that *do* care about `ownProps` due to #99.

*However* in many cases it’s possible to convert a component that needs `ownProps` to a component that doesn’t.

How? In #279, we added the ability to specify factories instead of `mapStateToProps` and `mapDispatchToProps`. Curiously, these factories *themselves* do receive `state` and `ownProps` as arguments. Of course, they are only invoked once, **but if you’re only using `ownProps` to read an ID** (which is a common case in large lists) and use stable `key`s, you should be able to change

```js
function mapStateToProps(state, ownProps) {
  return {
    item: state.items[ownProps.id]
  }
}
```

into

```js
function makeMapStateToProps(initialState, initialOwnProps) {
  var id = initialOwnProps.id
  return function mapStateToProps(state) {
    return {
      item: state.items[id]
    }
  }
}
```

See? Obviously `id` would never get updated, but we don’t need it to (in this particular case). And we can always add `ownProps` later (of course at the performance cost).

I think it’s a neat trick, and it will benefit from merging this optimization, since this optimization is relevant to any `mapStateToProps` that doesn’t depend on `ownProps`.

### DIFF
--- a/src/components/connect.js
+++ b/src/components/connect.js
@@ -219,6 +219,7 @@ export default function connect(mapStateToProps, mapDispatchToProps, mergeProps,
         this.mergedProps = null
         this.haveOwnPropsChanged = true
         this.hasStoreStateChanged = true
+        this.haveStatePropsBeenPrecalculated = false
         this.renderedElement = null
         this.finalMapDispatchToProps = null
         this.finalMapStateToProps = null
@@ -229,13 +230,21 @@ export default function connect(mapStateToProps, mapDispatchToProps, mergeProps,
           return
         }
 
-        const prevStoreState = this.state.storeState
         const storeState = this.store.getState()
-
-        if (!pure || prevStoreState !== storeState) {
-          this.hasStoreStateChanged = true
-          this.setState({ storeState })
+        const prevStoreState = this.state.storeState
+        if (pure && prevStoreState === storeState) {
+          return
         }
+
+        if (pure && !this.doStatePropsDependOnOwnProps) {
+          if (!this.updateStatePropsIfNeeded()) {
+            return
+          }
+          this.haveStatePropsBeenPrecalculated = true
+        }
+
+        this.hasStoreStateChanged = true
+        this.setState({ storeState })
       }
 
       getWrappedInstance() {
@@ -251,11 +260,13 @@ export default function connect(mapStateToProps, mapDispatchToProps, mergeProps,
         const {
           haveOwnPropsChanged,
           hasStoreStateChanged,
+          haveStatePropsBeenPrecalculated,
           renderedElement
         } = this
 
         this.haveOwnPropsChanged = false
         this.hasStoreStateChanged = false
+        this.haveStatePropsBeenPrecalculated = false
 
         let shouldUpdateStateProps = true
         let shouldUpdateDispatchProps = true
@@ -269,7 +280,9 @@ export default function connect(mapStateToProps, mapDispatchToProps, mergeProps,
 
         let haveStatePropsChanged = false
         let haveDispatchPropsChanged = false
-        if (shouldUpdateStateProps) {
+        if (haveStatePropsBeenPrecalculated) {
+          haveStatePropsChanged = true
+        } else if (shouldUpdateStateProps) {
           haveStatePropsChanged = this.updateStatePropsIfNeeded()
         }
         if (shouldUpdateDispatchProps) {
--- a/test/components/connect.spec.js
+++ b/test/components/connect.spec.js
@@ -1082,7 +1082,6 @@ describe('React', () => {
       )
       spy.destroy()
 
-
       spy = expect.spyOn(console, 'error')
       TestUtils.renderIntoDocument(
         <ProviderMock store={store}>
@@ -1545,6 +1544,51 @@ describe('React', () => {
       expect(mapStateCalls).toBe(2)
       // But render is not because it did not make any actual changes
       expect(renderCalls).toBe(1)
+    })
+
+    it('should bail out early if mapState does not depend on props', () => {
+      const store = createStore(stringBuilder)
+      let renderCalls = 0
+      let mapStateCalls = 0
+
+      @connect(state => {
+        mapStateCalls++
+        return state === 'aaa' ? { change: 1 } : {}
+      })
+      class Container extends Component {
+        render() {
+          renderCalls++
+          return <Passthrough {...this.props} />
+        }
+      }
+
+      TestUtils.renderIntoDocument(
+        <ProviderMock store={store}>
+          <Container />
+        </ProviderMock>
+      )
+
+      expect(renderCalls).toBe(1)
+      expect(mapStateCalls).toBe(1)
+
+      const spy = expect.spyOn(Container.prototype, 'setState').andCallThrough()
+
+      store.dispatch({ type: 'APPEND', body: 'a' })
+      expect(mapStateCalls).toBe(2)
+      expect(renderCalls).toBe(1)
+      expect(spy.calls.length).toBe(0)
+
+      store.dispatch({ type: 'APPEND', body: 'a' })
+      expect(mapStateCalls).toBe(3)
+      expect(renderCalls).toBe(1)
+      expect(spy.calls.length).toBe(0)
+
+      store.dispatch({ type: 'APPEND', body: 'a' })
+      expect(mapStateCalls).toBe(4)
+      expect(renderCalls).toBe(2)
+      expect(spy.calls.length).toBe(1)
+
+      spy.destroy()
     })
 
     it('should allow providing a factory function to mapStateToProps', () => {


gaearon:
cc @slorber, @ellbee, @tgriesser who might be interested in this

ellbee:
This one looks like a clear win to me, but I wonder if it would be worth working on a benchmark suite again like the one started in #104? I've used benchmark.js before, but not for testing UI code. Is it even really feasible?

gaearon:
For now, I’m testing against https://github.com/mweststrate/redux-todomvc manually.
My work in progress on optimizing app itself to Redux patterns is in https://github.com/mweststrate/redux-todomvc/pull/1.

ellbee:
Yeah, I have just been reading through that issue. It is what made me start thinking about it. I might have a play around with it.

gaearon:
:+1:

gaearon:
Oh wait, I’m supposed to emoji your post inline.

slorber:
@gaearon this is great news that connect not using component's props get faster :)

However I'm not sure the current implementation is the best we could do:

- The factory method works but is kind of unintuitive API at first
- If a connected component do care about ownProps, and these props can change (ie can't use the factory), then the component will setState even if the props don't change often


Instead of 

```javascript
function makeMapStateToProps(initialState, initialOwnProps) {
  var id = initialOwnProps.id
  return function mapStateToProps(state) {
    return {
      item: state.items[id]
    }
  }
}

connect(makeMapStateToProps)(Component)
```

I would rather have something like:


```javascript
connect({
   mapperPropsSelector: ownProps => {id: ownProps.id},
   mapper: (state,{id}) => state.items[id]
})(Component)
```

(this is probably not the best API but you get the idea)

What it permits here is to make it clear on what the mapping function has to rely on to do its job, so you can more easily detect when props needed for the mapping have changed and run the mapper only when needed


And I think, by default, not providing any ownProps to the mapping function (ie no selector provided) would actually encourage people to write more optimized code by default (unlike factories which would be opt-in optimization), but it would be a breaking change


(This is related to my proposal  for more flexible subscriptions here: https://github.com/reactjs/react-redux/issues/269#issuecomment-175904958)

gaearon:
I’m confused: how does accepting a selector for props solve the problem of `setState()`? As soon as something depends on `ownProps`, due to #99, we are forced to delay its calls until `render()` at which point we have already paid the price of `setState()`.

timbur:
Sorry if I'm going on a tangent with this comment but I'll share some performance optimizations I implemented earlier this month when I rewrote [`react-redux-provide`](https://github.com/loggur/react-redux-provide).  These optimizations might not perfectly translate to `react-redux` but perhaps similar algorithms could be implemented.

I added a `watch` method to the store which accepts a `reducerKey` and a `callback` function to be called whenever the reducer returns a different state (via a reducer enhancer).  The declarative nature of `react-redux-provide` makes it possible to automatically efficiently watch for changes to the relevant reducers, and in which case, a `doUpdate` flag is raised so than we can simply call `forceUpdate` after the action has completed, since we already know some relevant state has changed.  The wrapped component's props are cached and updated accordingly per reducer.  This means no need for shallow comparisons or loops, and `shouldComponentUpdate` can always return false (unless the component receives new props of its own, of course).  And so updates should be near instantaneous as they, in theory, only require a few clock cycles per relevant component - i.e., new state -> relevant components -> update.

Condensed version:
```js
for (let reducerKey of reducerKeys) {
  this.unsubscribe.push(
    store.watch(                  // when some reducer returns a new state
      reducerKey, nextState => {  // update wrapped component's props
        this.componentProps[reducerKey] = nextState;
        this.doUpdate = true;     // and we know for sure we should update
      }
    )
  );
}

this.unsubscribe.push(
  store.subscribe(() => {         // after action has completed
    if (this.doUpdate) {          // simply check for doUpdate flag
      this.forceUpdate();
    }
  })
);
```

See the full version where the same method is used for props derived from some combination of state and own props:
https://github.com/loggur/react-redux-provide/blob/master/src/provide.js#L93-L212

To elaborate a little bit more, deriving props from a combination of state and own props looks like this:
```js
// here we'll provide some item within a list
const merge = {
  item: {           // this is run only for components with an `item` propType
    keys: ['list'], // the provided `item` depends on the state of the `list`
    get(state, props, context) {
      // this function is executed only when `list` changes
      // and components will be updated only when the `item` has changed
      const { list } = state;
      const { itemIndex } = props;

      return list[itemIndex] || null;
    }
  }
};
```

Performance tests are coincidentally next on my todo list.  I was planning on seeing how things perform when incorporated into [`dbmon`](https://github.com/ryanflorence/reactconf-2015-HYPE/tree/master/demos/01-dbmon), but I think I'll also run some benchmarks on [`react-redux-provide/examples/todomvc`](https://github.com/loggur/react-redux-provide/tree/master/examples/todomvc).

slorber:
> I’m confused: how does accepting a selector for props solve the problem of setState()? As soon as something depends on ownProps, due to #99, we are forced to delay its calls until render() at which point we have already paid the price of setState().

@gaearon sorry it's hard to think well on such code late in the night :) My proposal may not prevent the `setState` cost finally.

The problem I see with current code is that for example if this component changes from

```javascript
<Todo id={123} className="someName"/>
```

 to 

```javascript
<Todo id={123} className="anotherClassName"/>
```

then yes we necessarily have to do a `setState` because ownProps have changed and we must render the child. The problem is that here our `mapStateToProps` is relying on ownProps but actually does only care about the `id`, and not `className`.

This line:

```javascript
shouldUpdateStateProps = hasStoreStateChanged || (
            haveOwnPropsChanged && this.doStatePropsDependOnOwnProps
          )
```

it will make `mapStateToProps` be run in such a case, while it could have been avoided if we could have known that `mapStateToProps` is not interested to receive the `className` property but only the Todo id.


Also, not a big deal, but not sure `this.doStatePropsDependOnOwnProps = this.finalMapStateToProps.length !== 1` is relyable if someone writes mapStateToProps with `arguments[0]` (yes it is unlikely... :p)

joonhyublee:
@gaearon Here is [my use case](https://github.com/yelouafi/redux-saga/issues/241#issuecomment-207202589), in which early bailout yields significant performance boost.

With regards to @slorber's concern that props *can* change, I encountered a similar issue, and I did this hack to get around it.

Let's say we have:
```javascript
//factory for map state to props, as per Dan's suggestion above
const makeMapStateToProps = (initialState, initialProps) => {
  const { userID } = initialProps; //props that's not expected to change (often)

    const userNickname = state.users[userID].nickname; //slice of state dependent on props
    const userAvatar = state.users[userID].avatar; //slice of state dependent on props

    return {
      userNickname,
      userAvatar
    };
  };

  return mapStateToProps;
}

const UserComponent = React.createClass({
  // component implementation
});

export default connect(makeMapStateToProps)(UserComponent);
```
Of course, userID isn't expected to change, but there may be some cases where it does. For example, the above component may be a react-router route component, which gets its userID from the pathname: `http://example.com/userpage/cat`. When the path changes to `http://example.com/userpage/dog` the component holds on to stale userID (cat) and this causes problems. (because react router doesn't re-mount for the same route)

I *could* revert back to `mapStateToProps (state, ownProps)`, but I would loose the performance boost gained from the early bailout. So instead, I do this:

```javascript
//instead of this,
export default connect(makeMapStateToProps)(UserComponent);

//do this:
const ConnectedComponent = connect(makeMapStateToProps)(UserComponent);
export default (props) => (
  <ConnectedComponent key={ props.userID }/>
);
```
Because of key, when userID is changed, the old component is completely unmounted, and new one is mounted instead, with the updated props. It's a bit hacky, but I think this is an acceptable trade-off. I guess for multiple props that aren't expected to change often I can stringify as a key:
```javascript
export default (props) => (
  <ConnectedComponent key={ props.userID + '_' + props.anotherSparselyChangedProp }/>
);
```
but that .. is beyond ugly. It does bother me there are so many different steps within different levels involved in simply invalidating and updating a component, but was the only way I could make this work.

slorber:
nice trick @joonhyublee :D 

Maybe this usecase will be easier to handle once the code of connect becomes much simpler, and it might with https://github.com/reactjs/react-redux/issues/368

PR #349, opened 2016-04-13
more test for shallowEqual

haohcraft:
Hey @gaearon ,
Something I don't understand.
Why the function `e1` and `_.get` are treated differently? Is it also due to different identity of function, or something else?

### DIFF
--- a/test/utils/shallowEqual.spec.js
+++ b/test/utils/shallowEqual.spec.js
@@ -1,5 +1,6 @@
 import expect from 'expect'
 import shallowEqual from '../../src/utils/shallowEqual'
+import _ from 'lodash'
 
 describe('Utils', () => {
   describe('shallowEqual', () => {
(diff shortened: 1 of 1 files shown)


gaearon:
I don’t think the difference you seem to talk about has relation to functions and their identities.

Here, you calculate `eVal` once, so `.e` field of both objects points to the same object. This is why the shallow comparison returns `true`.

```js
      const e1 = function () {return { a: 2 }}
      const eVal = e1()
      expect(
        shallowEqual(
          { a: 1, b: 2, c: o, d, e: eVal },
          { a: 1, b: 2, c: o, d, e: eVal }
        )
      ).toBe(true) 
```

Here, you call `e1()` two times, so it produces two different `{ a: 2 }` objects. `{}` notation creates a new object every time in JavaScript. So the first `.e` is not referentially equal to the second `.e`, even though they are deeply equal. This is why the shallow comparison returns `false`.

```js
      const e1 = function () {return { a: 2 }}
      expect(
        shallowEqual(
          { a: 1, b: 2,  e: e1() },
          { a: 1, b: 2,  e: e1() }
        )
      ).toBe(false) 
```

haohcraft:
thanks for such helpful explanation

Issue #353, opened 2016-04-14
Switch eslint off eslint-config-rackt


adrian-bobev:
The link to the eslint configuration is wrong. I guessed it should be https://github.com/pwmckenna/eslint-config-rackt
because in the package.json it looks that this package is used.

mxstbr:
I *think* the repo for that doesn't exist anymore, but maybe it got moved and I missed it – @mjackson any input here?

gaearon:
Yeah it’s dead and we plan to switch back to airbnb: https://github.com/reactjs/redux/pull/1559

mxstbr:
I've closed #354, but let's keep this issue open to remember to change the documentation when reactjs/redux#1559 is merged!

PR #354, opened 2016-04-14
Update CONTRIBUTING.md broken link

adrian-bobev:
Fixes https://github.com/reactjs/react-redux/issues/353


mxstbr:
The new link here is wrong too, the repository simply doesn't exist anymore and we'll switch back to airbnb. See @gaearon's comment in #353 – thanks so much for submitting this PR though, we appreciate it!

PR #358, opened 2016-04-18
Update Provider.js

koistya:


gaearon:
Any particular reason you prefer this style?

koistya:
Less variable declarations, smaller code

gaearon:
I wouldn’t think either is more efficient but I’m OK with this, it does read easier on eyes.

PR #373, opened 2016-05-01
Connect refactor to use transactional setState() 

slorber:
This is a pull request for https://github.com/reactjs/react-redux/issues/368

Not to be merged yet, as the code is not clean but it's enough to get some feedback, as all the existing tests are passing

### DIFF
--- a/src/components/connect.js
+++ b/src/components/connect.js
@@ -257,7 +257,7 @@ export default function connect(mapStateToProps, mapDispatchToProps, mergeProps,
         }
 
         // See #86 and #368
-        // We need to avoid computng state multiple times per batch so we only run the last one
+        // We need to avoid computing state multiple times per batch so we only run the last one
         this.lastSetStateFunction = setStateFunction
         this.setState((prevState, props) => {
           const isLast = this.lastSetStateFunction === setStateFunction
--- a/src/components/connect.js
+++ b/src/components/connect.js
@@ -73,8 +73,22 @@ export default function connect(mapStateToProps, mapDispatchToProps, mergeProps,
     }
 
     class Connect extends Component {
-      shouldComponentUpdate() {
-        return !pure || this.haveOwnPropsChanged || this.hasStoreStateChanged
+
+      shouldComponentUpdate(nextProps,nextState) {
+
+        if ( this.skipNextRender ) {
+          this.skipNextRender = false;
+          return false;
+        }
+
+        /*
+        console.log("shouldComponentUpdate",!pure || nextState.mergedPropsUpdated)
+        console.log("shouldComponentUpdate",!pure || nextState.mergedPropsUpdated)
+        console.log("shouldComponentUpdate",nextState)
+        console.log("shouldComponentUpdate nextState.mergedPropsUpdated",nextState.mergedPropsUpdated)
+        */
+          console.log("################## shouldComponentUpdate",nextState.mergedPropsUpdated)
+          return !pure || nextState.mergedPropsUpdated
       }
 
       constructor(props, context) {
@@ -88,9 +102,7 @@ export default function connect(mapStateToProps, mapDispatchToProps, mergeProps,
           `Either wrap the root component in a <Provider>, ` +
           `or explicitly pass "store" as a prop to "${connectDisplayName}".`
         )
-
-        const storeState = this.store.getState()
-        this.state = { storeState }
+        this.state = { }
         this.clearCache()
       }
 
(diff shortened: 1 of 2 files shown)


test/components/connect.spec.js
```diff
@@ -352,9 +352,11 @@ describe('React', () => {
         componentDidMount() {
           // Simulate deep object mutation
           this.state.bar.baz = 'through'
-          this.setState({
+
+          // TODO before setTimeout was not required. can I modify this test safely? it seems not a bid deal
+          setTimeout(() => this.setState({
```
slorber: This is the only modified test @gaearon but I think it's ok to modify

test/components/connect.spec.js
```diff
@@ -1430,13 +1432,7 @@ describe('React', () => {
...
       const mapDispatchSpy = expect.createSpy().andReturn({})
 
-      class ImpureComponent extends Component {
-        render() {
-          return <Passthrough statefulValue={this.props.value} />
-        }
-      }
-
-      const decorator = connect(
+      @connect(
```
slorber: This just refactors the test with @connect but the test is the same

src/components/connect.js
```diff
@@ -208,13 +190,22 @@ export default function connect(mapStateToProps, mapDispatchToProps, mergeProps,
         }
       }
 
-      componentDidMount() {
+      componentWillMount() {
```
slorber: Used WillMount instead of DidMount because now on the first render we have to trigger `handleChange` otherwise mergedProps don't get calculated. Is it a problem?

src/components/connect.js
```diff
@@ -73,8 +74,23 @@ export default function connect(mapStateToProps, mapDispatchToProps, mergeProps,
     }
 
     class Connect extends Component {
-      shouldComponentUpdate() {
-        return !pure || this.haveOwnPropsChanged || this.hasStoreStateChanged
+
+      shouldComponentUpdate(nextProps,nextState) {
+
```
slorber: Kind of hacky but if componentWillReceiveProps receive same props as prevProps, then we don't want to redo a shallowEqual here. I could have put the props in component state but not sure it would be better

test/components/connect.spec.js
```diff
@@ -1469,17 +1468,17 @@ describe('React', () => {
       const target = TestUtils.findRenderedComponentWithType(tree, Passthrough)
       const wrapper = TestUtils.findRenderedComponentWithType(tree, StatefulWrapper)
 
-      expect(mapStateSpy.calls.length).toBe(2)
-      expect(mapDispatchSpy.calls.length).toBe(2)
+      expect(mapStateSpy.calls.length).toBe(1)
```
slorber: I think the test is wrong and that the number of calls should be 1 instead of 2. I think there's a bug in current implementation that does an additional non-required call

src/components/connect.js
```diff
@@ -47,7 +37,6 @@ export default function connect(mapStateToProps, mapDispatchToProps, mergeProps,
 
   const finalMergeProps = mergeProps || defaultMergeProps
   const { pure = true, withRef = false } = options
-  const checkMergedEquals = pure && finalMergeProps !== defaultMergeProps
```
slorber: @tgriesser can you take a look at that plz? I've been able to make the tests pass without using this variable. Is there a missing test-case?

src/components/connect.js
gaearon: Yes, this would mean a memory leak on server rendering.

src/components/connect.js
```diff
@@ -208,13 +173,23 @@ export default function connect(mapStateToProps, mapDispatchToProps, mergeProps,
         }
       }
 
+      componentWillMount() {
```
slorber: @gaearon I've been able to remove the store subscription here, but I still need to be able to trigger the computation from here, otherwise nothing can be put into state for the first render and we are back to doing the computation in render()

is that ok?

slorber:
@gaearon did you have time to check my PR?

I've tried to run benchmarks of @mweststrate . Not sure to have done them correctly, but I actually see increased performances :) It's not huge, and maybe I did something wrong during the benchmark.

```
# Change 1:
- Before: ~50ms
- After: ~45ms

# Add 1:
- Before: ~110ms
- After: ~103ms
```

Maybe you can run them yourself and confirm?

I've published the benchmark app [before](http://slorber.github.io/connect-refactor/before.html] and [after](http://slorber.github.io/connect-refactor/after.html) so that anyone can take a look

gaearon:
Hey, great work. I’m super busy this couple of weeks so I’ll leave this hanging for now. If all tests are passing and you’re confident about it, we can cut a beta.

slorber:
thanks :)

I'm going on holiday 2 weeks until the 21 so you will have some time to check the code :)

I'm confident globally but not really about the `checkMergedEquals` attribute that @tgriesser added. Don't really understand the usecase as all my tests are passing without using that, but it has probably be done for a good reason so...

src/components/connect.js
gaearon: I think the premise was that if we use `defaultMergeProps`, we *know* that it’s *not* going to be shallowly equal if either `stateProps` or `dispatchProps` changed. Therefore not even worth doing a `shallowEqual` check.

timdorr:
Superseded by #416

PR #389, opened 2016-05-24
add flow type definition file

nmn:
This adds a flow type declaration file so that anyone using it from NPM should just get the types 'for free'

It assumes the latest version of flow (0.25, though 0.24 should work too), and adds mostly adequate type definitions for both `connect` and `Provider`. Obviously most of the code is there to type `connect` which can get pretty tricky.

Known Issues: Flow correctly doesn't support stateless React components. The type definition for connect doesn't change that fact. (I tried to do a trick to make such that any stateless function component decorated with `connect` would just start getting type checked, but that didn't quite work in practice.)

### DIFF
--- a/lib/index.js.flow
+++ b/lib/index.js.flow
@@ -2,12 +2,14 @@
 type ConnectAll = <D, P, S, C: React$Component<D, P, S>, SP, DP, Dispatch: Function>(
   mapStateToProps: (state: Object) => SP,
   mapDispatchToProps: (dispatch: Dispatch) => DP,
+  mergeProps: null | void,
   options?: {pure?: boolean, withRef?: boolean}
 ) => (component: Class<C>) => Class<React$Component<D, $Diff<$Diff<P, DP>, SP>, S>>
 
 type ConnectAllStateless = <P, SP, DP, Dispatch: Function>(
   mapStateToProps: (state: Object) => SP,
   mapDispatchToProps: (dispatch: Dispatch) => DP,
+  mergeProps: null | void,
   options?: {pure?: boolean, withRef?: boolean}
 ) => (component: (props: P) => any) => Class<React$Component<void, $Diff<$Diff<P, DP>, SP>, void>>
 
@@ -28,22 +30,28 @@ type ConnectMergedStateless = <P, SP, DP, MP, Dispatch: Function>(
 type ConnectNoState = <D, P, S, C: React$Component<D, P, S>, DP, Dispatch: Function>(
     mapStateToProps: null | void,
     mapDispatchToProps: (dispatch: Dispatch) => DP,
+    mergeProps: null | void,
     options?: {pure?: boolean, withRef?: boolean}
   ) => (component: Class<C>) => Class<React$Component<D, $Diff<P, DP>, S>>
 
 type ConnectNoStateStatless = <P, DP, Dispatch: Function>(
     mapStateToProps: null | void,
     mapDispatchToProps: (dispatch: Dispatch) => DP,
+    mergeProps: null | void,
     options?: {pure?: boolean, withRef?: boolean}
   ) => (component: (props: P) => any) => Class<React$Component<void, $Diff<P, DP>, void>>
 
-type ConnectDisptch = <D, P, S, C: React$Component<D, P, S>, SP, Dispatch: Function>(
+type ConnectDispatch = <D, P, S, C: React$Component<D, P, S>, SP, Dispatch: Function>(
   mapStateToProps: (state: Object) => SP,
+  mapDispatchToProps: null | void,
+  mergeProps: null | void,
   options?: {pure?: boolean, withRef?: boolean}
 ) => (component: Class<C>) => Class<React$Component<D, $Diff<$Diff<P, {dispatch: Dispatch}>, SP>, S>>
 
-type ConnectDisptchStateless = <P, SP, Dispatch: Function>(
+type ConnectDispatchStateless = <P, SP, Dispatch: Function>(
   mapStateToProps: (state: Object) => SP,
+  mapDispatchToProps: null | void,
+  mergeProps: null | void,
   options?: {pure?: boolean, withRef?: boolean}
 ) => (component: (props: P) => any) => Class<React$Component<void, $Diff<$Diff<P, {dispatch: Dispatch}>, SP>, void>>
 
@@ -60,8 +68,8 @@ declare export var connect
     & ConnectMergedStateless
     & ConnectNoState
     & ConnectNoStateStatless
-    & ConnectDisptch
-    & ConnectDisptchStateless
+    & ConnectDispatch
+    & ConnectDispatchStateless
     & ConnectDefault
     & ConnectDefaultStateless;
 declare export var Provider: ReactClass<{store: Object, children?: any}>;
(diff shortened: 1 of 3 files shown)


lib/index.js.flow
```diff
@@ -0,0 +1,69 @@
...
+    options?: {pure?: boolean, withRef?: boolean}
+  ) => (component: Class<C>) => Class<React$Component<D, $Diff<P, DP>, S>>
+
+type ConnectNoStateStatless = <P, DP, Dispatch: Function>(
+    mapStateToProps: null | void,
+    mapDispatchToProps: (dispatch: Dispatch) => DP,
+    options?: {pure?: boolean, withRef?: boolean}
+  ) => (component: (props: P) => any) => Class<React$Component<void, $Diff<P, DP>, void>>
+
+type ConnectDisptch = <D, P, S, C: React$Component<D, P, S>, SP, Dispatch: Function>(
```
gaearon: Should be `ConnectDispatch`?

lib/index.js.flow
```diff
@@ -0,0 +1,69 @@
...
+  mapStateToProps: (state: Object) => SP,
+  options?: {pure?: boolean, withRef?: boolean}
+) => (component: Class<C>) => Class<React$Component<D, $Diff<$Diff<P, {dispatch: Dispatch}>, SP>, S>>
+
+type ConnectDisptchStateless = <P, SP, Dispatch: Function>(
+  mapStateToProps: (state: Object) => SP,
+  options?: {pure?: boolean, withRef?: boolean}
```
gaearon: I think options are always forced to be the last argument. (I wanted to make passing them inconvenient because they enable escape hatches like `withRef` and `pure: false`.) So you probably don’t need so many overloads.

Only `connect(fn, null, null, options)` would work, not `connect(fn, options)`.

lib/index.js.flow
```diff
@@ -0,0 +1,69 @@
+/* @flow */
+type ConnectAll = <D, P, S, C: React$Component<D, P, S>, SP, DP, Dispatch: Function>(
+  mapStateToProps: (state: Object) => SP,
+  mapDispatchToProps: (dispatch: Dispatch) => DP,
```
gaearon: An important case that we want to support is when `mapDispatchToProps` is an object with action creators (`{ addTodo, removeTodo }`) rather than a function. Then injected props will be functions with the same names (`{ addTodo, removeTodo }`) but those functions would dispatch automatically.

lib/index.js.flow
nmn: Oh I didn't know that. Still the overloads are still needed because every combination changes the type of the returned value. Every overload in the file has a different return type! I'll change the type defs though.

lib/index.js.flow
nmn: Cool. Thankfully that's an easy thing to account for. Will fix.

nmn:
@gaearon To support the case when `mapDispatchToProps` is an object means I have to add even more overloads. As the type defs will change a bit.

Short explanation:
```
type ConnectAll = <D, P, S, C: React$Component<D, P, S>, SP, DP, Dispatch: Function>(
  mapStateToProps: (state: Object) => SP,
  mapDispatchToProps: (dispatch: Dispatch) => DP,
  mergeProps: null | void,
  options?: {pure?: boolean, withRef?: boolean}
) => (component: Class<C>) => Class<React$Component<D, $Diff<$Diff<P, DP>, SP>, S>>

type ConnectAllObj = <D, P, S, C: React$Component<D, P, S>, SP, DP, Dispatch: Function>(
  mapStateToProps: (state: Object) => SP,
  mapDispatchToProps: DP,
  mergeProps: null | void,
  options?: {pure?: boolean, withRef?: boolean}
) => (component: Class<C>) => Class<React$Component<D, $Diff<$Diff<P, {[key: $Keys<DP>]: any}>, SP>, S>>
```

The extra work comes as if you pass the object in, instead of the argument being the same type as that of the props, they have to be mapped to functions.

I'll try to add all the extra overloads and check if things still work. If all the cases don't work correctly, is there a subset of use cases you'd like to cover in the type definitions?

nmn:
Update: One other possibility is to use a more generic type:
```
type Connect = <D, P, S, C: React$Component<D, P, S>, SP, DP>(
  mapStateToProps?: (state: Object) => SP,
  mapDispatchToProps?: DP | (dispatch: Function) => DP,
  mergeProps?: (stateProps: SP, dispatchProps: DP, props: P) => Object,
  options?: {pure?: boolean, withRef?: boolean}
) => (component: Class<C>) => Class<React$Component<D, $Shape<P>, S>>
```

This has two downsides:

1. You have to use `undefined` to skip arguments instead of `null`.
2. The returned type is typed as strictly as the type defs defined in the current PR.

This type only checks that the props you HAVE provided have the right type. (e. g. It won't let you pass a string instead of a number) But it will let you completely skip required props and not warn you.

Do you think that is preferable?

nmn:
**Update:**

- I updated the type definitions to accept the options only as the 4th argument.
- I chose to not include the `mapDispatchToProps is an object` case as after much research, I couldn't find a way to define types for that correctly. I assume people using flow will be willing to write slightly more code to get type-safety.

agentcooper:
@nmn can you maybe give an example of the annotated connected component which is compatible with these type definitions?

btw, you can probably use `type ConnectOptions = {pure?: boolean, withRef?: boolean}` to decrease duplication.

agentcooper:
One more thing: you support `ownProps` param in `mergeProps` (as `origProps`), but not in `mapStateToProps` and `mapDispatchToProps`.

nmn:
@agentcooper 

1. The idea is that if you type your own component you shouldn't need anything extra. So the example is to just use connect like normal!

2. I think I missed the ownProps arg in the other two args. I'll add it in.

ccorcos:
I think its quite important to have some typesafey around state and dispatch. State isnt just any object, its a specific object you're expecting within the application. And dispatch isnt just a function, but a function that accepts specific actions. Perhaps connect could have a type signature like `Connects<State,Dispatch>` so that I can define my state as an object literal type and dispatch as only accepting specific actions.

agentcooper:
There are pretty good definitions available at https://github.com/reactjs/redux/pull/1887, for both redux and react-redux.

ccorcos:
thanks @agentcooper! those are some great lib defs!

nmn:
@agentcooper Those type definitions do look interesting. They are definitely more sound theoretically. But in the past, the most correct definitions have not worked for me in practice.

I'll take these for a spin and report back, now that Flow has been updated a few times.

timdorr:
Same as #538: Can this be rebased against the `next` branch? I don't think typings showing up in a patch or minor would follow semver, since things might break. Plus, it'll keep the complaints down. Thanks!

mull:
Has this been abandoned?

green-arrow:
Bump on this.

Also opened up a relevant issue on flow-typed for the shorthand notation of `mapDispatchToProps`: https://github.com/flowtype/flow-typed/issues/746

nmn:
This PR is on the back-burner as flow-typed already has the type-defs.

There are some problems with function components that I'm trying to get fixed.

lib/index.js.flow
```diff
@@ -0,0 +1,77 @@
+/* @flow */
+type ConnectAll = <D, P, S, C: React$Component<D, P, S>, SP, DP, Dispatch: Function>(
+  mapStateToProps: (state: Object) => SP,
```
apsavin: mapStateToProps can have additional parameter - own props of the connected component.

timdorr:
Closing for similar reasons to #814. Maintaining typings in-house is bad for DX.

green-arrow:
@timdorr - I think you want #815 (looks like #814 was closed due to a wrong branch). Thanks for the update on this!

PR #392, opened 2016-05-25
Update peer dependencies

mlucool:


gaearon:
Hi, thanks for the PR! However `^15.0.0-0` includes `15.0.0` and any future `15.x` versions, so this is not necessary.

PR #395, opened 2016-05-28
Add `displayName` option to connect()

epeli:
I'm using bunch of custom HOCs with `connect()` like

```js
function addPostTitle(postId) {
  return connect(
    state => ({title: state.posts[postId]}),
    null, null, {name: "addPostTitle"})
  );
}
```

This causes React devtools to be polluted with useless `Connect(Connect(Connect(Something)))` style components. This would allow to use more meaningful names.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -71,6 +71,7 @@ Instead, it *returns* a new, connected component class, for you to use.
 * [`options`] *(Object)* If specified, further customizes the behavior of the connector.
   * [`pure = true`] *(Boolean)*: If true, implements `shouldComponentUpdate` and shallowly compares the result of `mergeProps`, preventing unnecessary updates, assuming that the component is a “pure” component and does not rely on any input or state other than its props and the selected Redux store’s state. *Defaults to `true`.*
   * [`withRef = false`] *(Boolean)*: If true, stores a ref to the wrapped component instance and makes it available via `getWrappedInstance()` method. *Defaults to `false`.*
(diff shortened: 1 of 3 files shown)


gaearon:
I think I’d rather not add a separate API for this.
You can reassign it manually, so I think this would work best.

```js
var c = connect(...)
c.displayName = ...
```

epeli:
I know but it's gets really cumbersome if I want to have the original name in it:

```js
function addPostTitle(postId) {
  return Component => {
    var C = connect(state => ({title: state.posts[postId]}))(Component);
    C.displayName =`addPostTitle(${getDisplayName(Component)})`;
    return C;
  };
}
```

Also I need to the implement the `getDisplayName` or add dependency which implements it.

gaearon:
OK. Let’s rename it to `displayName`?

epeli:
> OK. Let’s rename it to displayName?

Yeah. It's more consistent with other APIs.

epeli:
PR updated. Also added a simple test case for it.

TigerC10:
Am I the only one that thinks that the `Connect` wrapper should just leverage the same `displayName` as the original wrapped component?

> @ianobermiller (FormidableLabs/radium#271 on [Jul 17, 2015](https://github.com/FormidableLabs/radium/issues/271#issuecomment-122227566))
> Can you imagine if every decorator did the same thing? `Radium(Relay(autobound(Button)))` 👎

timdorr:
Superseded by #416

PR #408, opened 2016-06-17
RR 2.0 example snippet

ev1stensberg:
Added RR 2.0 example

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -49,6 +49,22 @@ ReactDOM.render(
 )
 ```
 
+##### React Router 2.0
+
+```js
+ReactDOM.render(
+  <Provider store={store}>
+    <Router history={history}>
+      <Route path="/" component={App}>
+        <Route path="foo" component={Foo}/>
+        <Route path="bar" component={Bar}/>
+      </Route>
+    </Router>
+  </Provider>,
+  document.getElementById('root')
+)
+```
+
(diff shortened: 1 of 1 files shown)


docs/api.md
```diff
@@ -49,6 +49,22 @@ ReactDOM.render(
...
+      <Route path="/" component={App}>
+        <Route path="foo" component={Foo}/>
+        <Route path="bar" component={Bar}/>
+      </Route>
+    </Router>
+  </Provider>,
+  document.getElementById('root')
+)
+```
+
```
ev1stensberg: One line too much?

docs/api.md
```diff
@@ -49,6 +49,22 @@ ReactDOM.render(
...
+  <Provider store={store}>
+    <Router history={history}>
+      <Route path="/" component={App}>
+        <Route path="foo" component={Foo}/>
+        <Route path="bar" component={Bar}/>
+      </Route>
+    </Router>
+  </Provider>,
+  document.getElementById('root')
+)
```
ev1stensberg: One line too much+

ev1stensberg:
~ This snippet was taken from somewhere else in the Repo, but I think it is convenient to have it here as well.

ev1stensberg:
It's also the same, but people may think it is different from 1.0. Alternatively we can change `React Router 1.0` to `2.0`

timdorr:
I'll just rename this to remove the version part.

Issue #407, opened 2016-06-16
Completely rewrite connect() to offer advanced API, separate concerns, and (probably) resolve a lot of those pesky edge cases


jimbolla:
[I rewrote connect() and extracted a connectAdvanced()](https://github.com/jimbolla/react-redux/tree/connect-rewrite/) It now uses [reselect](https://github.com/reactjs/reselect) to handle merging/caching of the state/props/dispatch. It also extracts out a `connectAdvanced()` function that `connect()` just delegates to. 

I'm not quite ready to ready to submit a PR (still tweaking), but I'd like to suggest that this eventually become the new implementation.

It passes tests and linting, ~~with the only breaking change to the existing API being related to the "advanced scenarios" for factories that require setting another options flag. (I haven't figured out how to detect the factory methods automatically... will give it another go, but it's been tricky given how radically different my approach is.)~~ solved

Things I'm still need to do:
- [x] support factorying map*ToProps functions without an explcit flag
- [x] add more comments
- [ ] performance testing. the tests run slightly faster on my machine but we're talking insignificant amounts (250ms vs 230ms). do performance tests exist for this?
- [ ] reconcile with pending pull requests, as there would be zero chance of a successful merge. I've already covered the changes and new features in several of them, but some of the others will need reimplemented.
- [ ] keep tinkering. there's some functionality in `connect()` that I'd like to make reusable for someone who wants extend its functionality in userspace

markerikson:
The first concern would probably be that you're explicitly depending on Reselect.  While Reselect is heavily used in the Redux ecosystem, I'd be hesitant to make it a direct dependency.  (Of course, I'm also just an opinion, hardly the final arbiter.)

jimbolla:
What would be the cause for hesitation?

btipling:
The connect code is a bit hard to work with in tests. It would be really nice to be able to set withRef to true, but there's so much hidden inside that connect code, not accessible outside of it. Even the [`Connect` component](https://github.com/reactjs/react-redux/blob/master/src/components/connect.js#L75) itself is defined inside the `connect` function. That's just crazy to me, why was it written that way? It's taking important bits of information from the closure it's in and making it impossible to extend that `Connect` component or do anything with it to make testing easier.

So I think a bit of a refactor there would be nice but your own [`advancedConnect`](https://github.com/jimbolla/react-redux/blob/connect-rewrite/src/components/connectAdvanced.js) has this problem too. So I'm not sure this is the way to do it.

[...]

Issue #398, opened 2016-06-02
Wasteful operations when mapStateToProps has 2 arguments, store triggers change event but props and state unchanged


vlan-saxo:
As title suggests, when mapStateToProps is using both state and ownProps to select chunk of redux state, when store has some state updated, but the particular chunk of state which is returned by mapStateToProps has not been changed, connect produces wasteful render attempt. This is specifically bad in places when there is thousands of components wrapped with connect, or there is just couple of such components, store is updated quite frequently, but state of wrapped components is not changed. Issue is not reproduced when mapStateToProps is not using ownProps.

Performance degradation is seen in the task manager and in JS profiler, and becomes a visually noticeable problem on documents with complex DOM in Edge and Firefox. Measurements from React.addons.perf.printWasted show that on 10 seconds intervals the wasteful operations can take more than 1 second, the wasteful operations are happening inside Connect(Component).

Looking into profiling using Chrome profiler, a lot of time is spent inside ReactComponent.setState, looking into source code of connect.js, the problem is apparently in the handleChange.js, it does an unconditional this.setState when this.doStatePropsDependOnOwnProps is true. Perhaps an option to customize this behavior, e.g. by introducing support for custom comparer would be really nice.

Small sample project with reproduction has been set in https://github.com/vlan-saxo/react-redux-bug, use `npm install`, `npm start` and go to `http://localhost:8080/webpack-dev-server/`

larpo:
I ran into the same issue while going through react perftools results. There probably is a valid reason for the doStatePropsDependOnOwnProps check, but I think the docs should be more clear about this limitation.

Anyway, the workaround in my case was to use a custom mergeProps function, eg. changing from
```
connect(
    ({items}, {selected}) => ({item: items[selected]})
)(Component)
``` 
to
```
connect(
    ({items}) => ({items}),
    null,
    ({items}, _ , parentProps) => ({...parentProps, item: items[parentProps.selected] })
)(Component)
```

Looks ugly, but the perf issue is gone.

jimbolla:
I think this will be addressed by #407. I'm rewriting the Connect component to only fire setState() if the final props have changed. I'm seeing HUGE performance gains.

markerikson:
@jimbolla : yeah, this would be another good before/after comparison project.

gaearon:
>This is specifically bad in places when there is thousands of components wrapped with connect

[...]

Issue #368, opened 2016-04-25
Use transactional setState() inside connect()


gaearon:
We currently delay calling `mapStateToProps` and `mapDispatchToProps` in the common case because it is too early to supply `ownProps` inside `handleChange`, as those props may themselves be obtained from the previous version of the state by the parent component, and may not have been received as the new props just yet, so we bump into issues with staleness (#86).

Right now we call them inside `render()` so `ownProps` are always up-to-date but this is a bit odd, makes the code complicated and in some cases robs us of possible performance improvements, as caching the element isn’t quite as good as providing `shouldComponentUpdate` (#366).

One possible solution to this would be to **use the transactional `setState()` <s>overlord</s> overload**.
It is [described in the documentation](https://facebook.github.io/react/docs/component-api.html#setstate):

>It's also possible to pass a function with the signature function(state, props). This can be useful in some cases when you want to enqueue an atomic update that consults the previous value of state+props before setting any values. For instance, suppose we wanted to increment a value in state:

>```js
setState(function(previousState, currentProps) {
  return {myInteger: previousState.myInteger + 1};
});
>```

This looks pretty much exactly like our use case, and I have a feeling that this would let us use `setState` normally without having to resort to calling `mapStateToProps` and friends in `render()`. In this case, we would probably keep the merged props rather than the store state, in the local state.

We still want to keep the “fast path” that avoids `setState` altogether in `handleChange`, but other than that, the rest of the code should be vastly simplified by this, or at least become more conventional.

I’m still not 100% sure this is going to work, but I encourage anyone who wants to dig deeper into how React and Redux work together to give this a try. We have an extensive test suite that should cover any call count regressions. If you are working on this, please leave a comment in this issue.

amccloud:
Looking into it! I'm trying to wrap my head around what would no longer be necessary. Looks like this could clean up quite a bit of hand holding around `haveStatePropsBeenPrecalculated` and stateProps in `render()`.

amccloud:
Making progress. 8 failing test mostly around invocation counts. Will work on this more tomorrow.

gaearon:
Awesome! If you get stuck please feel free to share your work in progress so somebody can pick it up later.

slorber:
Really happy to see all those red lines on first commits :)

udnisap:
My approach was to use `haveMergedPropsChanged` as the flag for `shouldComponentUpdate` and call 
[...]

Issue #322, opened 2016-03-16
Initializing state dispatching actions in component, the right and the wrong way


srlopez:
Studying 'React Native' & 'Redux', I wanted to initialize a 'State' in componenetWillMount() function to see with the data is rendered in my component.

The component is a 'smart component' and  'State' and 'Actions' are mapped to props.

The first time you instantiate the component, the 'State' is empty and then an Action is dispatched in componetDidMount() to set some values in the 'State', which should re-invoke render () and show mutations in the screen.
Well, the truth is that render() is not called, but the state is updated, so is not a Redux issue

Instead, if I dispatch the action in a SetTimer even in componenetWillMount() then render() is invoked, and also the State is updated correctly.

I don't think this is a good way. 

Wich will be the tight place to dispatch actions to initialize State?

```
export default class extends Component {
  componentWillMount(){
    // This is binded to dispatch(...)
    // But this way the action is dispatched but the stete recived remains unchanged
    this.props.formActions.update('required','required value')
    // This way the actions is dispatched and re-render the component
    //setTimeout(() => { this.props.formActions.update('url','https://mi/url/de_prueba') }, 0);
  }

  render() {
    const { values } = this.props.formState;
    const { update, reset } = this.props.formActions;

    return (
      <View style={styles.container}>
        <Form values={this.props.formState.values} update={update} reset={reset}>
          <FTextInput name='url' label='Prueba url' placeholder='url'/>
          <FTextInput name='email' label='Prueba email' placeholder='email'/>
          <FTextInput name='required' label='Prueba required' placeholder='required'/>
        </Form>
        <EasyRow navigator={this.props.navigator} >
          <EasyLink label='Back'/>
        </EasyRow>
      </View>
    )
  }
}
`

```

Really thanks in advance

gaearon:
>Well, the truth is that render() is not called, but the state is updated, so is not a Redux issue

This actually *is* a React Redux issue because if what you say is true, it means the library erroneously bails out of causing a re-render when it should cause it.

Can you please submit a pull request with a failing test demonstrating this behavior?

srlopez:
Yes, Dan @gaearon,
This is the [repo](https://github.com/srlopez/rntest), and in the [src/pages/TestForm.js](https://github.com/srlopez/rntest/blob/master/src/pages/TetsForm.js) you can see:

```
componentWillMount(){
    // This component is binded to dispatch(...)
    // 1.-  this way the action is dispatched but the stete recived remains unchanged
[...]

PR #416, opened 2016-06-24
Rewrite connect() for better performance and extensibility

jimbolla:
# TL;DR

Rewrote `connect`, same basic API plus advanced options/API, all tests pass, roughly 8x faster, more modular/extensible design

# Overview

I rewrote `connect` into modular pieces because I wanted to be able to extend with custom behavior in my own projects. Now connect is a facade around `connectAdvanced`, by passing it a compatible `selectorFactory` function.

I also was able to greatly improve performance by changing the store subscriptions to execute top-down to work with React's natural flow of updates; component instances lower in the tree always get updated after those above them, avoiding unnecessary re-renders.

# Design/Architecture

I split the original `connect` into many functions+files to compartmentalize concepts for better readability and extensibility. The important pieces:

- components/
 - `connectAdvanced.js`: the HOC that connects to the store and determines when to re-render
 - `Provider.js`: (hasn't changed)
- selectors/
 - `connect.js`: composes the other functions into a fully-compatible API, by creating a selectorFactory and options object to pass to `connectAdvanced`.
 that performs memoiztion and detects if the first run returns another function, indicating a factory
 - `mapDispatchToProps.js`: used to create a selector factory from the `mapDispatchToProps` parameter, to be passed to `selectorFactory.js` as `initMapDispatchToProps`. Detects whether `mapDispatchToProps` is missing, an object, or a function
 - `mapStateToProps.js`: used to create a selector factory from the `mapStateToProps` parameter, to be passed to `selectorFactory.js` as `initMapStateToProps`. Detects whether `mapStateToProps` is missing or a function
 - `mergeProps.js`: used to create a selector factory from the `mergeProps` parameter, to be passed to `selectorFactory.js` as `initMergeProps`. Detects whether `mergeProps` is missing or a function.
- `selectorFactory.js`: given `dispatch`, `pure`, `initMapStateToProps`, `initMapDispatchToProps`, and `initMergeProps`, creates a `connectAdvanced`-compatible selector
- `wrapMapToProps.js`: helper functions for wrapping values of `mapStateToProps` and `mapDispatchToProps` in compatible selector factories
 - 
- utils/
 - `Subscription.js`: encapsulates the hierachial subscription concept. used by `connectAdvanced.js` to pass a parent's store Subscription to its children via context
 - `verifyPlainObject.js`: used to show a warning if `mapStateToProps`, `mapDispatchToProps`, or `mergeProps` returns something other than a plain object
 

## file graph
graph
 ```
digraph files {
  "connectAdvanced.js" -> "React"
  "connectAdvanced.js" -> "Subscription.js"

  "connect.js" -> "connectAdvanced.js"
  "connect.js" -> "mapDispatchToProps.js"
  "connect.js" -> "mapStateToProps.js"
  "connect.js" -> "mergeProps.js"
  "connect.js" -> "selectorFactory.js"

  "mapDispatchToProps.js" -> "wrapMapToProps.js"
  "mapStateToProps.js" -> "wrapMapToProps.js"
}
```

The modular structure of all the functions in `connect/` should allow greater reuse for anyone that wants to create their own `connect` variant. For example, one could create a variant that handles when `mapStateToProps` is an object by using reselect's createStructuredSelector:

### customConnect.js:
```javascript
import connect from 'react-redux/lib/connect/connect'
import defaultMapStateToPropsFactories from 'react-redux/lib/connect/mapStateToProps'

function createStatePropsSelectorFromObject(mapStateToProps) {
  const initSelector = () => createStructuredSelector(mapStateToProps)
  initSelector.dependsOnProps = true
  return initSelector
}

function whenStateIsObject(mapStateToProps) {
  return (mapStateToProps && typeof mapStateToProps === 'object')
    ? createStatePropsSelectorFromObject(mapStateToProps)
    : undefined 
  }
}

export default function customConnect(mapStateToProps, mapDispatchToProps, mergeProps, options) {
  return connect(mapStateToProps, mapDispatchToProps, mergeProps, {
    ...options,
    mapStateToPropsFactories: defaultMapStateFactories.concat(whenStateIsObject)
  })
}

```
### ExampleComponent.js
```javascript
// ...
export default customConnect({
  thing: thingSelector,
  thong: thongSelector
})(ExampleComponent)
```

And for scenarios where connect's three-function API is too constrictive, one can directly call, or build a wrapper around, `connectAdvanced` where they have full control over turning `state` + `props` + `dispatch` into a props object.

# Performance

I'm using a [modified version of react-redux-perf](https://github.com/jimbolla/react-redux-perf) to performance test+profile the changes. It's configured to try to fire up to 200 actions per second (but becomes CPU bound), with 301 connected components. There are 2 scenarios being tested:
- NB: a parent component with 300 child components, with no other React components between them.
- WB: the same setup as NB but there's a "Blocker" React component between the parent and children that always returns false for `shouldComponentUpdate`.

I measured the milliseconds needed to render a frame using the [stats.js](https://github.com/mrdoob/stats.js/) used by react-redux-perf:

MS: avg (min-max) | current NB | rewrite NB | current WB | rewrite WB
----------------- | ---------- | ---------- | ---------- | ----------
Chrome | 170 (159-223) | 20 (17-55) | 170 (167-231) | 17 (15-59)
Firefox | 370 (331-567) | 20 (16-51) | 430 (371-606) | 19 (15-60)
IE11 | 270 (127-301) | 40 (36-128) | 300 (129-323) | 33 (30-124)
Edge | 240 (220-371) | 37 (32-102) | 260 (97-318) | 28 (24-100)

On the conservitive end, the rewrite is about 8x faster under these circumstances, with Firefox even doubling that improvement. Much of the perf gains are attributed to avoiding calls to `setState()` after a store update unless a re-render is necessary.

In order to make that work with nested connected components, store subscriptions were changed from "sideways" to top-down; parent components always update before their child components. Connected components detected whether they are nested by looking for an object of type `Subscription` in the React `context` with the key `storeSubscription`. This allows `Subscription` objects build into a composite pattern.

After that I've used Chrome and Firefox's profilers to watch for functions that could be optimized. At this point, the most expensive method is `shallowEqual`, accounting for 4% and 1.5% CPU in Chrome and Firefox, respectively. 

# connectAdvanced(selectorFactory, options) API

In addition to the changed related to performance, the other key change is an additional API for `connectAdvanced()`. `connectAdvanced` is now the base for `connect` but is less opinionated about how to combine `state`, `props`, and `dispatch`. It makes no assumptions about defaults or intermediate memoization of results, and leaves those concerns up to the caller. It does memoize the inbound and outbound props objects. A full signature for `connectAdvanced` with its `selectorFactory` would look like:

```javascript
export default connectAdvanced(
  // selectorFactory receives dispatch and lots of metadata as named parameters. Most won't be
  // useful to a direct caller, but could be useful to a library author wrapping connectAdvanced
  (dispatch, options) => (nextState, nextProps) => ({
    someProp: //...
    anotherProp: //...
  }),
  // options with their defaults:
  {
    getDisplayName: name => `ConnectAdvanced(${name})`,
    methodName: 'connectAdvanced',
    renderCountProp: undefined,
    shouldHandleStateChange: true,
    storeKey: 'store',
    withRef: false,
    // you can also add extra options that will be passed through to your selectorFactory:
    custom1: 'hey',
    fizzbuzz: fizzbuzzFunc
  }
)(MyComponent)

```

A simple usage may look like:
```javascript
export default connectAdvanced(dispatch => {
  const bound = bindActionCreators(actionCreators, dispatch)
  return (store, props) => ({
    thing: state.things[props.thingId],
    saveThing: bound.saveThing
  })
})(MyComponent)

```

An example using `reselect` to create a bound actionCreator with a prop partially bound:
```javascript
export default connectAdvanced(dispatch => {
  return createStructuredSelector({
    thing: (state, props) => state.things[props.thingId],
    saveThing: createSelector(
      (_, props) => props.thingId,
      (thingId) => inputs => dispatch(actionCreators.saveThing(thingId, inputs))
    )
  })
})(MyComponent)
```

An example doing custom memoization with actionCreator with a prop partially bound:
```javascript
export default connectAdvanced(dispatch => {
  return createStructuredSelector({
    let thingId
    let thing
    let result
    const saveThing = inputs => dispatch(actionCreators.saveThing(thingId, inputs))

    return (nextState, nextProps) => {
      const nextThingId = nextProps.thingId
      const nextThing = nextState.things[nextThingId]
      if (nextThingId !== thingId || nextThing !== thing) {
        thingId = nextThingId
        thing = nextThing
        result = { thing, saveThing }
      }
      return result
    }
  })
})(MyComponent)
```

Note these are meant as examples and not necessarily "best practices."

# Pros/cons

I understand there is great risk to accepting such drastic changes, that would have to be justified with significant benefits. I'll reiterate the two main benefits I believe these changes offer:

1. Performance: There's potentially huge perf gains in situations where the number of connected components is high, stemming from conceptual changes to subscriptions so they go with the natural flow of events in React vs across them, as well as method profiling+optimizing using Chrome/FF.
2. Extensibility/Maintainability: By splitting the responibilities of connect into many smaller functions, it should be easier both for react-redux contributors to work with the codebase and end users to extend its functionality though the additional APIs. If users can add their desired features in their own projects, that will reduce the number of feature requests to the core project.

Despite passing all the automated tests as well as week of manual testing, there is risk of impacting users dependent on implicit behavior, or that performance is worse in some unexpected circumstances. To minimize risk of impacting end users and downstream library authors, I think it would be wise to pull these changes into a "next" branch and first release an alpha package. This would give early adopters a chance to test it and provide feedback

# Thanks

I'd like to thank the other github users who have so far offered feedback on these changes in #407, especially @markerikson who has gone above and beyond.


src/components/connect.js
```diff
@@ -1,367 +1,46 @@
...
-}
-
-let errorObject = { value: null }
-function tryCatch(fn, ctx) {
-  try {
-    return fn.apply(ctx)
-  } catch (e) {
-    errorObject.value = e
-    return errorObject
+import { flow } from 'lodash'
```
davidkpiano: Probably preferred to `import flow from 'lodash/flow'`

gnoff:
I find this change very interesting. I would imagine that before it (if) it is merged that your new selectors as well as the connectAdvanced implementation should have tests independent of the unit tests provided by the base connect api. I've already gone through everything once but I honestly find it hard to grasp the interplay of the various selector factories and selector helpers that work to implement the connect api. I'm hoping a second pass and a few more days may make this a little easier on the noggin. either way, this is very impressive.

jimbolla:
@gnoff. I can probably refactor the factories/selectors relationship to be a little clearer. I may have made it too abstract for the sake of hypothetical reuse. Keep an eye out for an update.

As for tests, right now everything is indeed being tested indirectly through the original tests. It would be valuable to write tests directly against some of the lower level parts but I wasn't ready to do that until I get some kind of feedback on the likelihood of these changes being accepted. If the project maintainers say they'll accept the PR if I write some tests, then I'll def write the tests.

gaearon:
Very interesting. I'm not opposed to merging something as radical if you can show that it doesn't break existing apps. It would be worth releasing this as an alpha, gather feedback from a few big Redux users, and then make a decision. 

Before we do this I would like to ask you to remove any new APIs (functions, methods or props). We don't want people to start depending on them in case this approach doesn't work out. We can always add new things later but let's ensure this doesn't have compatibility gaps first.

gaearon:
Can I ask you to manually test Redux examples in the repo as well as some popular apps?
@markerikson might suggest a few.

jimbolla:
@gaearon I removed connectAdvanced from index.js and filtered out any extra options props to connect(). Does that adequately hide the new functionality, or do we have to worry about users importing from `react-redux/lib/something`? I'll test the repo's examples and any recommended projects this weekend.

src/utils/memoizeProps.js
```diff
@@ -0,0 +1,18 @@
+import shallowEqual from '../utils/shallowEqual'
```
wtgtybhertgeghgtwtg: Couldn't this just be `import shallowEqual from './shallowEqual'`?

ellbee:
I've taken a quick look at this and I'll echo what others have said—it is very interesting, especially the performance improvements around the subscription callbacks.

Hopefully I'll have time to try it out some more and do some tests over the weekend, but I did notice that it seems like `utils/wrapActionCreators.js` can be removed as it is not used anymore?

markerikson:
@jimbolla : perhaps some of the repos listed at https://github.com/markerikson/redux-ecosystem-links/blob/master/apps-and-examples.md might be suitable for experimentation?

jimbolla:
@gaearon I've run all the examples in redux/examples against the rewrite and didn't find any problems. Steps were:

1. npm run build:examples
2. npm: run test:examples
3 overwrite node_modules/react-redux/lib with rewrite in each example
4. repeat 1 & 2. all tests still passed
5. run each example manually to see if there was any unexpected behavior. I did not see any

Next I am going to look at @markerikson 's list as well as reach out to the folks on reactiflux chat for repos to test.

jimbolla:
I went down [this list](https://github.com/markerikson/redux-ecosystem-links/blob/master/apps-and-examples.md), cloned the ones that had tests (looked for a test command in package.json), and tried to install and run each. Some failed for various reasons (needed python or mongodb, for examples) but for those that would successfully run their tests, I was able to supplant their react-redux node_module with my own build and successfully re-run tests.

I also did this for redux-form, and it's almost-400 tests still passed. I attempted react-redux-form but it dies inside something react native related before I even got to my version.

src/utils/Subscription.js
```diff
@@ -0,0 +1,56 @@
+
+// enapsulates the subscription logic for connecting a component to the redux store, as well as
```
luqmaan: `encapsulates`

gaearon:
Can you also try some React Native examples just in case?

jimbolla:
@gaearon I haven't been able to get anything React Native running on my machine. Once I got past errors related to installing python, now it's complaining about missing c++ things.

jimbolla:
If no one else is able to test this with react native, next week I'll set up a VM and attempt to get RN in that for testing.

markerikson:
I did a review pass through the new codebase, trying to get a feel for what's going on.  Jim was hanging around in Reactiflux, so I asked him several questions about behavior in the process.  The one vague hypothetical concern I saw would be if somehow a child unsubscription happened during `Subscription.notifyNestedSubs`, causing a callback to be removed shortly before the loop tries to call it.  Should be easily fixable with an object copy, similar to how Redux handles its own notification process.

Code quality looks very good.  Functions are generally short and readable.  The helper functions don't generally have comments, but are clear enough that that's not an issue.  The higher-level logic in `connectAdvanced()` has considerably more comments, especially around the arguments.  Very well-written overall.

I built the branch and swapped out the `react-redux` folder in my own app's `node_modules`, rebuilt my dependencies bundle with Webpack's DllPlugin, and loaded up my app.  Didn't do an exhaustive test or any specific perf benchmarks, but did run through several chunks of functionality.  Everything looked just fine.

My only concern at this point is trying to follow how all of the various levels of selectors fit together.  Even with the diagram in this PR, things are a bit hazy.  The `match` approach used for handling the various possible argument forms for `mapState` and `mapDispatch` also took me a couple minutes to grasp, although the individual case handling functions themselves are very nice and encapsulated.  Not immediately sure what would be needed to make the selector connections and interplay more understandable.

Overall, the PR is looking very solid.  I am generally :+1: on the PR at this point.  @gaearon , any thoughts?  Anyone else have a chance to play around with this?

jimbolla:
Thanks for taking a look.

I added a guard to the notification loop in case there's a case where a component does something that causes its siblings to unsubscribe during the loop. I'm not sure how it'd be possible unless the component was coded to have side effects in render() such as router.push() or something, but since it's only one extra line of code to check, it's worth doing.

I'm going to attempt a comment in connect.js to explain the interplay of the selectors. It's not super obvious because there's some intentional indirection to allow for easier injection of new functionality.

markerikson:
That new big comment definitely helps, thanks.  Could we maybe get a couple comments clarifying how the `match` function works?  I see what it's doing now, but it wasn't immediately obvious at first.  Also, any particular reason why match is iterating backwards?  Could both the iteration and the export ordering of the selectors be flipped?

Sinistralis:
Talked with Mark on Reactiflux today after running across this and pulled it into our company project. All our tests pass and I've run through multiple workflows with the app and everything works fine. 

Just to note, our application runs with the following packages. I'll only list the ones that are relevant here:

`
    "react": "15.1.0",
    "react-redux": "4.4.5",
    "redux": "3.5.2",
    "redux-saga": "0.10.5",
    "reselect": "2.5.1",
`

We also make use of a few factory functions in some of our connect calls as we are controlling how often updates are sent to the connected component.


The chrome extension dev tool also handles this change completely fine.


Unfortunately I cannot contribute to the React-Native cases yet since we havn't gotten a chance to dig into it yet.

ellbee:
I also ran this with our company project with no observable problems and all tests passing.

I'm using ` "react": "15.2.0", "react-redux": "4.4.5", "redux": "3.5.2", "reselect": "2.5.1"`.

markerikson:
@ellbee : just to confirm, you did use Jim's branch of `react-redux`?

ellbee:
@markerikson Yep, used the branch. `"react-redux": "4.4.5"` is the version that we are running in production. Sorry if it wasn't clear!

jimbolla:
@markerikson I'll add comments to match... will explain why it goes backwards.

@Sinistralis I describe how I did perf testing under the Performance header at the top of this page. Basically just recording before/after values for stats.js. Also, just using FF/Chrome's profiler to see methods with longest times.

gaearon:
I’m generally 👍 on this as long as we release it as a beta first. I’m not following the thread very carefully so if you feel it’s ready please ping me on Twitter to remind me to cut the release. Thanks for working on this!

markerikson:
I'd say it's probably about ready for a beta.  It's passing the API tests, it's been tried with a bunch of applications with no issues, and the code looks pretty reasonable at this point.  Anyone else have thoughts?

markerikson:
@jimbolla , @reactjs/redux : if no one else has any comments in the next few hours, I'll do as Dan asked and ping him to set up a beta.

gnoff:
I think a beta is a great idea.

timdorr:
BTW, because the diff isn't super-helpful, just browse from here: https://github.com/jimbolla/react-redux/tree/connect-rewrite/src/components

Maybe put this on a dist-tag `react-redux@rewrite` or something not in the semver vernacular, just in case this doesn't work out for some reason. I'm confident it will be fine, but I would hold off making plans that this is the New Way™ until after we've got some feedback from the broader community.

jimbolla:
I think tagging it with `@rewrite` or similar is a good idea. I don't intend to make any other changes right now besides add more comments unless anyone has any feedback on things that could be changed.

gnoff:
@timdorr @jimbolla @markerikson depends on how many people we want using this in the wild and how soon. If we tag it as `rewrite` then we reduce the number of people who are likely to come across it and exercise it. We would still need a `alpha` or `beta` phase then to get critical mass to move to general release.

if we use more traditional prerelease vernacular then we are likely to pick up a much larger proportion of experimenters and early adopters.

since the public API does not change we can silently bail out of the rewrite with no community fallout. I'm more concerned about when we expose `connectAdvanced` api as public. This is where the extra very obvious opt-in to non-standard release makes the most sense to me.

That said, being conservative isn't generally a bad thing, I can see the argument to protect this more deliberately from accidental opt-in.

timdorr:
There are only two hard things in computer science: cache invalidation and naming things. (There are only two hard problems in distributed systems: 2. Exactly-once delivery 1. Guaranteed order of messages 2. Exactly-once delivery)

Let's just call it `react-redux@next` then. No implication of the contents being good or bad, just something we might be trying next. I think most people are used to `*@next` dist-tags anyways.

gnoff:
@next is good :)

markerikson:
Cool.  I'll go ahead and ping Dan, then.

src/components/connectAdvanced.js
```diff
@@ -0,0 +1,222 @@
...
+
+    // the func used to compute this HOC's displayName from the wrapped component's displayName.
+    // probably overridden by wrapper functions such as connect()
+    getDisplayName = name => `ConnectAdvanced(${name})`,
+
+    // shown in error messages
+    // probably overridden by wrapper functions such as connect()
+    methodName = 'connectAdvanced',
+
+    // if true, shouldComponentUpdate will only be true of the selector recomputes for nextProps.
```
andydavies92: > `...will only be true of the selector recomputes...`

Should that be `if` rather than `of`?

gaearon:
Hmm. I just tried this, and it doesn’t seem as fast at least in the particular scenario I’m testing.
Here’s what I tried:

1. Clone https://github.com/mweststrate/redux-todomvc
2. `npm run build`
3. Open Chrome Timeline and press Record
4. Toggle the first todo four times
5. Calculate the average scripting time by click

screen shot 2016-07-08 at 16 23 25

I consistently get about 45ms average with the current version, and about 60ms after this rewrite.
Would you like to look into this?

I understand it’s an artificial benchmark but still.

jimbolla:
Ah. It took me a while to figure it out, but it's related to `TodoItem`'s `makeMapStateToProps`:

```
const makeMapStateToProps = (initialState, initialProps) => {
  const { id } = initialProps
    const { todos } = state
    const todo = todos.byId[id]
    const relatedTodo = todo.relatedId && todos.byId[todo.relatedId]
    return {
      todo,
      relatedTodo
    }
  }
  return mapStateToProps
}
```

It grabs the only prop it needs on first call, so its inner function doesn't depend on props. This allows the current version to skip out on props changes. If I change that `mapStateToProps` to grab `id` as a second parameter, the current version doubles in time.

My version currently always does ownProps memoization even when it's not needed, like in this case. I need to change it so that it knows whether it needs to. Working on that now; hopefully I'll have that updated tonight.

jimbolla:
Still working on this but I just pushed some changes for refactorings/optimizations. Will get back to it tonight after some family time.

src/components/connectAdvanced.js
```diff
@@ -0,0 +1,234 @@
...
+      export default connectAdvanced((dispatch, options) => (state, props) => ({
+        thing: state.things[props.thingId],
+        saveThing: fields => dispatch(actionCreators.saveThing(props.thingId, fields)),
+      }))(YourComponent)
+
+    Access to dispatch is provided to the factory so selectorFactories can bind actionCreators
+    outside of their selector as an optimization. Options passed to connectAdvanced are passed to
+    the selectorFactory, along with displayName and WrappedComponent, as the second argument. 
+
+    Note that selectorFactory is responisible for all caching/memoization of inbound and outbound
```
timdorr: Tiny typo here: "responsible"

gaearon:
Ping me again when it’s ready? Thanks!

jimbolla:
I'm just about done with my refactorings/optimizations and will recap soon, but I also wanted to go through the open issues/PRs to investigate if/how they relate to this PR:

* #269 I'll have to look closer at the one to see, because it's fairly involved, but I think it's at least partially related to some extra calls to `setState` in the current impl that are avoided my mine. Not 100% sure.
* #292/#293 This would be an alternate solution to #292. Instead of relying on batchedUpdates, it forces the notification order of updates to be top down. I added its passing test to this PR.
* #322 Unsure/unconfirmed. There's a test that covers actions dispatched in `componentWillMount`, so I'm not sure why it wouldn't detect those that happen in `componentDidMount`. 
* #323 Can be added to `mapStateToProps.js` as another factory, or can be added in userland with a wrapper around `connect` that passes the implementation of that factory in the `mapStateToPropsFactories` options arg. I can help with the code migration if we want to move forward with both this PR and #323.
* #345 Done.
* #368/#373 This would be an alternate solution to #368, similar to my comments on #292/#293
* #395 Done, but made it a function. Developer can override getDisplayName (defaults to `name => ``Connect(${name})`` `). Added its passing test to this PR.
* #398 Pretty sure this is covered, but needs a test to confirm.
* #428/#429 Done. Added passing test from this PR.
* #436 Done. Added its code to connectAdvanced and added its passing test to this PR.

jimbolla:
I went through a round of refactorings/optimizations based on the results @gaearon found in redux-todomvc.  The scenario is that we have a list of 10,000 todo items components, each of which is connected. The speed being measured is when a todo is marked completed. To measure the results consistently, I set up a timer to check the first 10 items in the list like this:
```javascript
for (let i = 0; i < 10; i++) {
  const id = 'prefilled-' + i
  const delay = 15000 + (i * 100)
  setTimeout(() => store.dispatch(completeTodo(id)), delay)
}
```
This gave me enough time to let the page load and start Chrome's profiler. (Also profiled in Firefox but didn't bother writing down timings, but the results were consistent.) The `connect()` call of importance is the one for TodoItem. Note that this connect uses the factory style to grab the `id` from `initialProps` so that its selector doesn't have to depend on props :

### version 1
```javascript
const makeMapStateToProps = (initialState, initialProps) => {
  const { id } = initialProps
    const { todos } = state
    const todo = todos.byId[id]
    const relatedTodo = todo.relatedId && todos.byId[todo.relatedId]
    return {
      todo,
      relatedTodo
    }
  }
  return mapStateToProps
}

export default connect(
  makeMapStateToProps,
  { completeTodo, editTodo, deleteTodo }
)(TodoItem)
```
The second test I profiled was to change the `mapStateToProps` to the following. Note that this version grabs `id` from props in the final selector:

### version 2
```javascript
const makeMapStateToProps = (initialState, initialProps) => {
  const mapStateToProps = (state, { id }) => {
    const { todos } = state
    const todo = todos.byId[id]
    const relatedTodo = todo.relatedId && todos.byId[todo.relatedId]
    return {
      todo,
      relatedTodo
    }
  }
  return mapStateToProps
}
```
For the third test, I removed the extra factory wrapper that's no longer needed from version 2:
### version 3
```javascript
  const mapStateToProps = (state, { id }) => {
    const { todos } = state
    const todo = todos.byId[id]
    const relatedTodo = todo.relatedId && todos.byId[todo.relatedId]
    return {
      todo,
      relatedTodo
    }
  }
```
Finally I wrote a version that used `connectAdvanced` directly, which handles its own value memoization:

### connectAdvanced
```javascript
const selectorFactory = (dispatch) => {
  const bound = bindActionCreators({ completeTodo, editTodo, deleteTodo }, dispatch)
  let todo = undefined
  let relatedTodo = undefined
  let result = undefined

  return ({ todos }, { id }) => {
    const nextTodo = todos.byId[id]
    const nextRelatedTodo = nextTodo.relatedId && todos.byId[nextTodo.relatedId]
    
    if (todo !== nextTodo || relatedTodo !== nextRelatedTodo) {
      todo = nextTodo
      relatedTodo = nextRelatedTodo
      result = {
        completeTodo: bound.completeTodo,
        editTodo: bound.editTodo,
        deleteTodo: bound.deleteTodo,
        todo,
        relatedTodo
      }
    }
    return result
  }
}
export default connectAdvanced(selectorFactory)(TodoItem)
```

### results
Aveage time in ms per `dispatch(completeTodo(...))` for 10 samples:

test | current | rewrite
---- | -------- | --------
version 1 | 28.6 | 36.9
version 2 | 223.2 | 35.8
version 3 | 215.4 | 32.4
advanced | N/A | 24.3

So where I ended up is that for the original (version 1) `makeMapStateToProps`, the rewrite performs slightly slower, but performs significantly faster for a more "typical" selector (version 3). I think this is related to the added code to solve #292, #368, and #398. But the `connectAdvanced` version can eek out a little bit more speed (mostly by avoiding some calls to `shallowEqual`) by handling its own memoization.

jimbolla:
At this point, I think I'm done optimizing+refactoring, pending any feedback. I can go update the original PR description to align to the latest code changes, if anyone would find that helpful. Beyond that, I'm ready for this to be tested/reviewed by more people.

A couple other notes:
* `Subscription` is not dependent on React in any way. It might be useful to move this into redux in time. Its code is based on the code for `createStore`.
* Most of the tests for `connect` can probably be split to either test `connectAdvanced` or connect's functions directly, now that connect simply joins those together. The code for connect() currently looks like:
```javascript
export default function connect(...args) {
  const options = buildConnectOptions(...args)
  return connectAdvanced(options.selectorFactory, options)
}
```

PCreations:
It's with no doubt an awesome work, but, with your last tests in  mind are you saying that the whole point of this rewrite is to gain about 10ms (version 1 without rewrite VS version advanced) when rendering a very large set of connected component ? No offence here, juste to be sure i'm not missing something super obvious (other than better code / responsabilities splitting).

timdorr:
@PCreations It's roughly an order of magnitude faster for many common types of selectors. Not a huge deal for a single component, but if you have a deep tree and lots of connected components, it will start to add up. Especially for those on a hot path.

But probably more important here is the code cleanup. Things are more readable and organized now. Cruft has built up over time on the current release, so this is a good chance to refactor and level up the code style a bit.

PCreations:
Makes perfect sense, thanks for the clarification !

stevewillard:
Excited to try this out! Is this ready as a beta release candidate, or does it need more testing?

timdorr:
@stevewillard You may be able to install directly from @jimbolla's branch:

```
npm install --save jimbolla/react-redux#connect-rewrite
```
It *should* work, but you might need to mess with your build setup to get the code to be transpiled. Alternatively, you can check it out locally, run the build steps, and install against that local copy.

jimbolla:
@stevewillard I pinged Dan about it and he said he'd take a look this week. I'm not sure if that means release a beta or do more of his own testing first. If you want to test it yourself in the meantime, what I do is build the branch (npm install && npm run build) and then copy the lib & dist folders into my node_modules/react-redux folder.

jimbolla:
@gaearon Is there anything else you'd like me to do/test with this?

timdorr:
Unfortunately, only Dan has the publish bit set on npm, so only he can push out the planned `react-redux@next` dist-tagged testing release.

@gaearon Can you package up a release for testing? Everything here looks good to me. If not, I can take care of it if you are OK with adding me to npm.

jimbolla:
In the interim, I published `@jimbolla/react-redux` to npm for anyone that's eager to experiment with this. If you're using webpack, you can add an alias to substitute this package for the official one w/o having to change any other individual files. Just add the following code to your webpack.config.js:
```js
module.exports = {
  resolve: {
    alias: {
      'react-redux': '@jimbolla/react-redux'
    }
  }
}
```

gaearon:
@timdorr Sorry, I’ve been unexpectedly swamped by Create React App which started as a hackathon project 😄 . I’ll give you the npm ownership so you could publish it.

gaearon:
@timdorr You should have access to `redux`, `react-redux`, `redux-thunk`.

max-mykhailenko:
@jimbolla When do you plan to release it?

timdorr:
This version of react-redux is now available on npm for you to install and test:
```sh
npm install --save react-redux@next
```
It's tentatively called 5.0.0, although technically it has no breaking API changes. semver schmemver!

Please report any bugs you find with it as a separate issue on the repo. 

As some housekeeping, I'm going to close out this PR and instead merge it into the `next` branch on the repo (since this is against `master`). @jimbolla, you (or anyone else for that matter) can open up any followup PRs with further improvements and bug fixes against that branch. I'll go through and close out any issues that this PR solves (so many!) and we can plow forward on getting this released. I'll give it a week or so and we'll promote to beta status so we can get even more people trying it out. 

Everyone please give @jimbolla a huge round of virtual applause for this PR. He's done some really great work here!

zavan:
👏👏👏🎉🎉

jimbolla:
I'm excited to see how this goes. :D

markerikson:
And thank _you_, @jimbolla , for your initiative and all the time and effort you've put into working on this, as well as everyone else who has participated in the discussion.  This has been a very nice example of open source collaboration at work.

OshotOkill:
Oh my gwd @jimbolla love you man for all the hard work you've done!!!

slorber:
thanks @jimbolla . Didn't have much time to see what you have done so far but it looks great and will probably solve some of my own issues

irvinebroque:
Not sure how you're collecting feedback on the 5.0 alpha release, but we've tried it with two of our apps and it's working extremely well with zero code changes. 👏 👏 👏 - really nicely done, and the writeup in this PR helped me understand a lot about what react-redux does under the hood.

timdorr:
@irvinebroque Open a new issue here if you find anything. We'll want a reproducible test case (a "repro") to most effectively fix it, but at least notifying us would be a great start!

markerikson:
@irvinebroque : Dan wrote up a gist a while back that shows what `connect()` does at a conceptual level: https://gist.github.com/gaearon/1d19088790e70ac32ea636c025ba424e .  Both the original version of `connect` and Jim's rewrite are way more complicated than that internally, but that gist demonstrates the overall intent: subscribe to the store, map state, merge props, render the wrapped component.

psulek:
@timdorr upgrading to `5.0.0-beta.1` but i cant see exported `connectAdvanced` in lib. I see file `src\components\connectAdvanced.js` but it is not exported in index so i cant use it directly by my own. Can you add that export and publish updated beta version?

jimbolla:
@psulek You can still use it directly if you do `import connectAdvanced from 'react-redux/lib/components/connectAdvanced'` in the meantime. If you're trying out `connectAdvanced` would you also review and provide feedback on the related docs in PR #480?

cpsubrian:
Dumped the beta into a fairly complex single-page-app I'm working on and aside from peer-dep warnings it worked like a charm. Don't have any metrics to report but at least I can confirm that the backwards-compat appears to be working for my use-case.

chrisvasz:
Love this! I just dropped the beta into an existing application that uses Redux to store form information. In my application, each field is connected to the store, similar to `redux-form`'s approach after the v6 upgrade. I loaded a fairly complex form and noticed with `Perf.printWasted()` that each field was performing a "wasted" render each time any other field changed. This quickly adds up to hundreds of wasted renders on a medium-sized form. (Imagine that each character in an `<input>` causes a state change in the store. For an average typist, that is 3-4 characters per second. Multiply that by the number of form fields and you arrive at hundreds of wasted renders _per second_ on a form with just 25 fields, which is a small number for our use case.)

With the new `react-redux`, against the exact same code, the hundreds of wasted renders have dropped to zero! Love it! Way to go @jimbolla!

Sinistralis:
Dropped this into the same application I tested earlier. This test was performed on mock data, so the data is slightly different but the structure, relationship, etc of the data are all the same across both. 

The test was a drag and drop implementation using lists and smart components. The difference with this change is MASSIVE.

capture
capture2

slorber:
Hi,

This is good news, but can someone explain the performance improvements? I'm not sure to understand why it performs better than before as a drop-in replacement

markerikson:
@jimbolla can explain further, but it's a combination of several factors.  Primarily, there's a ton of custom memoization going on using selectors, which avoids the need to call `setState`.  v4, on the other hand, called `setState` and then did the real checking in the wrapper's `render` function.  It also ensures that subscriptions are handled top-down, fixing a loophole where children could get notified by the store before ancestors and causing unnecessary render checks.  See the rest of this thread, as well as #407 for discussion and details.

jimbolla:
@slorber @markerikson The short short version... Forcing the order of components' store subscriptions to trigger top-down allows connect to avoid extra calls to setState, render, and mapStateToProps'n'friends.

Sinistralis:
Doesn't that mean it also fixes the issue where can you end up getting components in in-determinant states because they render before their parent, which could result in a child trying to access non-existent data because they haven't been unmounted yet?

I've actually run into this problem, and tested to see if the issue still occurs with v5 and it does not, so another issue fixed?

markerikson:
Y'know, that's a good point.  I've seen that in my own prototype app, where I frequently use connected parents rendering connected list items.  I bet there's a good chance v5 resolves that.

jimbolla:
Yep. That was actually the motivation for making the change. The perf was a nice surprise.

slorber:
That's nice! so finally it somehow breaks compatibility (not in a way that's likely to break someone's app) and the new behavior makes connect more performant, in addition to being more flexible.

However @jimbolla can you take a look at this usecase of a list of 100k connected elements? http://stackoverflow.com/a/34788570/82609
I've seen your comments pointing out that the new implementation will permit to solve this problem in a concrete way, but I'm not sure how it can be done however. Any code to share?

jimbolla:
@slorber A few thoughts on that:

* Can squeeze some extra perf by using `connectAdvanced` with a custom selector instead of `connect`. It would probably be worth it at 100k connected components.
* Another option besides those you suggested would be grouping your components into pages of (let's say) 100, so instead of 100k connected components, you'd have 1000 connected pages containing 100 unconnected components. Not sure if this would be faster or not. But it's worth experimenting
* In v5, connect/connectAdvanced take an option `shouldHandleStateChanges`. (connect sets it based on whether mapStateToProps is provided, connectAdvanced defaults it to true) You could connect a component that reads from state, but doesn't subscribe to store changes. This means it would only trigger rerenders when it receives new props from above. This may be be faster or slower.

jide:
Hi !

Since I updated a project to react-redux 5.0.0-beta.2, I see this error popping up, any clue ?
capture d ecran 2016-10-05 a 15 43 28
Which comes from : 
```
      if (typeof ownProps[key] !== 'undefined') {
        (0, _warning2.default)(false, 'Duplicate key ' + key + ' sent from both parent and state');
        break;
      }
```

timdorr:
@jide Looks like we're calling `warning()` incorrectly in that commit. I'll fix and push out another beta.

jide:
Ok thanx !

timdorr:
OK, pushed beta.3 to remove that check for now.

src/utils/shallowEqual.js
```diff
@@ -1,23 +1,19 @@
...
 
-  // Test for A's keys different from B.
-  const hasOwn = Object.prototype.hasOwnProperty
-  for (let i = 0; i < keysA.length; i++) {
-    if (!hasOwn.call(objB, keysA[i]) ||
-        objA[keysA[i]] !== objB[keysA[i]]) {
-      return false
-    }
+  for (let key in b) {
+    if (hasOwn.call(b, key)) countB++
```
jcready: This could exit before completely iterating through all the `b` keys by checking if `countB` is greater than `countA` in each iteration:

```js
for (let key in b) {
  if (hasOwn.call(b, key) && countA <= countB++) return false
}

return true
```

timdorr: A PR to change that out would be appreciated!

natevw:
Looks like this has been released. Updating to 5.0.4 fixed some seemingly-bizarre bugs (caused by a previously-incorrect assumption that props were reliably passed from the top down), and haven't noticed any regressions. Thanks for making this happen!

PR #417, opened 2016-06-24
Add note about passing `ownProps` functions

erykpiast:


markerikson:
Seems reasonably useful.  Thanks!

docs/api.md
```diff
@@ -72,6 +72,26 @@ Instead, it *returns* a new, connected component class, for you to use.
   * [`pure = true`] *(Boolean)*: If true, implements `shouldComponentUpdate` and shallowly compares the result of `mergeProps`, preventing unnecessary updates, assuming that the component is a “pure” component and does not rely on any input or state other than its props and the selected Redux store’s state. *Defaults to `true`.*
   * [`withRef = false`] *(Boolean)*: If true, stores a ref to the wrapped component instance and makes it available via `getWrappedInstance()` method. *Defaults to `false`.*
 
+> Note: `ownProps` is passed to `mapStateToProps` and `mapDispatchToProps` only if formal definition of the function contains two mandatory parameters (function length has to be greater or equal 2). For example, functions defined like below won't receive `ownProps` as the second argument.
+```javascript
+function mapStateToProps() {
+  console.log(arguments[0]); // state
+  console.log(arguments[1]); // undefined
```
gaearon: I believe this is actually incorrect. We have a test that verifies functions with zero args get both arguments. (Because it probably means function was generated) We only really bail out if there is **exactly one** argument.

gaearon:
Would you mind please making another PR that describes actual behavior? I'd appreciate if you could test the examples first to make sure they work as described.

erykpiast:
Sure, I didn't know that o_O

markerikson:
Whoops, sorry - should have double-checked the behavior myself before merging.

docs/api.md
jimbolla: Can confirm what @gaearon is correct. The test is "depends on props = map length !== 1"

PR #421, opened 2016-06-28
adding rc1 as a peer dependency

ericraio:


markerikson:
I don't _think_ that's necessary.    Are you seeing a specific issue of some kind with the current version spec?

gaearon:
Unfortunately for non-0-minor versions it would be required. But we won't do this. 15.2.0 is expected to ship really soon so it is just not worth it.

PR #425, opened 2016-07-04
Update updateStatePropsIfNeeded()

arusakov:
If component not pure (options {pure: false}), I think, that we don't need shallow compare states in updateStatePropsIfNeeded() and need always return true.


timdorr:
Superceded by #429, which has an included test.

PR #426, opened 2016-07-04
Clarify mapStateToProps re-evaluation when ownProps are used

nico-schneider:
As a result of http://stackoverflow.com/q/38189783/6548463, the docs now state a little more clearly that the component will additionally re-evaluate `mapStateToProps` when props from the parent have shallowly changed.


timdorr:
Thanks!

PR #430, opened 2016-07-08
[skip ci] Tiny document fix: `args` is not used

rhysd:
I found a typo in variadic arguments example for `mapStateToProps`.  `args` is not used.


markerikson:
Yeah, that seems reasonable.  Thanks!

PR #436, opened 2016-07-15
Throw if no component is passed to connect

Aweary:
Currently, if you fail to pass a component to the function returned by `connect` it fails in a confusing way (usually `inst.render is not a function`) that can be hard to track down if you're new to Redux.

This PR adds a check, making sure a function is passed to `connect`, throwing a useful error if not.

### DIFF
--- a/src/components/connect.js
+++ b/src/components/connect.js
@@ -53,6 +53,11 @@ export default function connect(mapStateToProps, mapDispatchToProps, mergeProps,
   const version = nextVersion++
 
   return function wrapWithConnect(WrappedComponent) {
+    invariant(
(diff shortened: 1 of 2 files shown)


src/components/connect.js
```diff
@@ -53,6 +53,11 @@ export default function connect(mapStateToProps, mapDispatchToProps, mergeProps,
   const version = nextVersion++
 
   return function wrapWithConnect(WrappedComponent) {
+    invariant(
+      typeof WrappedComponent == 'function',
```
iamakulov: Nit: `===`?

timdorr:
Superseded by #416

PR #437, opened 2016-07-15
Proposal for standardising getWrappedInstance methods on wrappers

horyd:
I've come into this issue once or twice now, primarily as a result of the amount of component wrappers a given React component can accumulate. Right now I have some with ```connect``` (react-redux), ```injectIntl``` (react-intl) and ```withStyles``` (isomorphic-style-loader) all at once.

So if you have a component that is wrapped three times, and you want to get to the last wrapped instance, you need to do ```this.refs.component.getWrappedInstance().getWrappedInstance().getWrappedInstance()```...

It would be nice for there to be a standard for how you might access wrapped instances where there are multiple levels at play. This concept proposes something like ```this.refs.yourComponent.getWrappedInstance(3)``` to do that. I also like the idea of having a means to go all the way down to the last component without necessarily having to specify the amount of layers to unwrap. Just an idea.

I'm posting this here as I think this library has the popularity to set a good standard with how this might be done, which others could follow.

Let me know thoughts :)

### DIFF
--- a/src/components/connect.js
+++ b/src/components/connect.js
@@ -267,11 +267,15 @@ export default function connect(mapStateToProps, mapDispatchToProps, mergeProps,
           `To access the wrapped instance, you need to specify ` +
           `{ withRef: true } as the fourth argument of the connect() call.`
         )
-        let wrappedInstance = this.refs.wrappedInstance;
-				if (unwraps > 1) {
-          wrappedInstance = wrappedInstance.getWrappedInstance(unwraps - 1);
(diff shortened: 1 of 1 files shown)


gaearon:
Hi, thanks for PR! I don’t think we’d want to enforce a convention like this here.

Really, the core issue is that React doesn’t allow forwarding refs.
There is an issue for this, so you can track it here: https://github.com/facebook/react/issues/4213.

horyd:
Ah I see, hadn't come across that before but good to see an ongoing conversation on the topic :) Thanks!

PR #447, opened 2016-07-29
Move loose-envify to devDependencies

ratson:


gaearon:
We did this before, and it broke people. Unfortunately, unless I’m missing something, this is how Browserify wants it to be. Thanks for the PR though!

Issue #457, opened 2016-08-15
After certain time leads to Redux warning "setState() on an unmounted component"


Restuta:
I am testing v5 alpha from #416 and noticed that after some time every re-render in one of my components starts triggering following warning: 

`Can only update a mounted or mounting component. This usually means you called setState() on an unmounted component. This is a no-op. Please check the code `

My best guess is that there is a memory leak somewhere and new connect holds a reference to unmounted component. Triple checked that it doesn't happen with `4.4.5`. 

It doesn't happen right away, only after I switch back and forth between routes that re-render a lot of components. I don't have time to dig deeper and set up a test project, but I can create a branch in my public repo where it's happening to reproduce. Please let me know if you want this.

max-mykhailenko:
I have the same issue

jimbolla:
A repro would be very helpful.  Thanks.

Restuta:
@max-mykhailenko do you have simple steps to repro?

timdorr:
@Restuta Do you have a reproducible test case as well? 

Does this occur in the canonical `real-world` example? https://github.com/reactjs/redux/tree/master/examples/real-world

Restuta:
@timdorr not it does not, I created a branch that you can use to repro it.

Branch: https://github.com/Restuta/rcn.io/tree/react-redux-connect-bug

Steps to repro:

* `git clone git@github.com:Restuta/rcn.io.git`
* cd `rcn.io`
* `npm i && npm run dev`
* open http://localhost:8888/events/8 in your browser
* open console
* click on MTB in the header and then back to "Event 8"

You should see in console:

"Warning: setState(...): Can only update a mounted or mounting component. This usually means you called setState() on an unmounted component. This is a no-op. Please check the code for the Connect(EventDetails) component."

Notice that the only change from "dev" branch is bump of `redux-react` and you can test it yourself that issue is not happening with `4.4.5`.

jimbolla:
Thanks,  I'll have a look at this tonight.

jimbolla:
@Restuta That git repo address doesn't seem to work for me. Is it supposed to be `https://github.com/Restuta/rcn.io.git`?

Restuta:
@jimbolla that address in SSH format, of you prefer HTTPS then yes, use that one

jimbolla:
[...]

PR #465, opened 2016-08-18
bind proper store context in connectAdvanced and the Subscription util

vhmth:
Addresses https://github.com/reactjs/react-redux/issues/464.

The issue explains this fix pretty well, but basically in the previous version of `connect` the store's context was respected when calling `getState` or `subscribe`. I move that we should have that be respected in this version as well.

Also left you a quick video:

https://www.opentest.co/share/2093c4a0652111e6a1ffdd04f7a207b4

### DIFF
--- a/src/components/connectAdvanced.js
+++ b/src/components/connectAdvanced.js
@@ -159,7 +159,8 @@ export default function connectAdvanced(
       }
 
       initSelector() {
-        const { dispatch, getState } = this.store
+        const { dispatch } = this.store
+        let getState = this.store.getState.bind(this.store)
         const sourceSelector = selectorFactory(dispatch, selectorFactoryOptions)
 
         // wrap the selector in an object that tracks its results between runs
--- a/src/components/connectAdvanced.js
+++ b/src/components/connectAdvanced.js
@@ -106,6 +106,10 @@ export default function connectAdvanced(
           `or explicitly pass "${storeKey}" as a prop to "${displayName}".`
         )
 
+        // make sure `getState` is properly bound in order to avoid breaking
+        // custom store implementations that rely on the store's context
+        this.getState = this.store.getState.bind(this.store);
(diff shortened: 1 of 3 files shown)


jimbolla:
If this is important for some 3rd party libraries, could you also add a test that verifies confirms we're doing the right thing?

src/components/connectAdvanced.js
```diff
@@ -159,7 +159,8 @@ export default function connectAdvanced(
       }
 
       initSelector() {
-        const { dispatch, getState } = this.store
+        const { dispatch } = this.store
+        let getState = this.store.getState.bind(this.store)
```
timdorr: nit: Please use a const here since this var will not be modified.

vhmth:
@jimbolla will do as soon as I get back to my computer.

src/components/connectAdvanced.js
vhmth: 👌🏾 will do.

vhmth:
Added a test which uses a generic `ContextBoundStore` class instance. This class can be used in other places as well if need be going forward.

I also cached a bound version of `getState` within the `Connect` component in `connectAdvanced` so that its use is promoted if used elsewhere down the road.

vhmth:
Out of curiosity, how does the org process of publishing/updating the `next` package work?

When y'all are good with this, I'd like to update our beta extension right away to solve an issue we have with our users, so just trying to get a time frame there.

jimbolla:
Can someone assign this to the 5.0 milestone?

timdorr:
Thanks!

I'll probably publish another release (as beta) once we get the fix for #457 in.

vhmth:
Sounds good @timdorr. Saw that @jimbolla already has a fix he's playing around with. Again - thanks for the work on this next release guys!

PR #467, opened 2016-08-19
Fix for #457: ensure nested subscriptions are being unsubscribed/cleared

jimbolla:
I still want to write a test for this, but creating the PR to get the ball rolling.

### DIFF
--- a/src/utils/Subscription.js
+++ b/src/utils/Subscription.js
@@ -1,6 +1,44 @@
 // encapsulates the subscription logic for connecting a component to the redux store, as
 // well as nesting subscriptions of descendant components, so that we can ensure the
 // ancestor components re-render before descendants
+
+function initListeners() {
+  let count = 0
+  let current = []
+  let next = []
+
+  return {
+    clear() {
+      count = 0
+      next = null
+      current = null
+    },
+
+    notify() {
+      current = next
+      for (let i = 0; i < count; i++) {
+        current[i]()
+      }
+    },
+
+    subscribe(listener) {
+      let isSubscribed = true
+      if (next === current) next = current.slice()
+      next.push(listener)
+      count++
+
+      return function unsubscribe() {
+        if (!isSubscribed || count === 0) return
+        isSubscribed = false
+
+        if (next === current) next = current.slice()
+        next.splice(next.indexOf(listener), 1)
+        count--
+      }
+    }
+  }
+}
+
 export default class Subscription {
   constructor(store, parentSub) {
     this.subscribe = parentSub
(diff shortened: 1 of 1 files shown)


jimbolla:
Code coverage... that's new? I guess now I really have to write that test. :laughing:

timdorr:
Yeah, I added that because we had coverage tooling, but nothing was actually checking it :)

src/utils/Subscription.js
```diff
@@ -1,45 +1,61 @@
...
+    },
+
+    subscribe(listener) {
+      let isSubscribed = true
+      if (next === current) next = current.slice()
+      next.push(listener)
+      count++
+
+      return function unsubscribe() {
+        if (!isSubscribed || count === 0) return
```
vhmth: Out of curiosity what use case does the `count` guard us against? A listener trying to `unsubscribe` after a subscription to the store has already unsubscribed?

vhmth:
I noticed that in [trySubscribe](https://github.com/reactjs/react-redux/pull/467/files#diff-aef7770e642ef4054826f6c0aab908c8R67) there's a line where we pass in `this.onStateChange` to the `subscribe` method. Is this something the owner of the `Subscription` instance sets on it? Trying to find where it's documented.

src/utils/Subscription.js
```diff
@@ -55,7 +71,10 @@ export default class Subscription {
   tryUnsubscribe() {
     if (this.unsubscribe) {
       this.unsubscribe()
+      this.listeners.clear()
```
vhmth: Good catch.

src/utils/Subscription.js
vhmth: Counts always make me weary - especially when there's already an array that internally holds its count anyhow. Can we safely use `current.count` as the source of truth?

src/utils/Subscription.js
```diff
@@ -1,45 +1,61 @@
...
+
+function initListeners() {
+  let count = 0
+  let current = []
+  let next = []
+
+  return {
+    clear() {
+      count = 0
+      next = null
```
vhmth: Maybe we should clear `next` and `current` to be empty arrays. That way we can always check either's `length` property instead of keeping `count`.

src/utils/Subscription.js
jimbolla: holding the count is sliiiiightly faster, especially since the state-changed notification loop is the most trafficked block of code.

src/utils/Subscription.js
jimbolla: See prev comment about count. Setting the arrays to null means if somehow code gets back in here and tries to operator on the null arrays, it will throw an error. `subscribe` should never get called again once it's been unsubscribed, so creating empty arrays would just be extra object allocation.

src/utils/Subscription.js
vhmth: Ah - got it!

timdorr: I would get rid of the counter. If you want to cache length, just do that in your for loop: `for (let I = 0, len = current.length; i < len; i++)`

PR #472, opened 2016-08-24
tests and fixes #457

### LINKED ISSUE #457: After certain time leads to Redux warning "setState() on an unmounted component"

jimbolla:


timdorr:
Why did you close the other PR?

src/utils/Subscription.js
```diff
@@ -1,45 +1,61 @@
...
+  return {
+    clear() {
+      count = 0
+      next = null
+      current = null
+    },
+
+    notify() {
+      current = next
+      for (let i = 0; i < count; i++) {
```
timdorr: Please remove this `count` variable. It risks things getting out of sync somehow. Calling `.length` is just as fast, especially in engines like v8 that use hidden classes.

jimbolla:
github did it automatically when i deleted my branch and recreated it. I wanted to have the failing test *before* the fix so I was diddling with my git history. Ended up messing up my branches and just squashed the changes and recreated the branch.

timdorr:
You can just fix that locally and force push next time. A branch is just a pointer to a particular commit hash, and force pushing lets you change the remote branch/pointer to a commit not along the same chain. Github fixed handling that case a year or two ago, so it no longer loses all your comments and stuff like before.

test/components/connect.spec.js
```diff
@@ -910,6 +910,86 @@ describe('React', () => {
...
+            default: throw new Error('Unknown location: ' + location)
+          }
+        }
+
+        render() {
+          return (<App setLocation={this.setLocation}>
+            {this.getChildComponent(this.state.location.pathname)}
+          </App>)
+        }
+      }
```
timdorr: From one of the react-router maintainers, this is a top quality mock router 😄

src/utils/Subscription.js
```diff
@@ -1,45 +1,61 @@
 // encapsulates the subscription logic for connecting a component to the redux store, as
 // well as nesting subscriptions of descendant components, so that we can ensure the
 // ancestor components re-render before descendants
```
timdorr: I know this is a bit out of scope, but perhaps we can mention the whole `current`/`next` setup below and why it's structured that way, now that all the code is in one spot right below this. It's not super-obvious why you would do that just reading through (do we have tests for `Subscription`? That would be another way of documenting it.)

src/utils/Subscription.js
```diff
@@ -25,15 +23,13 @@ function initListeners() {
       let isSubscribed = true
       if (next === current) next = current.slice()
       next.push(listener)
-      count++
 
       return function unsubscribe() {
-        if (!isSubscribed || count === 0) return
+        if (!isSubscribed || !current) return
```
timdorr: Should be `current.length === 0`. Any array, even an empty one, is always truthy (so this will always be false).

jimbolla: Nope. It's guarding for null current, as it gets set inside `clear()`.

src/utils/Subscription.js
jimbolla: The next/current pattern is straight out of the redux source. My hope is to eventually refactor that code into something reusable by the main store or Subscription.js

timdorr: I know @acdlite wanted to do something like that in reactjs/redux#1729, but it's gone inactive (although, @acdlite did just merge in some stuff on the upstream repo just a few hours ago...). Might be good to focus efforts there so we can use the exact same code in both places.

src/utils/Subscription.js
timdorr: Hmm, I wonder if that should reset back to an empty array then? nulls are potentially dangerous when they sneak into the code to actually mean a 3rd (or sometimes 4th) state. It usually means you need another flag to make that state more explicit, which you already have with `isSubscribed`. I've got a whiff of a code smell here...

test/components/connect.spec.js
vhmth: Agreed. Easy to extend as well for the next person who wants to hoist it and have it take a map of route to component.

src/utils/Subscription.js
jimbolla: Yeah. My thought was to wait until these changes have landed, take care of any outstanding issues and let the dust settle before further tinkering.

jimbolla:
@timdorr Do you think this is ready for another alpha release?

timdorr:
Yeah, I'll do that right now.

timdorr:
`react-redux-5.0.0-beta.1` has been published to `react-router@next`!

PR #476, opened 2016-08-28
Organize new connect.options

jimbolla:
As I'm updating docs for v5, I realized it would be better to organize the new options this way.


timdorr:
This PR is against the wrong branch. Can you redo against `next`? Thanks.

jimbolla:
whoops. yeah

PR #477, opened 2016-08-29
groups factories options and move *areEqual defaults into connect

jimbolla:
take 2

### DIFF
--- a/src/connect/connect.js
+++ b/src/connect/connect.js
@@ -1,4 +1,5 @@
 import connectAdvanced from '../components/connectAdvanced'
+import shallowEqual from '../utils/shallowEqual'
 import defaultMapDispatchToPropsFactories from './mapDispatchToProps'
 import defaultMapStateToPropsFactories from './mapStateToProps'
 import defaultMergePropsFactories from './mergeProps'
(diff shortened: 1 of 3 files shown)


jimbolla:
I'm confused by the code coverage fail here, since it doesn't seem to line up with what I changed.

src/connect/connect.js
```diff
@@ -29,16 +30,24 @@ function match(arg, factories) {
...
-    mapStateToPropsFactories = defaultMapStateToPropsFactories,
-    mapDispatchToPropsFactories = defaultMapDispatchToPropsFactories,
-    mergePropsFactories = defaultMergePropsFactories,
-    selectorFactory = defaultSelectorFactory,
     pure = true,
+    areStatesEqual = strictEqual,
+    areOwnPropsEqual = shallowEqual,
+    areStatePropsEqual = shallowEqual,
+    areMergedPropsEqual = shallowEqual,
+    factories: {
```
timdorr: I don't really agree with this. I'd rather keep the options map flat, even though it's somewhat large. It keeps things relatively symmetrical when dealing with them.

jimbolla: As I'm writing the docs for the new connect features, i'm realizing the 4 factories are different than the other options; they don't make as much sense to use at the individual call sites like the other options. They're intention is to be used in a custom wrapper function like 'customConnect`. That's why I'm trying to separate them from the other options. Maybe they shouldn't be in options at all. Maybe I should use a customization pattern like [redux-thunk](https://github.com/gaearon/redux-thunk/blob/master/src/index.js) or [reselect's createSelector](https://github.com/reactjs/reselect/blob/master/src/index.js#L66). Any thoughts on that?

jimbolla:
Ok. I'm happier with this. Now the "factories" aren't part of the main API; they're only accessible via `createConnect` which isn't exported from `index.js`. I'd like to get some feedback on how other devs might use it before making it official. We can expose it later in a minor release. The upside to this is I can defer writing up docs for it until later.

timdorr:
Putting this one in too.

PR #480, opened 2016-08-31
Update API docs for v5

jimbolla:
This can certainly be better, but it's not a bad start.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1,5 +1,6 @@
 ## API
 
+<a id="provider"></a>
 ### `<Provider store>`
 
 Makes the Redux store available to the `connect()` calls in the component hierarchy below. Normally, you can’t use `connect()` without wrapping the root component in `<Provider>`.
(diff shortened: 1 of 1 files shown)


docs/api.md
```diff
@@ -294,3 +291,80 @@ function mergeProps(stateProps, dispatchProps, ownProps) {
...
+import * as actionCreators from './actionCreators'
+import { bindActionCreators } from 'redux'
+
+function selectorFactory(dispatch) {
+  let state = {}
+  let ownProps = {}
+  let result = {}
+  const actions = bindActionCreators(actionCreators, dispatch)
+  const addTodo = (text) => actions.addTodo(ownProps.userId, text)
+  const memoize = initShallowMemoizer()
```
slorber: Hey, not sure to have seen this memoizer used in this example

jimbolla: Whoops. It was supposed to be part of the example. Must've got lost. I'll update.

timdorr:
OK, let's merge this in for now.

mondaychen:
Hi.

I was reading https://github.com/reactjs/react-redux/blob/master/docs/api.md#examples-1 and this PR contains the latest commit changing that part.

I guess you have a plan to update the example for `connectAdvanced`, but it was not done yet. That's why it now confuses me.

- It seems that `nextProps` should be `nextOwnProps`. Correct? Otherwise it would be an undefined variable.
- What's the use of `state`? It keeps track of the newest state, but is not used to compute anything in the scope.

PR #483, opened 2016-09-05
Enable actions via context for dependency injection

alexlbr:
This PR is to enable actions via context to make writing unit tests for containers easier using dependency injection.
- Explanation of the problem: http://alexlobera.com/react-dependency-injection-write-easier-unit-tests/
- Example using this PR: https://github.com/alexlbr/redix/tree/master/examples/react-redux-context-actions

### DIFF
--- a/src/components/connect.js
+++ b/src/components/connect.js
@@ -146,9 +146,7 @@ export default function connect(mapStateToProps, mapDispatchToProps, mergeProps,
       }
 
       configureFinalMapDispatch(store, props, actions) {
-
         const mappedDispatch = mapDispatch(store.dispatch, props, actions)
-
         const isFactory = typeof mappedDispatch === 'function'
 
         this.finalMapDispatchToProps = isFactory ? mappedDispatch : mapDispatch
--- a/src/components/connect.js
+++ b/src/components/connect.js
@@ -6,6 +6,8 @@ import warning from '../utils/warning'
 import isPlainObject from 'lodash/isPlainObject'
 import hoistStatics from 'hoist-non-react-statics'
 import invariant from 'invariant'
+import { bindActionCreators } from 'redux'
+import { PropTypes } from 'react'
 
 const defaultMapStateToProps = state => ({}) // eslint-disable-line no-unused-vars
 const defaultMapDispatchToProps = dispatch => ({ dispatch })
@@ -81,6 +83,7 @@ export default function connect(mapStateToProps, mapDispatchToProps, mergeProps,
         super(props, context)
         this.version = version
         this.store = props.store || context.store
+        this.actions = context.actions;
 
         invariant(this.store,
           `Could not find "store" in either the context or ` +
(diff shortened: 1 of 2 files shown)


markerikson:
A few initial thoughts:

Your article complains about being unable to call methods on the wrapped instance for testing.  First, one common approach is to test the "plain" component in isolation - verify that it behaves as expected when given props, and don't worry about actually trying to truly connect it.  This lends itself well to mocking, as you can pass in fake callbacks as props.  Also, even when rendering a connected component in a test, you can access the wrapped component if you pass the `{withRef: true}` option to `connect`.

Also, testing the entire connected component is generally less valuable - what you really are doing is verifying that HoCs like `connect` work properly, which is not what your own app should be trying to verify.  You're more likely to get value out of testing the `mapState` functions and the plain components.

Finally, this PR is generally obsolete already, because the React-Redux 5.0 beta release (as written in #416  ) is a complete rewrite of `connect`'s implementation.

alexlbr:
Thanks for your quick reply. Good point about testing connect in my app. I totally agree and it's not what I'm trying to test. What I'd like to test is that my container is connected to the right action. By that I don't mean that connect is not working properly, but some dev could change by mistake my mapDispatchToProps. How would you test that the code I write in the container to connect the actions is correct?

 #416 looks awesome! looking forward to the release.

markerikson:
Admittedly, unit testing is still something I'm pretty weak on.  That said, I've definitely never considered trying to test that the stuff I'm passing to `mapDispatch` is correct.  I'm not sure how you'd do so, and I'm not sure I see the point in it (especially if you're using the object shorthand syntax, and not doing any manual binding in an actual `mapDispatch` function).

jimbolla:
You could write your test such that it wires up your component to a simple store reducer like
```js
function testReducer(state, action) {
  if (action.type === 'THE_EXPECTED_ACTION')
    return { ...state, correctActionFired: true }
  // ...
}
```
And then assert that `store.getState().correctActionFired` is true.

alexlbr:
@markerikson object shorthand syntax is very handy, and I do use it. But we can still introduce errors. Imagine we have a container that the only thing it does is to dispatch an action (to fetch data for instance) when the component is mounted. I'd be very nice if I can double check that the action it dispatches is the one we want.

@jimbolla that's really interesting approach, I'm going to rethink my implementation with that in mind. Though it also needs `connect` to be executed, and it's not ideal as markerikson pointed. But I can't think of any approach to verify the action without connecting the component. 

Another thing that I'm trying to achieve with my PR is to avoid expensive operations that the actions may execute (like fetching data from an API). Using Sinon to generate test doubles of the actions is not easy.

Keep your good feedback and ideas guys, very appreciated!

gaearon:
This doesn't work with code splitting. We don't want to encourage approaches that don't work with code splitting if we can avoid it.

gaearon:
>By that I don't mean that connect is not working properly, but some dev could change by mistake my mapDispatchToProps.

Can you test that function alone in this case?

gaearon:
FWIW I think we should expose both mapStateToProps and mapDispatchToProps as statics on connected component for convenience so they can be tested independently without being explicitly exported.

markerikson:
Dan.  You're supposed to be on vacation.  STOP COMMENTING ON ISSUES AND GO DO VACATION-Y THINGS! :)

(But yeah, exposing `mapState` and `mapDispatch` seems totally reasonable, since we already do the component as well.)

timdorr:
So, I'm going to close this out for the issues listed above. Also, we're full steam ahead on React Redux 5, so it would be best to build things like this against the `next` branch.

PR #490, opened 2016-09-17
added ShouldComponentUpdate component to stop re-rendering

l2silver:
@markerikson Hey Mark, thanks for looking over react-redux-blackbox for me. I took a look at the new v5 react-redux, learned a lot about how to do perf testing, and the thought process that goes into accepting a major rewrite on a plugin.

Anyways, before v5 is official rolled out, I thought I would add these minor changes to improve v4 performance (Things that I picked up on while writing blackbox). Actually, if I'm not mistaken, the only real performance boost that v5 delivers is nullified with these changes. The big one is adding the ShouldComponentUpdate component, which stops rerendering if the computed props and own props are the same as their previous versions.

### DIFF
--- a/src/components/connect.js
+++ b/src/components/connect.js
@@ -31,6 +31,17 @@ function tryCatch(fn, ctx) {
 
 // Helps track hot reloading.
 let nextVersion = 0
+let uniqueId = 0
+
+//Stop component from re-rendering when own props and recalculated state and dispatch props shallow equal previous props
+class ShouldComponentUpdate extends Component {
+  shouldComponentUpdate(nextProps) {
+    return !nextProps.stop
+  }
+  render() {
+    return this.props.renderElement()
+  }
+}
 
 export default function connect(mapStateToProps, mapDispatchToProps, mergeProps, options = {}) {
   const shouldSubscribe = Boolean(mapStateToProps)
@@ -90,6 +101,10 @@ export default function connect(mapStateToProps, mapDispatchToProps, mergeProps,
         )
 
         const storeState = this.store.getState()
+        
+        //We need a unique id to assign to the key of ShouldComponentUpdate
+        this.uniqueId = uniqueId++
+        
         this.state = { storeState }
         this.clearCache()
       }
@@ -234,6 +249,8 @@ export default function connect(mapStateToProps, mapDispatchToProps, mergeProps,
         this.renderedElement = null
         this.finalMapDispatchToProps = null
         this.finalMapStateToProps = null
+        this.uniqueId = null
+        this.wrappedInstance = null
       }
 
       handleChange() {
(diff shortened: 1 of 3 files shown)


markerikson:
Hmm.  Given the path we're on towards v5, I'm not sure there's much point in changing anything further with v4.

Also, how specifically does this improve perf?  Do you have some benchmarks and comparisons you can provide?  How does this compare with v5?

Tagging @jimbolla and @timdorr .

timdorr:
I'm not seeing any particular benefi either. We already make these kinds of optimizations elsewhere in the code. Actually, it looks like you've removed some cases...

jimbolla:
Much of the perf boost in v5 comes from avoiding unnecessary calls to `setState` by re-ordering the subscriptions and doing the props+state diffing logic earlier *before* that happens. That was actually huge. As it stands I think v5 is complete and waiting on feedback on functionality/docs from the community.

timdorr:
Yeah, I've been bogged down with React Router hooplah. I want to get to v5 here, but haters gonna hate on router :P

l2silver:
Fair enough. I just thought I would put it out there in case there's a rollback or something.

@jimbolla it's funny you should say that about needless setState calls because I thought that was the important problem to solve when doing blackbox. But if I'm not mistaken, react made some internal changes to deal with this issue. (it's extremely possible that I'm completely missing something here, so please correct me if I am wrong.)

Here's the scenario
 lets say you have two connect components that will setState on the same store changes, and one component is inside the other. When the store changes, they both setState,  but the parent setState takes priority, and the child setState will only fire if the render propagation fails to reach it. (Because of a false shouldCompomentUpdate in between them)

Is there still a performance gain then?

l2silver:
I tried some performance tests using https://github.com/jimbolla/react-redux-perf

but I got about the exact same results for vanilla react-redux, v5, and my updated react-redux

is there something I am doing wrong? My process was

sudo npm link (in react-redux)
co master
prepublish

react-redux-perf
build
node server

report for each version

jimbolla:
My perf testing approach was to build jimbolla/react-redux-perf (and later mweststrate/redux-todomvc) against react-redux@latest then test. Then I modified the build script in my working copy of react-redux to build into the node_modules/react-redux folder of the perf project, then rebuild.

As for the `setState` scenario, the problem in the current implementation is that because of the `componentDidMount` ordering, the children's subscription always gets notified first, so it's always going to fire `setState` for the children, and then eventually the parents. 

That's why v5 adds the hierarchical subscription code to ensure notifications flow top-down. This means (unless a blocking `shouldComponentUpdate` occurs), the parent will get notified, recompute its props, call `setState`, rerender, triggering connected children to receive new props, recompute props, and rerender, all in one update cycle. By the time the children's store listeners fire, they've already recomputed and rerendered, so they avoid the `setState` altogether.

l2silver:
@jimbolla Ah, got it! Brilliant. Now it makes sense.

PR #500, opened 2016-09-24
WARN on duplicate prop

### LINKED ISSUE #499: Connect doesn't warn if same property is sent from both parent and state
The default merge method in connect:

```
const defaultMergeProps = (stateProps, dispatchProps, parentProps) => ({
  ...parentProps,
  ...stateProps,
  ...dispatchProps
})
```

Will just overwrite existing parent props with state props when same key is used.

IMHO it'd be better to show a warning to console, saving developers debugging time & effort

ynonp:
Fixes #499 - Add a warning to console when same key is used in both parent and state props

### DIFF
--- a/src/components/connect.js
+++ b/src/components/connect.js
@@ -9,11 +9,24 @@ import invariant from 'invariant'
 
 const defaultMapStateToProps = state => ({}) // eslint-disable-line no-unused-vars
 const defaultMapDispatchToProps = dispatch => ({ dispatch })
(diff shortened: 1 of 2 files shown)


timdorr:
So, I would remake this against the `next` branch. I don't think we should land something like this in a patch release, since it might produce spurious output for devs. However, since 5.0's right around the corner, it's a perfect chance to get something like this in.

jimbolla:
The file you would need to modify in the `next` branch is here: https://github.com/reactjs/react-redux/blob/next/src/connect/mergeProps.js#L5

PR #506, opened 2016-09-28
Set displayName explicitly on Provider

roncohen:
When code gets uglified for production builds, original names of identifiers gets lots and React components show up with their uglified name. To remedy this, `displayName` can be set explicitly.

For `React.createClass` style components, displayName gets automatically set by babel.
React components declared using ES6-classes however, have mangled names. This is the case for `Provider`. With this PR, the name is explicitly set which should make the name survive the uglification process.

### DIFF
--- a/src/components/Provider.js
+++ b/src/components/Provider.js
@@ -51,3 +51,4 @@ Provider.propTypes = {
 Provider.childContextTypes = {
   store: storeShape.isRequired
 }
+Provider.displayName = 'Provider'


timdorr:
Can you make this against the next branch? master isn't likely to get another release and it would be good to have in the upcoming 5.0 release.

roncohen:
@timdorr done!

timdorr:
Thanks!

PR #508, opened 2016-10-02
warn on duplicate props

### LINKED ISSUE #499: Connect doesn't warn if same property is sent from both parent and state

ynonp:
Fix #499  - check in dev mode if same key appears in own props and store props and warn


timdorr:
LGTM. Thanks for the refactor!

PR #517, opened 2016-10-09
Fix error from clear() called during notify()

appden:
The `current` variable would be set to `null` when `clear()` was called, which would cause an exception if there was more than one listener.

Since this pattern was modeled off of `createStore()` in `redux`, you can see it does the same thing as this [here](https://github.com/reactjs/redux/blob/master/src/createStore.js#L175-L179).

### DIFF
--- a/src/utils/Subscription.js
+++ b/src/utils/Subscription.js
@@ -17,9 +17,9 @@ function createListenerCollection() {
     },
 
     notify() {
-      current = next
-      for (let i = 0; i < current.length; i++) {
-        current[i]()
+      const listeners = current = next
+      for (let i = 0; i < listeners.length; i++) {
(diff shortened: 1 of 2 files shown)


jimbolla:
Do you have a failing repro so that we can add a test that covers the behavior change?

appden:
@jimbolla I rebased and amended the commit to include a test that would fail with an exception prior to this change. It's a pretty contrived reduction of what happened in my codebase.

jimbolla:
Looks good to me. Second opinion, @markerikson or @timdorr?

timdorr:
Sounds good!

appden:
Would you mind releasing another beta version of so I can use an official version of this module? Thanks!

PR #523, opened 2016-10-17
Update mergeProps to optionally accept a factory style function

neeharv:
I have a very similar use case to https://github.com/reactjs/react-redux/pull/488

I've reused @bsideup's code as-is, with the same test. We've started moving to v5 in production, hoping to see this soon in there!


src/connect/mergeProps.js
```diff
@@ -12,7 +12,11 @@ export function wrapMergePropsFunc(mergeProps) {
     let mergedProps
 
     return function mergePropsProxy(stateProps, dispatchProps, ownProps) {
-      const nextMergedProps = mergeProps(stateProps, dispatchProps, ownProps)
+      let nextMergedProps = mergeProps(stateProps, dispatchProps, ownProps)
+
+      if (typeof nextMergedProps === 'function') {
+        nextMergedProps = nextMergedProps(stateProps, dispatchProps, ownProps)
+      }
```
jimbolla: I don't think this is complete; it looks like it's still going to call the original factory method every time through the loop. I'd expect to see `mergeProps` copied into a variable that is overwritten with its result when the result is a function, similar to [wrapMapToPropsFunc](https://github.com/reactjs/react-redux/blob/next/src/connect/wrapMapToProps.js#L53)

test/components/connect.spec.js
```diff
@@ -480,6 +480,37 @@ describe('React', () => {
...
+      }))
+
+      function sum(a, b) {
+        return a + b
+      }
+
+      @connect(
+        state => state,
+        () => ({ sum }),
+        () => (stateProps, dispatchProps, parentProps) => ({
```
jimbolla: The test should probably cover that the factory part of the mergeProps method is only called once, related to my previous comment.

neeharv: Makes sense. Should this be in a separate test or I'll augment the existing one?

timdorr:
Any update on this, @neeharv?

neeharv:
@timdorr this slipped my mind completely! I'll work on the changes requested over this weekend

neeharv:
@timdorr @jimbolla I've cleaned up the test and memoized the response of the factory. Also updated API docs. Do let me know if there is anything else you'd like me to do. Thanks!

timdorr:
Can you revise the branch on the latest master? There are a lot of junk commits in here.

neeharv:
@timdorr much easier to just do it in a new branch for me. The new PR is here is #626. Closing this so you can take a look there! Thanks.

PR #524, opened 2016-10-18
Replace lodash with lodash.isplainobject

JulianMayorga:


gaearon:
Hi, thanks for the PR! I don't see the benefit of this change, so I'm inclined to close it.
In case there is any confusion, these two forms are completely equivalent in bundle size.
We are importing a specific module from Lodash rather than the whole Lodash.

JulianMayorga:
Hi, thank you for responding to my wordless PR! Sorry for not providing any kind of context in my PR, it was late at night and I was trying (and failing) to deploy a universal react-redux app.

I tried to deploy an app using zeit.co/now, but my build step was failing with an error that said `Module not found: Error: Can't resolve 'lodash/isPlainObject' in '/home/nowuser/src/node_modules/e3d8a6548416ea90c89022179a97a8ef8f8428d0/lib/components'`. This error was coming from react-redux, and I hacked a solution changing lodash with lodash.isplainobject. I really don't know why the latter works, because this error only happens on the remote deployment, everything works great in my machine.

I should have directly opened an issue in zeit/now instead of a PR here. Sorry for the inconvenience!

gaearon:
Maybe there's some issue that caused a wrong version of Lodash to be used in your setup?

JulianMayorga:
That may be the case, the problem is that the only things I know about this error are [these logs](https://full-stack-blog-app-ilsipdurew.now.sh/). Thanks anyways!

PR #537, opened 2016-11-04
Include minified version of redux within production builds

akaila:
- Even though redux is a peer dependency and is externalized two versions of redux can creep in prod builds when main project uses minified redux
- This change changes the minified react-redux for production builds reference minified version of redux while externalizing it

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -10,11 +10,14 @@ var reactExternal = {
   amd: 'react'
 }
 
+var isProduction = (env === 'production');
+var reduxInclude = isProduction ? 'redux/dist/redux.min.js' : 'redux';
+
 var reduxExternal = {
   root: 'Redux',
-  commonjs2: 'redux',
-  commonjs: 'redux',
-  amd: 'redux'
+  commonjs2: reduxInclude,
+  commonjs: reduxInclude,
+  amd: reduxInclude
 }
 
 var config = {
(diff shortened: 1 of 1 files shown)


timdorr:
This isn't needed. Your own production build process will do the minification for you. And this only provides references to the 'redux' package, so the version chosen by react-redux won't result in duplicates by itself.

akaila:
My main package includes both redux/bin/redux.min.js (which is recommended for prod builds) along with react-redux (which in turn references redux and not minified version of redux). When I run react-native packager on my project, it is obtaining redux twice as a result. We should not be referencing non-minified version of redux from within react-redux.

timdorr:
You don't need to reference pre-minified versions within packages themselves. When everything gets bundled up, the code should be minified and dev-only stuff should be removed. It does this process on the entire bundle *after* it's been built.

You shouldn't refer to pre-built versions of libraries in your code. Just use normal modules and let your build tool optimize when it's done bundling.

akaila:
I would do that except in redux for debug builds there is an exception that pops up indicating I should be including the minified version. As I understood there was some code within if (__dev__) block that is culled in the prod builds.

akaila:
I think I now get what you are saying. Please discard this request and thank you very much for responding quickly !

timdorr:
Yep, and that's what [all the junk above that](https://github.com/reactjs/redux/blob/0139a9601dae810db533762c2bc4a745a37d4cf6/src/index.js#L14-L19) is for. If you have your NODE_ENV environment variable set to "production" (which you should to get a faster production build of React), then all that gets turned into a `if (false) {}` block and will get eliminated from your final bundle.

PR #538, opened 2016-11-07
Add typings

bbenezech:
See #433 for discussion.

I tried using DT's but I found some issues (connect with no argument was not adding dispatch prop, and some API calls where not possible). I fixed the issues for the most common use cases, described below.
The rest of the API should work like it does with DT version.
Please note: **always define templates for your connect calls** or you **will** have issues/subpar type checking (inference engine will replace non-inferable templates type with `{}`).

API Ref that we should targeted as much as possible: https://github.com/reactjs/react-redux/blob/master/docs/api.md#examples

These are the supported use cases that I tested and that 100% type check without any error or leniency:

### 1. Inject `dispatch` and don't listen to store

```typescript
import { Dispatch } from 'redux'
import { checkout } from '../actions'
type AppState = {}

interface OwnProps {
  cartId: number
}

interface P extends OwnProps {
  dispatch: Dispatch<AppState> // try removing it, you'll get an error
}

class Cart extends Component<P, {}> {
  this.props.dispatch(checkout(this.props.cartId))
}

export default connect<AppState, OwnProps>()(Cart)
```
- `AppState` is used as a template for the Dispatch type, as required by redux typings
- `OwnProps` exposes the component with its props signature, without complicated `InferableComponentDecorator` that won't allow me to inject dispatcher.

### 2. Hydrate action creators without subscribing to the store

```typescript
import { Dispatch } from 'redux'
import { checkout } from '../actions'

interface OwnProps {
  cartId: number
}

interface DispatchProps {
  checkout: (cartId: number) => void
}

interface P extends OwnProps, DispatchProps {}

class Cart extends Component<P, {}> {
  this.props.checkout(this.props.cartId) // checkout is hydrated
}

connect<{}, OwnProps, DispatchProps>(null, { checkout })(Cart)
```

### 3. Inject `dispatch` and props from global state

```typescript
import { Dispatch } from 'redux'
import { checkout } from '../actions'

type AppState = {
  items: {
    [cartId: number]: Item[]
  }
}

interface OwnProps {
  cartId: number
}

interface StateProps { 
  items: Item[],
}

interface P extends OwnProps, StateProps {
  dispatch: Dispatch<AppState>
}

class Cart extends Component<P, {}> {
  this.props.dispatch(checkout(this.props.cartId))
  this.props.items.each(item => ...)
}

export default connect<AppState, OwnProps, StateProps>((state, props) => ({
  // state, props and result hash are all completely safe
  items: state.items[props.cartId]
}))(Cart)
```

### 4. Inject props from global state and hydrate action creators

```typescript

type AppState = {
  items: {
    [cartId: number]: Item[]
  }
}

interface OwnProps {
  cartId: number
}

interface StateProps {
  items: Item[],
}

interface DispatchProps {
  checkout: (cartId: number) => void
}

interface P extends OwnProps, StateProps, DispatchProps {}

class Cart extends Component<P, {}> {
  this.props.checkout(this.props.cartId)
  this.props.items.each(item => ...)
}

export default connect<AppState, OwnProps, StateProps, DispatchProps>((state, props) => ({
  items: state.items[props.cartId]
}), { checkout })(Cart)
```
- `AppState` describes your App's state
- `OwnProps` describes your component's outside props (the ones that the parent component should specify)
- `StateProps` describes the component's props that come from the global state
- `DispatchProps` describes the component's props that come from hydrated action creators

Honestly, I only use case **(3.)**

This should get us started. Please feedback!

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -2,7 +2,7 @@ import { ComponentClass, Component, StatelessComponent, ReactNode } from 'react'
 import { Store, Dispatch, ActionCreator } from 'redux';
 
 interface ComponentDecorator<TOriginalProps, TOwnProps> {
-  (component: ComponentClass<TOriginalProps>|StatelessComponent<TOriginalProps>): ComponentClass<TOwnProps>;
+  (component: StatelessComponent<TOriginalProps>|ComponentClass<TOriginalProps>): ComponentClass<TOwnProps>;
 }
 
 /**
--- a/index.d.ts
+++ b/index.d.ts
@@ -63,8 +63,7 @@ interface Options {
  * generic parameter in Store here by using any as the type
  */
 export interface ProviderProps {
-  store?: Store<any>;
-  children?: ReactNode;
+  store: Store<any>;
 }
 
 export class Provider extends Component<ProviderProps, {}> { }
--- a/index.d.ts
+++ b/index.d.ts
@@ -16,22 +16,22 @@ interface ComponentDecorator<TOriginalProps, TOwnProps> {
 export function connect<State, TOwnProps>(): ComponentDecorator<{ dispatch: Dispatch<State> }, TOwnProps>;
 
 export function connect<State, TOwnProps, TStateProps>(
-  mapStateToProps: MapStateToProps<State, TOwnProps, TStateProps>,
+  mapStateToProps: FuncOrSelf<MapStateToProps<State, TOwnProps, TStateProps>>,
 ): ComponentDecorator<TStateProps & { dispatch: Dispatch<State> }, TOwnProps>;
 
 export function connect<State, TOwnProps, TStateProps, TDispatchProps>(
-  mapStateToProps: MapStateToProps<State, TOwnProps, TStateProps>,
-  mapDispatchToProps: MapDispatchToPropsFunction<State, TOwnProps, TDispatchProps>|TDispatchProps
+  mapStateToProps: FuncOrSelf<MapStateToProps<State, TOwnProps, TStateProps>>,
+  mapDispatchToProps: FuncOrSelf<MapDispatchToPropsFunction<State, TOwnProps, TDispatchProps>|TDispatchProps>
 ): ComponentDecorator<TStateProps & TDispatchProps, TOwnProps>;
 
 export function connect<State, TOwnProps, TDispatchProps>(
   mapStateToProps: null,
-  mapDispatchToProps: MapDispatchToPropsFunction<State, TOwnProps, TDispatchProps>|TDispatchProps
+  mapDispatchToProps: FuncOrSelf<MapDispatchToPropsFunction<State, TOwnProps, TDispatchProps>|TDispatchProps>
 ): ComponentDecorator<TDispatchProps, TOwnProps>;
 
 export function connect<State, TOwnProps, TStateProps, TDispatchProps, TMergeProps>(
-  mapStateToProps: MapStateToProps<State, TOwnProps, TStateProps>,
-  mapDispatchToProps: MapDispatchToPropsFunction<State, TOwnProps, TDispatchProps>|TDispatchProps,
+  mapStateToProps: FuncOrSelf<MapStateToProps<State, TOwnProps, TStateProps>>,
+  mapDispatchToProps: FuncOrSelf<MapDispatchToPropsFunction<State, TOwnProps, TDispatchProps>|TDispatchProps>,
   mergeProps: MergeProps<TOwnProps, TStateProps, TDispatchProps, TMergeProps>,
   options?: Options
 ): ComponentDecorator<TStateProps & TDispatchProps, TOwnProps>;
@@ -52,6 +52,8 @@ interface Options {
   pure?: boolean;
   withRef?: boolean;
 }
+
+type FuncOrSelf<T> = T | (() => T);
 
 /**
  * Typescript does not support generic components in tsx yet in an intuïtive way which is the reason we avoid a
--- a/index.d.ts
+++ b/index.d.ts
@@ -21,20 +21,24 @@ export function connect<State, TOwnProps, TStateProps>(
 
 export function connect<State, TOwnProps, TStateProps, TDispatchProps>(
   mapStateToProps: FuncOrSelf<MapStateToProps<State, TOwnProps, TStateProps>>,
-  mapDispatchToProps: FuncOrSelf<MapDispatchToPropsFunction<State, TOwnProps, TDispatchProps>|TDispatchProps>
+  mapDispatchToProps: FuncOrSelf<MapDispatchToPropsFunction<State, TOwnProps, TDispatchProps> | MapDispatchToPropsObject & TDispatchProps>
 ): ComponentDecorator<TStateProps & TDispatchProps, TOwnProps>;
 
 export function connect<State, TOwnProps, TDispatchProps>(
   mapStateToProps: null,
-  mapDispatchToProps: FuncOrSelf<MapDispatchToPropsFunction<State, TOwnProps, TDispatchProps>|TDispatchProps>
+  mapDispatchToProps: FuncOrSelf<MapDispatchToPropsFunction<State, TOwnProps, TDispatchProps> | MapDispatchToPropsObject & TDispatchProps>
 ): ComponentDecorator<TDispatchProps, TOwnProps>;
 
 export function connect<State, TOwnProps, TStateProps, TDispatchProps, TMergeProps>(
   mapStateToProps: FuncOrSelf<MapStateToProps<State, TOwnProps, TStateProps>>,
-  mapDispatchToProps: FuncOrSelf<MapDispatchToPropsFunction<State, TOwnProps, TDispatchProps>|TDispatchProps>,
+  mapDispatchToProps: FuncOrSelf<MapDispatchToPropsFunction<State, TOwnProps, TDispatchProps> | MapDispatchToPropsObject & TDispatchProps>,
   mergeProps: MergeProps<TOwnProps, TStateProps, TDispatchProps, TMergeProps>,
   options?: Options
 ): ComponentDecorator<TStateProps & TDispatchProps, TOwnProps>;
+
+interface MapDispatchToPropsObject {
+  [name: string]: ActionCreator<any>;
+}
 
 interface MapStateToProps<State, TOwnProps, TStateProps> {
   (state: State, ownProps: TOwnProps): TStateProps;
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,66 @@
+import { ComponentClass, Component, StatelessComponent, ReactNode } from 'react';
+import { Store, Dispatch, ActionCreator } from 'redux';
+
+interface ComponentDecorator<TOriginalProps, TOwnProps> {
+  (component: ComponentClass<TOriginalProps>|StatelessComponent<TOriginalProps>): ComponentClass<TOwnProps>;
+}
+
+/**
+ * Following 3 functions cover all possible ways connect could be invoked
+ *
+ * - State: Redux state interface (the same one used by Store<S>)
+ * - TStateProps: Result of MapStateToProps
+ * - TDispatchProps: Result of MapDispatchToProps
+ * - TOwnProps: Props passed to the wrapping component
+ */
+export function connect<TOwnProps>(): ComponentDecorator<{ dispatch: Dispatch<any> }, TOwnProps>;
+
+export function connect<State, TStateProps, TDispatchProps, TOwnProps>(
+  mapStateToProps: MapStateToProps<State, TStateProps, TOwnProps>,
+  mapDispatchToProps?: MapDispatchToPropsFunction<State, TDispatchProps, TOwnProps>|MapDispatchToPropsObject
+): ComponentDecorator<TStateProps & TDispatchProps, TOwnProps>;
+
+export function connect<State, TStateProps, TDispatchProps, TOwnProps>(
+  mapStateToProps: MapStateToProps<State, TStateProps, TOwnProps>,
+  mapDispatchToProps: MapDispatchToPropsFunction<State, TDispatchProps, TOwnProps>|MapDispatchToPropsObject,
+  mergeProps: MergeProps<TStateProps, TDispatchProps, TOwnProps>,
+  options?: Options
+): ComponentDecorator<TStateProps & TDispatchProps, TOwnProps>;
+
+interface MapStateToProps<State, TStateProps, TOwnProps> {
+  (state: State, ownProps?: TOwnProps): TStateProps;
+}
+
+/**
+ * State is not actually used here but included for consistency with Redux typings and MapStateToProps.
+ */
+interface MapDispatchToPropsFunction<State, TDispatchProps, TOwnProps> {
+  (dispatch: Dispatch<State>, ownProps?: TOwnProps): TDispatchProps;
+}
+
+/**
+ * Any since not every ActionCreator returns the same Action
+ */
+interface MapDispatchToPropsObject {
+  [name: string]: ActionCreator<any>;
+}
+
+interface MergeProps<TStateProps, TDispatchProps, TOwnProps> {
+  (stateProps: TStateProps, dispatchProps: TDispatchProps, ownProps: TOwnProps): TStateProps & TDispatchProps;
+}
+
+interface Options {
+  pure?: boolean;
+  withRef?: boolean;
+}
+
+/**
+ * Typescript does not support generic components in tsx yet in an intuïtive way which is the reason we avoid a
+ * generic parameter in Store here by using any as the type
+ */
+export interface ProviderProps {
+  store?: Store<any>;
+  children?: ReactNode;
+}
+
+export class Provider extends Component<ProviderProps, {}> { }
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,4 +1,4 @@
-import { ComponentClass, Component, StatelessComponent, ReactNode } from 'react';
+import { ComponentClass, Component, StatelessComponent } from 'react';
 import { Store, Dispatch, ActionCreator } from 'redux';
 
 interface ComponentDecorator<TOriginalProps, TOwnProps> {
--- a/index.d.ts
+++ b/index.d.ts
@@ -6,7 +6,7 @@ interface ComponentDecorator<TOriginalProps, TOwnProps> {
 }
 
 /**
- * Following 3 functions cover all possible ways connect could be invoked
+ * Following 5 functions cover all possible ways connect could be invoked
  *
  * - State: Redux state interface (the same one used by Store<S>)
  * - TOwnProps: Props passed to the wrapping component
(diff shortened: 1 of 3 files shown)


index.d.ts
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ * Any since not every ActionCreator returns the same Action
+ */
+interface MapDispatchToPropsObject<TDispatchProps> {
+  [name: string]: ActionCreator<TDispatchProps>;
+}
+
+interface MergeProps<TStateProps, TDispatchProps, TOwnProps> {
+  (stateProps: TStateProps, dispatchProps: TDispatchProps, ownProps: TOwnProps): TStateProps & TDispatchProps;
```
aikoven: That's not correct. `mergeProps` can return anything, it should be a type parameter of `MergeProps` as well.

index.d.ts
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ * Following 3 functions cover all possible ways connect could be invoked
+ *
+ * - State: Redux state interface (the same one used by Store<S>)
+ * - TStateProps: Result of MapStateToProps
+ * - TDispatchProps: Result of MapDispatchToProps
+ * - TOwnProps: Props passed to the wrapping component
+ */
+export function connect<State, TOwnProps>(): ComponentDecorator<{ dispatch: Dispatch<State> }, TOwnProps>;
```
aikoven: I guess it shouldn't replace props with `{dispatch}`, but merge them

index.d.ts
```diff
@@ -0,0 +1,70 @@
...
+): ComponentDecorator<TStateProps & TDispatchProps, TOwnProps>;
+
+export function connect<State, TOwnProps, TStateProps, TDispatchProps>(
+  mapStateToProps: MapStateToProps<State, TStateProps, TOwnProps>|null,
+  mapDispatchToProps: MapDispatchToPropsFunction<State, TDispatchProps, TOwnProps>|TDispatchProps,
+  mergeProps: MergeProps<TStateProps, TDispatchProps, TOwnProps>,
+  options?: Options
+): ComponentDecorator<TStateProps & TDispatchProps, TOwnProps>;
+
+interface MapStateToProps<State, TStateProps, TOwnProps> {
```
aikoven: Why not have type parameters in the same order as in the signature?

index.d.ts
```diff
@@ -0,0 +1,70 @@
...
+ * - State: Redux state interface (the same one used by Store<S>)
+ * - TStateProps: Result of MapStateToProps
+ * - TDispatchProps: Result of MapDispatchToProps
+ * - TOwnProps: Props passed to the wrapping component
+ */
+export function connect<State, TOwnProps>(): ComponentDecorator<{ dispatch: Dispatch<State> }, TOwnProps>;
+
+export function connect<State, TOwnProps, TStateProps>(
+  mapStateToProps: MapStateToProps<State, TStateProps, TOwnProps>,
+): ComponentDecorator<TStateProps & { dispatch: Dispatch<State> }, TOwnProps>;
```
aikoven: Wrapped component should get all of `TStateProps`, `TOwnProps` and `dispatch`

index.d.ts
```diff
@@ -0,0 +1,70 @@
...
+
+export function connect<State, TOwnProps, TStateProps, TDispatchProps>(
+  mapStateToProps: MapStateToProps<State, TStateProps, TOwnProps>|null,
```
aikoven: Let's make another signature for `null` as first argument. We'd get stronger typings for this case.

index.d.ts
```diff
@@ -0,0 +1,70 @@
...
+  mapDispatchToProps: MapDispatchToPropsFunction<State, TDispatchProps, TOwnProps>|TDispatchProps
```
aikoven: I think this should be made more strict. The case of `object` as `mapDispatchToProps` expects an object with action creators as values, not just any object. Let's make another signature for this :)

index.d.ts
```diff
@@ -0,0 +1,70 @@
...
+): ComponentDecorator<TStateProps & TDispatchProps, TOwnProps>;
+
+export function connect<State, TOwnProps, TStateProps, TDispatchProps>(
+  mapStateToProps: MapStateToProps<State, TStateProps, TOwnProps>|null,
```
aikoven: `mapState` / `mapDispatch` can be factory functions. See https://github.com/reactjs/reselect/tree/v2.5.3#sharing-selectors-with-props-across-multiple-components

index.d.ts
```diff
@@ -0,0 +1,70 @@
...
+ * State is not actually used here but included for consistency with Redux typings and MapStateToProps.
+ */
+interface MapDispatchToPropsFunction<State, TDispatchProps, TOwnProps> {
+  (dispatch: Dispatch<State>, ownProps: TOwnProps): TDispatchProps;
+}
+
+/**
+ * Any since not every ActionCreator returns the same Action
+ */
+interface MapDispatchToPropsObject<TDispatchProps> {
```
aikoven: This should be used as a possible type for `mapDispatch`

index.d.ts
```diff
@@ -0,0 +1,70 @@
...
+  [name: string]: ActionCreator<TDispatchProps>;
```
aikoven: That's not true, type argument of ActionCreator is the type of created Action. Here it could be anything

index.d.ts
```diff
@@ -0,0 +1,70 @@
...
+  withRef?: boolean;
+}
+
+/**
+ * Typescript does not support generic components in tsx yet in an intuïtive way which is the reason we avoid a
+ * generic parameter in Store here by using any as the type
+ */
+export interface ProviderProps {
+  store?: Store<any>;
+  children?: ReactNode;
```
aikoven: That's not needed as `children` prop is automatically added to components defined as classes.

index.d.ts
```diff
@@ -0,0 +1,70 @@
...
+  pure?: boolean;
+  withRef?: boolean;
+}
+
+/**
+ * Typescript does not support generic components in tsx yet in an intuïtive way which is the reason we avoid a
+ * generic parameter in Store here by using any as the type
+ */
+export interface ProviderProps {
+  store?: Store<any>;
```
aikoven: Is it really optional?

index.d.ts
bbenezech: Same as previous one. ComponentDecorator exposes the exported component with `TOwnProps` outside, with `TStateProps & { dispatch: Dispatch<State> }` satisfied by React-Redux.
I think you are possibly misreading what `ComponentDecorator` does or how which props of the HOC should be exposed outside.
I've tested that part extensively in my (big) app and it really behaves as supposed to.

index.d.ts
aikoven: It doesn't. ComponentDecorator in this case states that it attepts component with props `{dispatch: Dispatch<State>}` and returns component with props `TOwnProps`.

In reality the effect of `connect` is adding `dispatch` to other properties.

index.d.ts
bbenezech: This have been removed totally as well.

index.d.ts
bbenezech: This have been removed totally as well.

index.d.ts
bbenezech: Oh. Ok, sorry. Makes sense.

index.d.ts
bbenezech: I guess not. This should be fixed.

index.d.ts
bbenezech: Good. Let's fix this.

bbenezech: Done.

index.d.ts
bbenezech: Done.

aikoven:
> Please note: always define templates for your connect calls or you will have issues/subpar type checking (inference engine will replace non-inferable templates type with {}).

That's not generally true, the pattern I use everywhere is this:
```ts
type State = {foo: string};

const Component = connect(
  (state: State) => {
    return {
      bar: state.foo,
    };
  }
)(props => {
  props.bar  // inferred as `string`
})
```
This saves a lot of typing clutter. 

Even if you connect component classes, you can omit type parameters if only you annotate arguments for `mapState` / `mapDispatch`:
```ts
const connector = connect(
  (state: MyState, props: MyProps) => {...}
)
```
This way `connector` would be a decorator for components accepting `MyProps` plus stuff returned from `mapState` / `mapDispatch` (inferred automatically). And returned component would accept `MyProps`.

index.d.ts
bbenezech: Done

index.d.ts
```diff
@@ -0,0 +1,70 @@
+import { ComponentClass, Component, StatelessComponent, ReactNode } from 'react';
+import { Store, Dispatch, ActionCreator } from 'redux';
+
+interface ComponentDecorator<TOriginalProps, TOwnProps> {
+  (component: ComponentClass<TOriginalProps>|StatelessComponent<TOriginalProps>): ComponentClass<TOwnProps>;
```
bbenezech: Done

bbenezech:
@aikoven Sure.
```typescript
const connector = connect(
  (state: MyState, props: MyProps) => {...}
)
```

You are not type checking StateProps return. If you typo your hash, you won't know it. Then you need:

```typescript
const connector = connect(
  (state: MyState, props: MyProps): StateProps  => {...}
)
```

But then why not
```typescript
const connector<MyState, MyProps, StateProps> = connect(
  (stats, props)  => {...}
)
```

The advantage is you get an error if you forget one of the template arguments, whereas you forgot to type check StateProps and loss some type checking, hence my warning.

But this is borderline bikeshedding.

index.d.ts
bbenezech: Ok. I felt that not exposing state props and dispatch prop to parent component was actually a feature.
Is there a valid use case where we should let users overwrite HOC's props?
Or is it more that you feel type checking should try to stick to reality as much as possible?

index.d.ts
bbenezech: Done.

index.d.ts
bbenezech: Replied below

index.d.ts
bbenezech: Done

index.d.ts
bbenezech: Done

index.d.ts
bbenezech: This is the road taken in DT: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/react-redux/react-redux.d.ts#L73

I was against it because you lose the most important part (for me at least): make sure your `TDispatchProps` matches what you are returning from `mapDispatchToProps`.

With DT's version, this doesn't error:
```typescript
import { checkout } from 'actions'

interface TDispatchProps {
  checkout: (cartId: string) => void // should be number
  // or
  checout: (cartId: number) => void // typo
}
```

This works and check for both issues (TDispatchProps and hash of ActionCreator): https://github.com/reactjs/react-redux/pull/538/commits/670c85146170ff712587782303a2f7afbe5fbda3

Done

bbenezech:
Job's done! 🎉 
Thanks for your time @aikoven ! Your review was terrifically on spot! 🎯

timdorr:
Can this actually get built against the `next` branch? It would be a good idea for typings to show up in a major, rather than as a patch (they could be argued as minor, but I think we'd run into complaints there too).

aikoven:
> You are not type checking StateProps return. If you typo your hash, you won't know it.

That's just a matter of taste. I prefer not to specify an interface for an object that is consumed just a few lines later. My point is that you don't *have to* always specify type parameters because TS is quite good in inferring return types.

> If we do this:
> 
> ```ts
> ComponentDecorator<TStateProps & { dispatch: Dispatch<State> }, TStateProps & { dispatch?: Dispatch<State> } & TOwnProps>
> ```

That's not what I meant. The right one is:
```ts
ComponentDecorator<TStateProps & {dispatch: Dispatch<State>} & TOwnProps, TOwnProps>;
```

Own props passed from outside, State props and `dispatch` merged into and passed to the wrapped component.

index.d.ts
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ * Following 5 functions cover all possible ways connect could be invoked
+ *
+ * - State: Redux state interface (the same one used by Store<S>)
+ * - TOwnProps: Props passed to the wrapping component
+ * - TStateProps: Result of MapStateToProps
+ * - TDispatchProps: Result of MapDispatchToProps
+ */
+export function connect<State, TOwnProps>(): ComponentDecorator<{ dispatch: Dispatch<State> }, TOwnProps>;
```
aikoven: Correct return type:
```ts
ComponentDecorator<{dispatch: Dispatch<State>} & TOwnProps, TOwnProps>;
```

index.d.ts
```diff
@@ -0,0 +1,70 @@
...
+
+export function connect<State, TOwnProps, TStateProps>(
+  mapStateToProps: FuncOrSelf<MapStateToProps<State, TOwnProps, TStateProps>>,
+): ComponentDecorator<TStateProps & { dispatch: Dispatch<State> }, TOwnProps>;
```
aikoven: Correct return type:
```ts
ComponentDecorator<TStateProps & {dispatch: Dispatch<State>} & TOwnProps, TOwnProps>;
```

index.d.ts
```diff
@@ -0,0 +1,70 @@
...
+
+export function connect<State, TOwnProps, TStateProps, TDispatchProps>(
+  mapStateToProps: FuncOrSelf<MapStateToProps<State, TOwnProps, TStateProps>>,
+  mapDispatchToProps: FuncOrSelf<MapDispatchToPropsFunction<State, TOwnProps, TDispatchProps> | MapDispatchToPropsObject & TDispatchProps>
+): ComponentDecorator<TStateProps & TDispatchProps, TOwnProps>;
```
aikoven: Correct return type:
```ts
ComponentDecorator<TStateProps & TDispatchProps & TOwnProps, TOwnProps>;
```

index.d.ts
```diff
@@ -0,0 +1,70 @@
...
+ */
+export function connect<State, TOwnProps>(): ComponentDecorator<{ dispatch: Dispatch<State> }, TOwnProps>;
+
+export function connect<State, TOwnProps, TStateProps>(
+  mapStateToProps: FuncOrSelf<MapStateToProps<State, TOwnProps, TStateProps>>,
+): ComponentDecorator<TStateProps & { dispatch: Dispatch<State> }, TOwnProps>;
+
+export function connect<State, TOwnProps, TStateProps, TDispatchProps>(
+  mapStateToProps: FuncOrSelf<MapStateToProps<State, TOwnProps, TStateProps>>,
+  mapDispatchToProps: FuncOrSelf<MapDispatchToPropsFunction<State, TOwnProps, TDispatchProps> | MapDispatchToPropsObject & TDispatchProps>
```
aikoven: In a case of passing object to `mapDispatchToProps`, what is `TDispatchProps`? What's the difference with `MapDispatchToPropsObject`?

I think it should be a different signature for the case when `mapDispatchToProps` is object, something like this:
```ts
function connect<State, TOwnProps, TStateProps, TDispatchProps extends MapDispatchToPropsObject>(
  mapStateToProps: FuncOrSelf<MapStateToProps<State, TOwnProps, TStateProps>>,
  mapDispatchToProps: TDispatchProps
): ComponentDecorator<TStateProps & TDispatchProps & TOwnProps, TOwnProps>;
```

index.d.ts
```diff
@@ -0,0 +1,70 @@
...
+): ComponentDecorator<TStateProps & TDispatchProps, TOwnProps>;
+
+export function connect<State, TOwnProps, TDispatchProps>(
+  mapStateToProps: null,
+  mapDispatchToProps: FuncOrSelf<MapDispatchToPropsFunction<State, TOwnProps, TDispatchProps> | MapDispatchToPropsObject & TDispatchProps>
+): ComponentDecorator<TDispatchProps, TOwnProps>;
```
aikoven: Correct return type:
```ts
ComponentDecorator<TDispatchProps & TOwnProps, TOwnProps>;
```

Also see comment above on the case when `mapDispatchToProps` is object.

index.d.ts
```diff
@@ -0,0 +1,70 @@
...
+
+export function connect<State, TOwnProps, TStateProps, TDispatchProps, TMergeProps>(
+  mapStateToProps: FuncOrSelf<MapStateToProps<State, TOwnProps, TStateProps>>,
+  mapDispatchToProps: FuncOrSelf<MapDispatchToPropsFunction<State, TOwnProps, TDispatchProps> | MapDispatchToPropsObject & TDispatchProps>,
+  mergeProps: MergeProps<TOwnProps, TStateProps, TDispatchProps, TMergeProps>,
+  options?: Options
+): ComponentDecorator<TStateProps & TDispatchProps, TOwnProps>;
```
aikoven: Correct return type:
```ts
ComponentDecorator<TMergeProps, TOwnProps>;
```

Also see comment above on the case when mapDispatchToProps is object.

index.d.ts
```diff
@@ -0,0 +1,70 @@
...
+interface MapStateToProps<State, TOwnProps, TStateProps> {
+  (state: State, ownProps: TOwnProps): TStateProps;
+}
+
+interface MapDispatchToPropsFunction<State, TOwnProps, TDispatchProps> {
+  (dispatch: Dispatch<State>, ownProps: TOwnProps): TDispatchProps;
+}
+
+interface MergeProps<TOwnProps, TStateProps, TDispatchProps, TMergeProps> {
+  (ownProps: TOwnProps, stateProps: TStateProps, dispatchProps: TDispatchProps): TMergeProps;
```
aikoven: The order of arguments is incorrect, the right one would be: `stateProps, dispatchProps, ownProps`

index.d.ts
bbenezech: Done

index.d.ts
bbenezech: Done

index.d.ts
bbenezech: Done

index.d.ts
bbenezech: Done

index.d.ts
bbenezech: Done

index.d.ts
bbenezech: Oh my... I thought I was dealing with named arguments...
Fixed

index.d.ts
bbenezech: `TDispatchProps` may be set by the user. It's the list of dispatchable action creators he intends to use in the component.
`MapDispatchToPropsObject` is a constraint we add to make sure values of the hash are action creators.

Adding a different signature is problematic because the combination will be very large. Same issue as with `mapStateToProps: null`.

timdorr:
@bbenezech Can you rebase against `next` and switch out the base for this PR (hit Edit at the top of the page)? Thanks!

timdorr:
@bbenezech Looks like there's a conflict because the git branch isn't based on `next`. Can you rebase and force push your branch? I don't know if I can do that on my end with the collaborator permission on the PR. Thanks for working with me on this! 👍

bbenezech:
I can't change my own branch in edit. I opened a new PR.

Issue #525, opened 2016-10-19
Connected text input cursor position reset on input


thefloweringash:
I have an input field connected to a part of a redux store. When I type in the text field the cursor position is reset to the end of the text field. For example filling in "123", then attempting to insert "abc" at the beginning results in "a123bc". I've been unable to isolate the exact circumstances that cause this, but have a small reproduction at https://jsfiddle.net/1fm2cczq/. Occurs on 5.0.0-beta.3. Does not occur on 4.4.5.

jimbolla:
Hmmm yeah good find. I can reproduce this in my own app. This seems related to facebook/react#955, and my first hunch is that changing the subscriptions from firing bottom-up to top-down reveals this behavior. Will need to investigate further.

jimbolla:
As an experiment, I wrapped an `<input>` with this component to see what happens, and this does correct the behavior. So it seems to me that the issue is related to controlled inputs needing to be updated earlier or else React sorta goes off the rails with cursor position.

```js
class Input extends React.Component {
  constructor(props, ...args) {
    super(props, ...args);
    this.state = { value: props.value };
  }

  componentWillReceiveProps(nextProps) {
    if (this.state.value !== nextProps.value) {
      this.setState({ value: nextProps.value });
    }
  }

  onChange(evt) {
    this.setState({ value: evt.target.value }, () => this.props.onChange(evt));
  }

  render() {
    return (<input {...this.props} {...this.state} onChange={this.onChange} />);
  }
}
```

jimbolla:
I created a custom build of react-redux v5 that disables the top-down ordering so that subscriptions behave like v4. The cursor issue went away.  So I think I know what the problem is, but I don't yet have a great solution. So far, the first idea I have is to provide another option for `connect` that tells it to prioritize its store subscription before those of its parents. This option would have to be enabled for controlled inputs and textareas (I'm assuming) as a workaround for facebook/react#955 until that's fixed (if ever.)

markerikson:
Ew.  This... sounds nasty.

jimbolla:
@markerikson Indeed it is.

As another experiment, I just tried [switching React for Preact](https://preactjs.com/guide/switching-to-preact) in my own project and the bug goes away.

markerikson:
Hmm.  Pinging @gaearon and @timdorr for thoughts...

jimbolla:
[...]

PR #540, opened 2016-11-09
Adds react15CompatibilityMode setting

jimbolla:
`react15CompatibilityMode` is intended to be a temporary setting until our official solution is "use React v16." When set to true, which is the default in `<Provider>`, subscriptions are **not** re-ordered to fire top-down. This avoids the bug related to #525 but loses some of the perf gains and re-introduces the issues with state and ownProps being out of sync. This should behave pretty much like react-redux v4.

A developer that upgrades to this version w/o making any other changes should not experience the #525 bug out of the box, but won't get the good stuff mentioned above.

The recommended upgrade path is to set `react15CompatibilityMode` to false at the `<Provider>` level and then set it to true at the component level for all `connect`ed text inputs. This can be done either as a prop (confirmed working with redux-form's `<Field>` component) or as an options arg to `connect`. Those inputs are now vulnerable to the state/props sync issue that exists in v4, but will not have the #525 cursor issue.

Apps using React v16 (or whatever version includes https://github.com/facebook/react/pull/8204), or using Preact as a drop-in replacement for React should not have to set `react15CompatibilityMode` to true at the component level at all... just set `<Provider react15CompatibilityMode={false}>` .

Once React v16 becomes the popular choice, we can switch the default from true to false. Once v15 support is dropped, we can backout this change completely.

### DIFF
--- a/src/components/Provider.js
+++ b/src/components/Provider.js
@@ -20,7 +20,7 @@ function warnAboutReceivingStore() {
 
 export default class Provider extends Component {
   getChildContext() {
-    return { store: this.store }
+    return { store: this.store, react15CompatibilityMode: this.props.react15CompatibilityMode }
   }
 
   constructor(props, context) {
@@ -45,10 +45,15 @@ if (process.env.NODE_ENV !== 'production') {
 }
 
 Provider.propTypes = {
+  react15CompatibilityMode: PropTypes.bool,
   store: storeShape.isRequired,
   children: PropTypes.element.isRequired
 }
 Provider.childContextTypes = {
+  react15CompatibilityMode: PropTypes.bool.isRequired,
   store: storeShape.isRequired
+}
+Provider.defaultProps = {
+  react15CompatibilityMode: true
 }
 Provider.displayName = 'Provider'
--- a/src/components/Provider.js
+++ b/src/components/Provider.js
@@ -20,7 +20,7 @@ function warnAboutReceivingStore() {
 
 export default class Provider extends Component {
   getChildContext() {
-    return { store: this.store, react15CompatibilityMode: this.props.react15CompatibilityMode }
+    return { store: this.store }
   }
 
   constructor(props, context) {
@@ -45,15 +45,10 @@ if (process.env.NODE_ENV !== 'production') {
 }
 
 Provider.propTypes = {
-  react15CompatibilityMode: PropTypes.bool,
   store: storeShape.isRequired,
   children: PropTypes.element.isRequired
 }
 Provider.childContextTypes = {
-  react15CompatibilityMode: PropTypes.bool.isRequired,
   store: storeShape.isRequired
-}
-Provider.defaultProps = {
-  react15CompatibilityMode: true
 }
 Provider.displayName = 'Provider'
--- a/src/components/connectAdvanced.js
+++ b/src/components/connectAdvanced.js
@@ -18,7 +18,7 @@ export default function connectAdvanced(
 
     Access to dispatch is provided to the factory so selectorFactories can bind actionCreators
     outside of their selector as an optimization. Options passed to connectAdvanced are passed to
-    the selectorFactory, along with displayName and WrappedComponent, as the second argument. 
+    the selectorFactory, along with displayName and WrappedComponent, as the second argument.
 
     Note that selectorFactory is responsible for all caching/memoization of inbound and outbound
     props. Do not use connectAdvanced directly without memoizing results between calls to your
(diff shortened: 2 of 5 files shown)


src/components/Provider.js
```diff
@@ -20,7 +20,7 @@ function warnAboutReceivingStore() {
 
 export default class Provider extends Component {
   getChildContext() {
-    return { store: this.store }
+    return { store: this.store, react15CompatibilityMode: this.props.react15CompatibilityMode }
```
timdorr: This shouldn't be on `<Provider>`. It's on a per-component basis, like `pure`. I wouldn't worry about context at all.

jimbolla: Provider is where you set the default, so you don't have to change it on every single connected component when you want the new behavior.

timdorr: But you shouldn't be blanket applying a perf-reducing setting when it may not apply to most connected components. Just make it default at the `connect()` level and remove the context stuff.

jimbolla: I interpreted @gaearon's messages as we wanted to impose the least amount of breaking changes as possible, that's why I made it a global default setting which is "work like it used to, even if that means being not as fast as possible and maybe get stale props". It's certainly less code in react-redux to not have a global setting and require devs to go enable the compat setting on all their connected textboxes. I'm fine with either approach but I just need to be told which way we're going.

timdorr: Correct, and for 5.0 it will be opt-out of the slower (but more compatible) behavior. For 5.1, which will wait on React 16, it will be opt-in for those who need to stick with React <=15.

timdorr:
Have you run any perf tests on this? I assume it basically gets us back to 4.0 levels of speed, correct? That should be fine, since we're not regressing on speed at all. We can make a big warning label at the top of the release notes about it so no one misses the opt-out.

jimbolla:
@timdorr Currently swamped with other things. Will perf test when I have a large block of time. Hopefully this weekend.

timdorr:
Added a commit to remove the Provider API as requested. Again, we should be having people set this on a per-component basis, defaulting to `true` for 5.0 and defaulting to `false` for 5.1. This ensures a safe release that doesn't upset users, while still giving them an option to enable the more performant option if they deem it safe.

Also, I rebased this against the latest `next`, so the tests are now running against React 15. I'm sure the same behavior existed in 0.14, but this ensures we're dealing with the most popular version in use.

gnoff:
@timdorr why remove the provider api? I want to use this compat mode for redux-form components but want everything else in the app to use the efficient sub ordering. Having to set this on each connected component seems an unnecessary chore. have it be defaulted to true in 5.0 but why make opting into performance more difficult than it needs to be?

timdorr:
Just wrap connect and curry the options:

`export function fastConnect(a,b,c,options) { return connect(a,b,c, { react15CompatibilityMode: false, ...options }) }`

(On my phone, so I hope that came out right...)

gnoff:
This works but it also means code modding / find replacing a large number of files with the need to do so again as soon as 5.1 comes out. I'm not adamant that this get added back in but the trade offs aren't cost free for users ( I realize they aren't cost free for the lib either since it requires supporting ). I could be misguided here but I think the cost is higher on lib consumers. Just lending my annecdote to @jimbolla intuition that a configurable global default might be desired.

jimbolla:
@timdorr I'll concede taking it out of provider, but we should leave in the ability to pass the setting from props/context. Props is especially important if you want to control it through a library such as redux-form that hides the `connect` call.

timdorr:
Sorry, I'm not trying to be dismissive. Obviously I want this to be an awesome release for most users. I haven't run into a situation like this before where we essentially need to define a global to configure a library. I do want to get this right and not just shove something somewhere, because I feel like we'll live to regret that decision 😱 

What about a singleton module? 

```js
import { render } from 'react-dom'
import { setReact15CompatibilityMode } from 'react-redux'

setReact15CompatibilityMode(false)

render(<div />, root)
```

It could also be generic: `setDefaultOptions({ ... })`

jimbolla:
Would it make sense to just put the method on connect instead of being a top level export? Like this:

```js
connect.setReact15CompatibilityMode(false);
```

I think it makes it more clear that it's a setting that affects `connect` like this.

markerikson:
Frankly it all seems ugly any way you look at it, but I could get behind adding a method to `connect` like that.

timdorr:
@jimbolla Yeah, I think that works better. Maybe call it `connect.setDefaultReact15CompatibilityMode()` so it's clear that it's global and changes the default.

jimbolla:
Added `connect.setDefaultReact15CompatibilityMode()` to this PR

gnoff:
@jimbolla @timdorr @markerikson thanks all

timdorr:
Cool. Let's do this thing!

PR #541, opened 2016-11-09
add TypeScript typings

bbenezech:
Follow up to #538 
Rebased against next

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,5 +4,6 @@ node_js:
 script:
   - npm run lint
   - npm run test:cov
+  - npm run test:typescript
 after_success:
   - npm run coverage
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,70 @@
+import { ComponentClass, Component, StatelessComponent } from 'react';
+import { Store, Dispatch, ActionCreator } from 'redux';
+
+interface ComponentDecorator<TOriginalProps, TOwnProps> {
+  (component: StatelessComponent<TOriginalProps>|ComponentClass<TOriginalProps>): ComponentClass<TOwnProps>;
+}
+
+/**
+ * Following 5 functions cover all possible ways connect could be invoked
+ *
+ * - State: Redux state interface (the same one used by Store<S>)
+ * - TOwnProps: Props passed to the wrapping component
+ * - TStateProps: Result of MapStateToProps
+ * - TDispatchProps: Result of MapDispatchToProps
+ */
+function connect<State, TOwnProps>(): ComponentDecorator<{ dispatch: Dispatch<State> } & TOwnProps, TOwnProps>;
+
+function connect<State, TOwnProps, TStateProps>(
+  mapStateToProps: FuncOrSelf<MapStateToProps<State, TOwnProps, TStateProps>>,
+): ComponentDecorator<TStateProps & { dispatch: Dispatch<State> } & TOwnProps, TOwnProps>;
+
+function connect<State, TOwnProps, TStateProps, TDispatchProps>(
+  mapStateToProps: FuncOrSelf<MapStateToProps<State, TOwnProps, TStateProps>>,
+  mapDispatchToProps: FuncOrSelf<MapDispatchToPropsFunction<State, TOwnProps, TDispatchProps> | MapDispatchToPropsObject & TDispatchProps>
+): ComponentDecorator<TStateProps & TDispatchProps & TOwnProps, TOwnProps>;
+
+function connect<State, TOwnProps, TDispatchProps>(
+  mapStateToProps: null,
+  mapDispatchToProps: FuncOrSelf<MapDispatchToPropsFunction<State, TOwnProps, TDispatchProps> | MapDispatchToPropsObject & TDispatchProps>
+): ComponentDecorator<TDispatchProps & TOwnProps, TOwnProps>;
+
+function connect<State, TOwnProps, TStateProps, TDispatchProps, TMergeProps>(
+  mapStateToProps: FuncOrSelf<MapStateToProps<State, TOwnProps, TStateProps>>,
+  mapDispatchToProps: FuncOrSelf<MapDispatchToPropsFunction<State, TOwnProps, TDispatchProps>| MapDispatchToPropsObject & TDispatchProps>,
+  mergeProps: MergeProps<TOwnProps, TStateProps, TDispatchProps, TMergeProps>,
+  options?: Options
+): ComponentDecorator<TMergeProps, TOwnProps>;
+
+interface MapDispatchToPropsObject {
+  [name: string]: ActionCreator<any>;
+}
+
+interface MapStateToProps<State, TOwnProps, TStateProps> {
+  (state: State, ownProps: TOwnProps): TStateProps;
+}
+
+interface MapDispatchToPropsFunction<State, TOwnProps, TDispatchProps> {
+  (dispatch: Dispatch<State>, ownProps: TOwnProps): TDispatchProps;
+}
+
+interface MergeProps<TOwnProps, TStateProps, TDispatchProps, TMergeProps> {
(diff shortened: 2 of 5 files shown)


aikoven:
Continuing on previous thread:

I agree that having a combination of each allowed type for each overload makes too much bloat. 

I think we can omit the case when `mapStateToProps: null`, because if we have `connect(null)`, the `TStateProps` will be inferred to `{}` which is not that bad, because after merging `TStateProps & TOwnProps` becomes the same as just `TOwnProps`.

But for the case when second argument is object, I think we must include another signature, because in this case `TDispatchProps` is constrained to object of action creators, not just any object.

There are a few other things we could improve, but I don't have enough time right now for a thorough review. Will check back in a couple days.

bbenezech:
@aikoven 

> But for the case when second argument is object, I think we must include another signature, because in this case TDispatchProps is constrained to object of action creators, not just any object.

We ask `mapDispatchToProps` to be a `MapDispatchToPropsObject & TDispatchProps`

Not sure I understand.

And I tried 
```typescript
function connect<State, TOwnProps, TStateProps, TDispatchProps extends MapDispatchToPropsObject>(
  mapStateToProps: FuncOrSelf<MapStateToProps<State, TOwnProps, TStateProps>>|null,
  mapDispatchToProps: FuncOrSelf<TDispatchProps>,
): ComponentDecorator<TStateProps & TDispatchProps & TOwnProps, TOwnProps>;

```
can' t get past
```
[ts] Type 'DispatchProps' does not satisfy the constraint 'MapDispatchToPropsObject'.
       Index signature is missing in type 'DispatchProps'.
```
TS won't associate
```typescript
interface DispatchProps {
  someAction: any
}
```
with
```typescript
interface MapDispatchToPropsObject {
  [name: string]: ActionCreator<any>;
}
```

aikoven:
@bbenezech I've created a PR for your PR :) https://github.com/bbenezech/react-redux/pull/1

Added tests. Some of them fail because of `MapDispatchToPropsObject`, particularly:
```ts
  // typings:expect-error
  connect(
    null,
    {
      handleClick: () => 'not-an-action',
    },
  );

  connect(
    null,
    {
      handleClick: () => (dispatch: Dispatch<MyState>) => {
        dispatch({type: 'SOME_ACTION'});
      },
    },
  );
```
The first one should fail, the second one shouldn't. But we can't check it correctly because we don't know beforehand what kinds of values can be `dispatch`ed. So it seems that we have to allow `mapDispatchToProps` to be any object.

bbenezech:
@aikoven I narrowed allowed Actions. I am quite satisfied with the result. It fixes your test failure (i.e. it now fails). There is a very tiny little hack for Thunks (we provide the wrong return type if TDispatchProps is not provided, covariantly compatible of course), but it is going to make everyone life's so much easier I think it is a must. I added tests around it to be sure that behavior is perfect in any cases.

Have a look and tell me what you think.

aikoven:
Good work!

I'm not sure that `react-redux` should be explicitly concerned with `redux-thunk`. What about other middlewares that could add support for dispatching other stuff? I think the best we can do now is to allow any function as *action creator*. By the way, if an object is passed as 2nd argument to `connect`, it just uses `bindActionCreators` from Redux, so why don't we align it with Redux typings? https://github.com/reactjs/redux/blob/f8ec3ef1c3027d6959c85c97459c109574d28b3c/index.d.ts#L313-L355

Also, since we're working on the `next` branch, we have to add typings for newly added [`connectAdvanced`](https://github.com/reactjs/react-redux/blob/next/docs/api.md#connectadvancedselectorfactory-connectoptions).

bbenezech:
@timdorr Why is it closed? Should we stop our effort to release the typings inside react-redux repo?

timdorr:
I think this was auto closed by my adjusting of the branches on the repo. The next branch was deleted, so that's probably why. I think we just need to change the branch and then I can open it. I'm on mobile at the moment, but will do that soon.

bbenezech:
@timdorr Awesome thanks!

rokoroku:
@bbenezech Why don't we use Partial<T> type (introduced in Typescript 2.1) for annoying TOwnProps, TStateProps, TDispatchProps things? we can just pass two types to connect: TState and  TProps.

aikoven:
@rokoroku What's your intended use of `Partial` here?

rokoroku:
@aikoven We can simplify things. Writing types for every TOwnProps, TStateProps, TDispatchProps is really cumbersome. In this case, Partial\<T\> is very useful for reducing these tasks.

For example,

##### with given redux state
```ts
interface ReduxState {
  todos: TodoItem[];
};
```

##### suggested (with partial)
```ts

interface AppProps {
  todos: TodoItem[];
  addTodo: typeof TodoActions.addTodo;
};

interface AppState {
  /* something */
};

@connect<ReduxState, AppProps>(
  (state) => ({ todos: state.todos }),
  (dispatch) => ({ addTodo: bindActionCreators(TodoActions.addTodo, dispatch) })
)
export default class App extends React.Component<AppProps, AppState>{
  ...
}
```


##### current (without partial)
```ts
interface AppOwnProps {
  todos: TodoItem[];
}

interface AppStateProps {
  todos: TodoItem[];
}

interface AppDispatchProps {
  addTodo: typeof TodoActions.addTodo;
};

interface AppState {
  /* something */
};

@connect<ReduxState, AppOwnProps, AppStateProps, AppDispatchProps>(
  (state) => ({ todos: state.todos }),
  (dispatch) => ({ addTodo: bindActionCreators(TodoActions.addTodo, dispatch) })
)
export default class App extends React.Component<AppOwnProps & AppStateProps & AppDispatchProps, AppState>{
  ...
}
```

aikoven:
@rokoroku It's not really accurate.

```ts
interface AppProps {
  todos: TodoItem[];
  addTodo: typeof TodoActions.addTodo;
};

@connect<ReduxState, AppProps>(
  (state) => ({ todos: state.todos }),
  (dispatch) => ({ addTodo: bindActionCreators(TodoActions.addTodo, dispatch) })
)
export default class App extends React.Component<AppProps, AppState>{
  ...
}
```
In this case, props of connected component (the one returned from `connect(...)(Component)`) should have type `{}`, because both `todos` and `addTodo` are provided by `connect`. By using `Partial` we'd get `{todos?: ..., addTodo?: ...}` which is not correct.

Btw, TypeScript doesn't yet work well with decorators that change the type of decorated stuff, which `connect` does indeed.

rokoroku:
@aikoven However, we already set strict type here:  ```React.Component<MyProps, MyState>```
So the component will follow ``MyProps``, not ``Partial<MyProps>``.

Can't we add some abbreviated definitions while preserving existing definitions?
(And of course, I'm using the way you gave me :) )

rokoroku:
Ohh, I figured out your intentions.

I focused on the **inside of the component** (where our logic will be written), and you are likely  paying attention to the **``connect``-ed component**. (Am I right?)

Considering the result of ``connect``, you're right. But is it really necessary to export the decorated type (like ``dispatch`` in ``StateProps`` by default) and export it to the outside? I think the strength of the type definition is that it helps the developer to write the actual code, not the exact type.

But you guys don't need to take it seriously. I just wanted to talk about my thoughts :)

maclockard:
Is there any reason why connectAdvanced is not included in the typings?

theduke:
The current typings on DefinitelyTyped do not expose the connected components props on the connected component, breaking type checking when you have to pass some additional props.

Does anyone have a workaround for that?

Are the typings in this branch usable?

theduke:
@bbenezech thanks for the update, glad to hear.

aksharpatel47:
Hi @bbenezech: We are using react-redux with Typescript in one of our projects and would like to help in any way we can to get this merged. Let me know if you need any help.

timdorr:
Closing in favor of #815

PR #543, opened 2016-11-13
Add "^15.4.0-0" to compatible react peerDeps

zertosh:
In order to satisfy a prerelease version, the major/minor/patch versions must match exactly. This change makes `react-redux` `yarn` friendly when used with `react@15.4.0-rc.4`. Without this change, you always fail a `yarn check`.

cc: @gaearon @bestander.

### DIFF
--- a/package.json
+++ b/package.json
@@ -95,7 +95,7 @@
     "loose-envify": "^1.1.0"
   },
   "peerDependencies": {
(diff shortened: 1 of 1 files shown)


gaearon:
k

gaearon:
Should be in 4.4.6.

PR #546, opened 2016-11-21
Use React 15 in development

timdorr:
Noticed we were testing and building against React 0.14. This bumps us up to 15 for 5.0 so that we're testing against the current release version.


timdorr:
OK, passes tests. If there's any issue with using this version that I missed, please let me know!

jimbolla:
Doesn't this also bump the minimum React version to 15? I'd prefer we not do that, since we still use 14 on our single-sign-on project, because 15 broke IE 11 password autofill on controlled inputs, and it won't be fixed until React 16.

timdorr:
Nope, just the devDep.

jimbolla:
Oh I see now

PR #554, opened 2016-11-28
Shows more helpful error when connect() is passed invalid args.

jimbolla:
Instead of posting `initMapStateToProps is not a function` exception, show something more helpful like: `Invalid value of type string for mapStateToProps argument when connecting component AwesomeForm.`


timdorr:
Good stuff!

PR #555, opened 2016-11-28
Use isSubscribed method

AndrewHenderson:
`isSubscribed` method was unused. This makes use of that method internally to determine if the component should subscribe.


jimbolla:
`isSubscribed` is there for backwards compatibility reasons. Otherwise I would've just removed it completely.

timdorr:
I think this reads just a little bit better. And I would merge it in, but 5.0 is up next for release and this won't ever see the light of day as a result. Sorry about that!

PR #565, opened 2016-12-10
Eliminate addtional notifications to nested components

dk00:
When handling store updates, selector is sometimes called twice with same `state` and `ownProps` for nested components. I traced this part, and found things occurred in following order:

`parent: onStateChange, setState`
`nested: componentWillReceiveProps, selector, shouldComponentUpdate`
`parent: componentDidUpdate, notifyNestedSubs`
`nested: onStateChange, selector, (skip update), notifyNestedSubs`

The nested component was updated twice for one store update, by receiving (new?) `props`(`setState`), and being notified(`notifyNestedSubs`).

I tried to eliminate additional update attempts by removing `componentDidUpdate`, and run test to see how many cases would break...

> `63 passing (441ms)`

All test are passed, but it should not.

Nested components are *notified* by `setState` or `notifyNestedSubs` in `onStateChange`, or by React after receiving `props`. If React is told not to update after receiving `props`, all nested components should be notified by `notifyNestedSubs`, this is added to `componentWillReceiveProps`.

### DIFF
--- a/src/components/connectAdvanced.js
+++ b/src/components/connectAdvanced.js
@@ -200,11 +200,6 @@ export default function connectAdvanced(
             if (!this.selector.shouldComponentUpdate) {
               subscription.notifyNestedSubs()
             } else {
-              this.componentDidUpdate = function componentDidUpdate() {
-                this.componentDidUpdate = undefined
-                subscription.notifyNestedSubs()
-              }
-
               this.setState(dummyState)
             }
           }.bind(this)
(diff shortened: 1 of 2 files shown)


jimbolla:
This doesn't look right. Moving notifyNestedSub into compenentWillReceiveProps will cause it to not get called if there's a blocker component preventing it from receiving new props.

gaearon:
>This doesn't look right. Moving notifyNestedSub into compenentWillReceiveProps will cause it to not get called if there's a blocker component preventing it from receiving new props.

Is it worth adding a failing test for this?

jimbolla:
I just added a test as PR #566 that would fail if this PR is applied.

PR #576, opened 2016-12-14
Also attempt to exit early when iterating b's keys in shallowEqual

jcready:
[Prior conversation](https://github.com/reactjs/react-redux/pull/416#pullrequestreview-12930768)

### DIFF
--- a/src/utils/shallowEqual.js
+++ b/src/utils/shallowEqual.js
@@ -12,8 +12,8 @@ export default function shallowEqual(a, b) {
   }
 
   for (let key in b) {
-    if (hasOwn.call(b, key)) countB++
+    if (hasOwn.call(b, key) && countA < ++countB) return false
   }
 
-  return countA === countB
+  return true
 }


jimbolla:
Can you run some perf benchmarks just to be sure it's faster? You know that old saying about what happens when you ASSUME.

appden:
I'm more concerned of the correctness of `shallowEqual` in `react-redux`. There are a few subtle ways it could fail right now. I think it would be best to copy [`shallowEqual` from `fbjs`](https://github.com/facebook/fbjs/blob/master/packages/fbjs/src/core/shallowEqual.js).

jcready:
I agree with @appden here. For instance `shallowEqual` in react-redux doesn't account for `NaN`. If an object contained any `NaN` value then `shallowEqual` would always return `false`.

timdorr:
I'm going to close this out and we can switch over to the fbjs version. It seems to cover more edge cases and might even be faster in common cases for react-redux.

Issue #588, opened 2016-12-27
setState warning using v5.0.1 with redux-form


osigum:
Similar to the issue  #564 .

Just like in the above mentioned issue, I have a hidable form. The difference is: I have a ParentContainer that passes the **visible** prop to the ConnectedParent. Yet everything is fine unless I hide the Form dispatching asynchronously. In this case I get a warning: 

> setState(...): Can only update a mounted or mounting component. This usually means you called setState() on an unmounted component. This is a no-op. Please check the code for the Connect(Form(Form)) component.

I'm sorry, but i can't provide a clonable example. Some code to reproduce this bug:
```js
import React, { Component } from 'react';
import {connect} from 'react-redux';
import {Field, reduxForm} from 'redux-form';

const Form = () => <Field type="text" component="input" name="field" />;
const ReduxForm = reduxForm({ form: 'example' })(Form);


class Parent  extends Component {
  componentWillReceiveProps(newProps) {
    if (this.props.visible !== newProps.visible) {
      if (!newProps.visible) {
        newProps.dispatch({ type: 'ANYTHING' });
      }
    }
  }

  render() {
    return <div>{this.props.visible ? <ReduxForm /> : null}</div>;
  }
}
const ConnectedParent = connect()(Parent);


const ParentContainer = ({ dispatch, visible }) => (
  <div>
    <ConnectedParent visible={visible} />
    <button type="button" onClick={() => { dispatch({ type: 'SHOW_FORM' });} }>show</button>
    <button type="button" onClick={() => { dispatch({ type: 'HIDE_FORM' });} }>hide</button>
    <button type="button" onClick={() => { setTimeout(() => { dispatch({ type: 'HIDE_FORM' }); }, 0)} }>hide with timeout</button>
  </div>
);

const mapStateToProps = state => ({ visible: state.app.form });
const ConnectedParentContainer = connect(mapStateToProps)(ParentContainer);

export default ConnectedParentContainer ;
```

Corresponding reducer:
```js
const initialState = {
  form: false,
};

export default (state = initialState, action) => {
  switch (action.type) {
    case 'SHOW_FORM':
      return Object.assign({}, state, {
        form: true,
      });
    case 'HIDE_FORM':
      return Object.assign({}, state, {
        form: false,
      });
    default:
      return state;
  }
};
```
Use the ConnectedParentContainer component and the reducer in your app.

Steps to reproduce:
1) Click "show" button.
2) Click "hide with timeout" button.
[...]

PR #589, opened 2016-12-28
Fixes subscription bug when a new store is passed as a prop.

jimbolla:
Reported here: https://github.com/davidkpiano/react-redux-form/issues/592

### DIFF
--- a/src/components/connectAdvanced.js
+++ b/src/components/connectAdvanced.js
@@ -6,6 +6,30 @@ import Subscription from '../utils/Subscription'
 import storeShape from '../utils/storeShape'
 
 let hotReloadingVersion = 0
+const dummyState = {}
+function noop() {}
+
+function makeSelectorStateful(sourceSelector, store) {
+  // wrap the selector in an object that tracks its results between runs.
+  const selector = {
+    run: function runComponentSelector(props) {
+      try {
+        const nextProps = sourceSelector(store.getState(), props)
+        if (nextProps !== selector.props || selector.error) {
+          selector.shouldComponentUpdate = true
+          selector.props = nextProps
+          selector.error = null
+        }
+      } catch (error) {
+        selector.shouldComponentUpdate = true
+        selector.error = error
+      }
+    }
+  }
+
+  return selector
+}
+
 export default function connectAdvanced(
   /*
     selectorFactory is a func that is responsible for returning the selector function used to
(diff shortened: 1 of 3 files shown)


davidkpiano:
Thanks @jimbolla  - the general idea is that this should work:

```jsx
// overly simplified for clarity

<Provider store={store}>
  <Foo>
    <Bar store={ownStore} />
  </Foo>
</Provider>
```

If I'm not mistaken, `<Bar>` should subscribe to its `ownStore` as if it were its own `<Provider>` (because it essentially is) instead of depending on parent subscriptions.

jimbolla:
Correct. I think there's an edge case to this edge case I'm not covering so I'm gonna keep working on it, but this should be better.

jimbolla:
OK. Now I think I covered the edge cases correctly.

jimbolla:
@timdorr Thoughts on pushing a patch release for this and the other handful of fixes we have waiting to go out?

jimbolla:
I added an extra safeguard in here as well that I'm pretty sure will also fix #588 but I don't have an exact repro for that one, so I'm approximating.

davidkpiano:
Any updates on this? Will it be able to be merged in soon? 🙏

markerikson:
@jimbolla , @timdorr : following up on this at @davidkpiano 's request.  It's Jim's code thus far, and Tim's been cutting releases (something I should go figure out for myself at some point, but don't have time to dig into atm).  Are we good to go on this?

timdorr:
I haven't had time to really dig into this one. It's a non-trivial change, hence the hesitation to just blindly merge it in. If anyone else wants to review, I trust them to point out any problems.

markerikson:
Yeah, that's actually how I feel about this myself at the moment :)

jimbolla:
@timdorr @markerikson It might be easier to review the 2 commits separately. The bugfix one is pretty small and the refactor one just moves a few things and adds some more comments.

timdorr:
OK, LGTM. In the future, I'd submit refactors as either separate PRs or just directly onto master.

anru:
@timdorr please release this fix

BerndWessels:
@timdorr The `react-redux 5.0.3` update breaks HMR with latest `react-hot-loader 3.0.0-beta.6`.

Basically after a hot-reload the state is somehow broken.

This worked fine in `react-redux 5.0.2` :(

PR #591, opened 2016-12-29
Use shallowEqual from fbjs

### LINKED ISSUE #590: shallowEqual always return false if add properties to Object.prototype
```javascript
// copy from src/utils/shallowEqual.js
const hasOwn = Object.prototype.hasOwnProperty

function shallowEqual(a, b) {
  if (a === b) return true

  let countA = 0
  let countB = 0
  
  for (let key in a) {
    if (hasOwn.call(a, key) && a[key] !== b[key]) return false
    countA++  // should check hasOwn.call(a, key) === true and then countA++ ?
  }

  for (let key in b) {
    if (hasOwn.call(b, key)) countB++
  }

  return countA === countB
}

// my example
Object.prototype.bar = 'bar'

let a = {x: '1'}
let b = {x: '1'}

shallowEqual(a, b) // return false
```

it's a misunderstanding or there are some reasons for this?

timdorr:
Fixes #590

Also handles things like NaN.

### DIFF
--- a/src/utils/shallowEqual.js
+++ b/src/utils/shallowEqual.js
@@ -1,19 +1,32 @@
 const hasOwn = Object.prototype.hasOwnProperty
 
-export default function shallowEqual(a, b) {
-  if (a === b) return true
+function is(x, y) {
+  if (x === y) {
+    return x !== 0 || y !== 0 || 1 / x === 1 / y
+  } else {
+    return x !== x && y !== y
+  }
+}
 
-  let countA = 0
-  let countB = 0
-  
-  for (let key in a) {
-    if (hasOwn.call(a, key) && a[key] !== b[key]) return false
-    countA++
+export default function shallowEqual(objA, objB) {
+  if (is(objA, objB)) return true
+
+  if (typeof objA !== 'object' || objA === null ||
+      typeof objB !== 'object' || objB === null) {
+    return false
   }
 
-  for (let key in b) {
-    if (hasOwn.call(b, key)) countB++
+  const keysA = Object.keys(objA)
+  const keysB = Object.keys(objB)
+
+  if (keysA.length !== keysB.length) return false
+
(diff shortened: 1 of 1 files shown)


jimbolla:
We should perf test this, since it's called so much. It's useful to remember that we're not comparing any 2 rando values, but specifically props objects coming from react. I think we can make certain assumptions, like that they're objects, and typically (99% of time?) that the # of props is going to be the same between the 2 args. So the algorithm can optimize for things like that.

timdorr:
True, but I think the effects are minor. Given this is a hot path, it's going to get optimized pretty heavily. I think the only thing to do is to use the for in loop form instead, as v8 [has specific optimizations for that with a `hasOwnProperty` check](https://github.com/v8/v8/blob/master/src/runtime/runtime-forin.cc).

jimbolla:
If the perf of the fbjs version is fine, then maybe we should just import it from fbjs. fbjs is a dependency of react, so react-redux has a dependency on it indirectly already.

timdorr:
AFAIK, `shallowEqual` hasn't changed in quite some time in fbjs. I'm going to merge this as-is and we can look at switching it to an import at a later date.

jimbolla:
Works for me. This should be fast enough for normal usage. And if someone really needs a faster implementation than the fbjs, they have the power to override it in the options.

gaearon:
I wouldn't recommend depending on anything from `fbjs` directly.

gaearon:
(Copying is fine 😄 )

wtgtybhertgeghgtwtg:
This seems like it happens often; `recompose` just has `fbjs` as a dependency for shallow equal, and I think `react-side-effect` used to do that.  Do you think `fbjs` could move their shallow equal implementation into another library?  Copying doesn't seem like the best solution, either.

PR #600, opened 2017-01-16
possible syntax clarification

abglassford:
I attempted to clarify some pronouns in a couple sentences to avoid confusion when sentence has multiple objects. Please let me know if I'm confusing some things.


markerikson:
Hmm.  Appreciate the PR, but yeah, I think the new phrasing is incorrect.  The idea is that when the _store_ updates, the wrapper component (which is a subscriber to the store) will be notified.  So, it's not that the _component_ updates, it's the _store_.  

To be honest, I think both paragraphs could stand to be rewritten from scratch.  I'll have to add that to my todo list.

If you can fix up your edits to clarify that it's the store updates that lead to `mapState` being re-run, I'll get this in.

abglassford:
That seemed to be an easy fix. Unless I missed your point. Let me know if it's decent now.

markerikson:
Yeah, actually, let's make a couple more edits:

- Change "If a component is specified" to "if this argument is specified"
- Change "if you omit the component" to something more like "to skip subscribing to store updates, pass `null` or `undefined` for this argument (this applies for both `mapState` and `mapDispatch`)

abglassford:
So, when the argument `mapStateToProps` is specified, does it subscribe a _component_ to store updates or it will subscribe _itself_ to store updates? I just want to be clear on what is actually subscribing to the store.

markerikson:
`connect` generates a wrapper component.  That wrapper component will subscribe to the store, and when the store runs the subscriber callbacks, the wrapper component will call `mapState(store.getState())`.

Dan wrote a miniature version of `connect` that illustrates the basic idea: https://gist.github.com/gaearon/1d19088790e70ac32ea636c025ba424e

abglassford:
Gotcha. I'm going to leave it at this commit if it's sufficient. Thanks for taking the time to look at this and explain it to me!

markerikson:
One more phrasing tweak, if you could.  Let's change "it will subscribe" to "the new component will subscribe".

markerikson:
Awright, looks good.  Thanks for the help!  If you've got any other suggestions or improvements for either the Redux or React-Redux docs, please feel free to file PRs and we'll see what we can do.

abglassford:
Thank you!

PR #603, opened 2017-01-19
fix(connect): should not call update if props has not changed

ipluser:
- I dispatch a action, If the props value has not changed but it return new props.
- Next, I dispatch a action that not match, it return the prev props

the **react-redux@4.x** will call update to compute props, see the below example:

The reducer like this:

```js
const reducer = (prev = {}, action) => {
  return action.type === 'NOT_CHANGED'
    ? { ...prev, ...action.payload }
    : prev
}
```

And do these action:

```js
// cuz the props reference has changed, it will call update but the props value has not changed
store.dispatch({ type: 'NOT_CHANGED', payload: {} })

// then I dispatch the action that not match, the props has not changed, it will call update again
store.dispatch({ type: 'NOT_MATCH' })
```

Find source code **handleChange**:

```js
handleChange() {
  if (!this.unsubscribe) {
    return
  }

  const storeState = this.store.getState()
  const prevStoreState = this.state.storeState
  if (pure && prevStoreState === storeState) {
    return
  }

  if (pure && !this.doStatePropsDependOnOwnProps) {
    const haveStatePropsChanged = tryCatch(this.updateStatePropsIfNeeded, this)
    // if props has not changed and return
    // but the prev storeState has not updated, it will cause the next storeState check is wrong
    if (!haveStatePropsChanged) {
      return
    }
    if (haveStatePropsChanged === errorObject) {
      this.statePropsPrecalculationError = errorObject.value
    }
    this.haveStatePropsBeenPrecalculated = true
  }

  this.hasStoreStateChanged = true
  this.setState({ storeState })
}
```

I do this case with **react-redux@5.x**, this problem has not exist.

### DIFF
--- a/src/components/connect.js
+++ b/src/components/connect.js
@@ -90,6 +90,7 @@ export default function connect(mapStateToProps, mapDispatchToProps, mergeProps,
         )
 
         const storeState = this.store.getState()
+        this.prevStoreState = storeState
         this.state = { storeState }
         this.clearCache()
       }
@@ -242,10 +243,11 @@ export default function connect(mapStateToProps, mapDispatchToProps, mergeProps,
         }
 
         const storeState = this.store.getState()
-        const prevStoreState = this.state.storeState
-        if (pure && prevStoreState === storeState) {
+        if (pure && this.prevStoreState === storeState) {
           return
         }
+
+        this.prevStoreState = storeState
 
         if (pure && !this.doStatePropsDependOnOwnProps) {
           const haveStatePropsChanged = tryCatch(this.updateStatePropsIfNeeded, this)
--- a/test/components/connect.spec.js
+++ b/test/components/connect.spec.js
@@ -1789,6 +1789,50 @@ describe('React', () => {
       expect(renderCalls).toBe(1)
     })
 
+    it('should not call update if props has not changed', () => {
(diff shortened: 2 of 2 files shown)


jimbolla:
Curious... is there something preventing you from upgrading to v5?

timdorr:
Sounds like he's saying this doesn't exist in 5.0, which is true.

@ipluser I appreciate the effort here, but 4.x is in bugfix-only mode. If you need this kind of behavior, switch to 5.0, which shouldn't involve any code changes on your end.

ipluser:
@jimbolla  @timdorr 

As you say, our framework depend on **react-redux@4.x**, we plan to upgrading **react-redux** to v5 and increment framework **MAJOR** version, not MINOR.

e.g.

```
framework@1.x => framework@2.0
```

I'm happy to see **react-redux@5.x**, it's nice for **compatibility**, **performance improvements** and others features.
But sometime, something isn't controlled by me, I have to think about that the framework must be stable, safe and other.

I can't promise anything is ok if I upgrading **react-redux** to v5 and increment framework **MINOR** number.
It may bring some hidden problems, it has some risk. 

So I want to fix this issue for currently framework with **react-redux@4.x**.
And next, I will plan to upgrading **react-redux** to v5 and increment framework **MAJOR** version.

PR #605, opened 2017-01-22
Stringify wrong argument for easier debugging

pvorona:
Produce more meaningfull message in case of receiving not a function.


markerikson:
Seems reasonable.  Thanks!

PR #615, opened 2017-02-03
Update `mapDispatchToProps` API documentation

rmadsen:
I believe the documentation mistakenly references `mapDispatchToProps` when it means `mapStateToProps`. Users should pass `null` as the first argument (i.e. to `mapStateToProps` if they don't want to subscribe to store state), but still can get `dispatch` injected using `mapDispatchToProps`.  This commit fixes this reference.


markerikson:
Yeah, I think you're right.  Thanks!

naw:
Probably that whole sentence should be removed from the paragraph ---- it looks like it was just pasted from the paragraph describing `mapStateToProps` and inadvertently not removed.

PR #616, opened 2017-02-04
Allow 2nd method to get ownProps if factory doesn't require it. Fixes #604

### LINKED ISSUE #604: when mapStateToProps is a factory and has one, the new method does not get ownProps
```
mapStateToProps(initialState) {
    return (state, ownProps) => {
         // ownProps is undefined
    };
}
```

It works fine with no arguments or 2 arguments..

lukeapage:


jimbolla:
Can you also add a test that covers this? [This test](https://github.com/reactjs/react-redux/blob/master/test/components/connect.spec.js#L1842) is probably a good starting point.

lukeapage:
@jimbolla no problem, test added

Issue #617, opened 2017-02-04
Combine components the same way you combine reducers


simlrh:
Many reducers can be combined into one while remaining responsible for their own section of the state tree. They aren't required to know about the whole app, only their own state changes.

Components receive the whole state tree in mapStateToProps, so they need to know their position within the app as a whole to find the state relevant to them. It would aid reusability if components could be composed so that they only receive the state relevant to them.

(PR incoming)

timdorr:
Let's keep the discussion in #618

PR #618, opened 2017-02-04
Combine components the same way you combine reducers

simlrh:
Implements #617 

This PR allows you to combine `connect()`ed components in a similar way to `combineReducers()`, so that their `mapStateToProps` function only receives the section of the state tree relevant to them rather than the whole thing. 

This aids reusability as components do not need to know where in the state tree they appear. It should also help performance as mapStateToProps will only be called when the subtree is changed, not the global state.

This is achieved with a `SubProvider` component which provides a proxy store in its child context. The proxy store passes `dispatch()` and `subscribe()` through to the real store but intercepts `getState()` and only returns the property of the state object specified by the `SubProvider`'s `subState` prop.

You don't need to use SubProvider directly as there is also `combineConnected()` which takes an object containing `connnect()`ed components and returns them wrapped in SubProviders, eg:

```js
let { child1: Child1, child2: Child2 } = combineConnected({
  child1: Component1,
  child2: Component2
});

<Provider store={store}>
  <div>
    <Child1 />
    <Child2 />
  </div>
</Provider>
```

`mapStateToProps` on `Child1` will only receive `state.child1`, and `mapStateToProps` on `Child2` will only receive `state.child2`.

See example:
https://gist.github.com/simlrh/2212d5e2ae274839249618af5537dcfb

Documentation and tests have been added.

I know I didn't wait to find out if you actually want this feature, but I was going to do it anyway for my own education and use :)

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -2,7 +2,9 @@
 
 - [Redux: Usage with React](http://redux.js.org/docs/basics/UsageWithReact.html)
 - [API](api.md#api)
-  - [`<Provider store>`](api.md#provider-store)
-  - [`connect([mapStateToProps], [mapDispatchToProps], [mergeProps], [options])`](api.md#connectmapstatetoprops-mapdispatchtoprops-mergeprops-options)
-  - [`connectAdvanced(selectorFactory, [connectOptions])`](api.md#connectadvancedselectorfactory-connectoptions)
+  - [`<Provider store>`](api.md#provider)
+  - [`<SubProvider subState>`](api.md#subprovider)
+  - [`connect([mapStateToProps], [mapDispatchToProps], [mergeProps], [options])`](api.md#connect)
+  - [`connectAdvanced(selectorFactory, [connectOptions])`](api.md#connectadvanced)
+  - [`combineConnected(components)`](api.md#combineConnected)
 - [Troubleshooting](troubleshooting.md#troubleshooting)
--- a/docs/api.md
+++ b/docs/api.md
@@ -41,6 +41,33 @@ ReactDOM.render(
 )
 ```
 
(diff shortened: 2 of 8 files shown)


gaearon:
Maybe this could be a third-party library first, and then we can consider including it by default if a lot of people start using it?

simlrh:
Sounds good to me, I'll roll it into a library as soon as I can think of a pithy name

markerikson:
Yeah, for what it's worth, I've seen several existing third-party libraries that implement this kind of behavior.  As usual, I haven't tried any of them myself yet, but I've added all the ones I've seen to the [Component State and Encapsulation](https://github.com/markerikson/redux-ecosystem-links/blob/master/component-state.md) portion of my Redux addons catalog.  (Which is sadly in need of some additional organizing work, but that's very low on my priority list right now.)

It's certainly not a _bad_ idea in and of itself, I'm just not sure it needs to go directly in the core React-Redux library.  There's a strong parallel to how the core Redux library provides the bare basic `subscribe()` and `getState()` primitives, and that provides the building blocks for the community to build addons for specific use cases on top of those primitives.

I'll leave this open for discussion, but at least at the moment I suspect it probably won't get merged in.  Would definitely love to see this as a third-party lib, though.

simlrh:
I published on npm this afternoon: https://www.npmjs.com/package/react-redux-combine-components and https://github.com/simlrh/react-redux-combine-components (so much for pithy).

If I'd seen that list earlier I might not have bothered, but I have to say I like this solution :wink:

I'm happy whenever you want to close the PR. Thanks for your time!

markerikson:
Heh, sure.  I actually just went ahead and added your library to that "Component State" list, too.

Thanks for the PR!

PR #624, opened 2017-02-17
Fix the problem introduced by previous refactor in `shallowEqual`

jddxf:
Previous refactor `shallowEqual` has introduced an issue as follow
```
function A() {}
A.prototype.a = 1
const a = new A()
const b = { a: 2 }
shallowEqual(a, b) // returns true, expects false
```


timdorr:
This will be handled by #591. Thanks!

PR #625, opened 2017-02-17
Remove shouldComponentUpdate

### LINKED ISSUE #507: unimplement shouldComponentUpdate?
There was some [good discussion](https://twitter.com/ryanflorence/status/779320581678174208) on Twitter (between @gaearon, @markerikson, @timdorr, @mjackson, @ryanflorence, et al.) about `connect`'s component implementing `shouldComponentUpdate`.

If I understand correctly, the main issue being that any component wrapped in `connect` then blocks updates related to context changes, and this has an adverse impact on libraries that pass data via context that need to trigger rerenders (such as React Router's Link component.)

The changes in the `next` branch of React Redux make its implementation of `shouldComponentUpdate` much less necessary, mainly because now `setState` is no longer called as a response to every store state change, but only if the final merged props have changed. So now when `shouldComponentUpdate` is called as a result of calling `setState`, it's always going to return true anyways. (The call to setState could probably be replaced with forceUpdate and would work exactly the same.)

In the case of `shouldComponentUpdate` being called after receiving new props from parent, it's effectively just acting like `PureComponent`. That responsibility can be given to the components being wrapped, which would have better knowledge about if/how they should implement `shouldComponentUpdate`. I personally would use recompose and do something like:

``` js
export default compose(
   pure, // <-- from recompose
   connect(...),
)(MyComponent);
```

An alternative to removing connect's `shouldComponentUpdate` completely would be to make it an another `option` argument, and decide whether it should be opt-in or opt-out.

timdorr:
Fixes #507 and like every context subscription-related bug ever. 

This is just a spike of me editing on the web interface (not on a machine with Node installed...). Travis builds will fail, things will break, no one will be happy. But it gets the ball rolling on removing it, so it's a first step.

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -40,28 +40,6 @@ render() {
 Conveniently, this gives your components access to the router state!
 You can also upgrade to React Router 1.0 which shouldn’t have this problem. (Let us know if it does!)
 
-### My views aren’t updating when something changes outside of Redux
-
-If your views depend on global state or [React “context”](http://facebook.github.io/react/docs/context.html), you might find that views decorated with `connect()` will fail to update.
-
->This is because `connect()` implements [shouldComponentUpdate](https://facebook.github.io/react/docs/component-specs.html#updating-shouldcomponentupdate) by default, assuming that your component will produce the same results given the same props and state. This is a similar concept to React’s [PureRenderMixin](https://facebook.github.io/react/docs/pure-render-mixin.html).
-
-The _best_ solution to this is to make sure that your components are pure and pass any external state to them via props. This will ensure that your views do not re-render unless they actually need to re-render and will greatly speed up your application.
-
-If that’s not practical for whatever reason (for example, if you’re using a library that depends heavily on React context), you may pass the `pure: false` option to `connect()`:
-
-```
-function mapStateToProps(state) {
-  return { todos: state.todos }
-}
-
-export default connect(mapStateToProps, null, null, {
-  pure: false
-})(TodoApp)
-```
-
-This will remove the assumption that `TodoApp` is pure and cause it to update whenever its parent component renders. Note that this will make your application less performant, so only do this if you have no other option.
-
 ### Could not find "store" in either the context or props
 
 If you have context issues,
--- a/src/components/connectAdvanced.js
+++ b/src/components/connectAdvanced.js
@@ -161,10 +161,6 @@ export default function connectAdvanced(
         this.selector.run(nextProps)
       }
 
-      shouldComponentUpdate() {
-        return this.selector.shouldComponentUpdate
-      }
-
       componentWillUnmount() {
         if (this.subscription) this.subscription.tryUnsubscribe()
         this.subscription = null
--- a/src/components/connectAdvanced.js
+++ b/src/components/connectAdvanced.js
@@ -190,16 +190,16 @@ export default function connectAdvanced(
 
       initSubscription() {
         if (!shouldHandleStateChanges) return
-        
+
         // parentSub's source should match where store came from: props vs. context. A component
         // connected to the store via props shouldn't use subscription from context, or vice versa.
         const parentSub = (this.propsMode ? this.props : this.context)[subscriptionKey]
         this.subscription = new Subscription(this.store, parentSub, this.onStateChange.bind(this))
-        
+
         // `notifyNestedSubs` is duplicated to handle the case where the component is  unmounted in
         // the middle of the notification loop, where `this.subscription` will then be null. An
         // extra null check every change can be avoided by copying the method onto `this` and then
-        // replacing it with a no-op on unmount. This can probably be avoided if Subscription's 
+        // replacing it with a no-op on unmount. This can probably be avoided if Subscription's
         // listeners logic is changed to not call listeners that have been unsubscribed in the
         // middle of the notification loop.
         this.notifyNestedSubs = this.subscription.notifyNestedSubs.bind(this.subscription)
@@ -214,7 +214,7 @@ export default function connectAdvanced(
           this.componentDidUpdate = this.notifyNestedSubsOnComponentDidUpdate
           this.setState(dummyState)
         }
-      }      
+      }
 
       notifyNestedSubsOnComponentDidUpdate() {
         // `componentDidUpdate` is conditionally implemented when `onStateChange` determines it
--- a/src/connect/selectorFactory.js
+++ b/src/connect/selectorFactory.js
@@ -1,5 +1,5 @@
 import verifySubselectors from './verifySubselectors'
-  
+
 export function impureFinalPropsSelectorFactory(
   mapStateToProps,
   mapDispatchToProps,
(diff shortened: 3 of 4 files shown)


markerikson:
Whoa.  That is some _seriously_ advanced hacker-fu there.  _bows down_

So just to confirm my own understanding: now that we're doing everything with layers of memoized selectors, `sCU` is basically pointless because we're doing the same kind of work internally, and keeping the wrapped component from re-rendering unless something actually changed?

jimbolla:
@markerikson Yeah. I think https://github.com/reactjs/react-redux/issues/507#issue-179829234 explains why sCU isn't a requirement anymore.

gaearon:
>In the case of shouldComponentUpdate being called after receiving new props from parent, it's effectively just acting like PureComponent. That responsibility can be given to the components being wrapped, which would have better knowledge about if/how they should implement shouldComponentUpdate. I personally would use recompose and do something like:

That would be a breaking change, right? Many apps are likely relying on this behavior—if not for functionality then for performance. I agree it might be better not to impose purity on everyone, but people are already treating `connect()` as a way to give purity, and breaking that promise should only be done in a major IMO.

markerikson:
Yeah, it would probably be worth a major bump.  I guess what I'm trying to nail down is: does the memoized internals approach give the _effect_ of `sCU`, without actually needing to have it implemented at the React level?  if it's removed and the wrapper component is asked to re-render with the same values, what happens?  Or is `sCU` still necessary to keep things from rippling downwards?

gaearon:
I can’t answer this because I don’t know how the new implementation works.
I’d be happy to answer any questions about React (or the old implementation) though!

markerikson:
Yeah, that was really meant for @jimbolla .

Looking at that deleted couple lines, I'm pretty sure `sCU` is still necessary for preventing re-renders from parents, and that without it, the wrapper is really only able to deal with changes from `map*`-related values.

jimbolla:
This would certainly be a breaking change. sCU is no longer needed for changes coming from store, but still affects changes coming from props. People, including myself, rely on `connect()` as a convenient way to get the same benefits as PureComponent.

I'm feeling like this should wait until "React 16 - Now With FIber™" lands so that we can see how it affects performance. It's possible that R16 could change the way we get maximum performance and we end up wanting to keep sCU.

gaearon:
I would add that after React 16 we'll likely make Fiber async by default (you can flip the switch [here](https://github.com/facebook/react/blob/bc2702f8bd89737898874e05adc67a4b4f542929/src/renderers/dom/fiber/ReactDOMFiber.js#L326) if you build it yourself). As a result, `setState`s would be batched in the correct order by default, and I’m not sure the complexity of notifying the listeners in the right order will be as important. Also, `setState` callbacks (which, as I understand, we’re using to ensure the correct order) have performance implications in Fiber and prevent some optimizations so it would be nice to not rely on them.

markerikson:
Ah... as far as I know, we _don't_ use `setState` for ordering.  Jim wrote a nested subscription system - children effectively subscribe to their nearest connected ancestor if one exists, so that they're only notified after the ancestor is notified.

gaearon:
Heads up: React 16 might not need this but it is possible React will rely *more* on sCU in the future in order to better reuse work in async scenarios. The details are still very fuzzy (we just started looking into trying to use async rendering at FB) but I wouldn’t hurry with this change in case we need to backtrack later. The context API is another beast because that’ll need a redesign anyway (likely in 17).

timdorr:
OK, finally updated this damn thing.

I removed some tests related to f7dc41d9, which was some of the original attempts at optimizations. This raises a slightly modified question: Should we be responsible for handling sCU? 

The whole point was to be automatically aware of state changes that don't require updates and stop them with sCU. The unfortunate side effect was this required us to handle ownProps updates as well and we ended up applying our update-blocking logic to those props. 

But I think this may be a mistake. We're making assumptions about how we want connected components to update, while lacking actual knowledge of the business meaning of ownProps. I think implementing sCU should be left up to the user. Very commonly, they end up getting lazy and not ever doing it (I know I'm guilty of this!), even in non-`connect`ed components. We're basically saying to skip the optimization in development because React Redux will take care of it for you. But meanwhile, we don't know enough to make an intelligent decision about whether or not to update. Shallow comparisons may be insufficient, deep comparisons may negate the performance benefit. We just don't know.

So, this may involve more than just fixing some code. We would have to be sure to educate people about this. And it may be a little painful, but I think it's worthwhile to get people thinking about writing their own sCU's.

gaearon:
I am very concerned about changing something like this, especially after we already have a report about 5.x significantly regressing on performance. https://github.com/reactjs/react-redux/issues/686

gaearon:
Basically, the way I think about it that Redux plainly wouldn't work on larger apps unless you aggressively optimize sCU. It makes sense to me to do this by default, as this is a more opinionated way to use React. Just like Om.

Pensarfeo:
Does this really solve the problem? or simply shifts the responsibility to the users, like me :)
I think the issue of removing sCU goes into a more fundamental questions: what is the purpose of this library? I personally also use this library to control the update of my components according to the origin of data being passed to my components: either data from the store or props from the parent component.
If you remove the sCU we would not only push us to refactor lots and lots of code, but it would also make it very annoying to do so since we wont have a way to easily differentiate where are the props coming from.
Why not adding an option to update the component on context change too? Or using a library like the one described by markerikson above?

gaearon:
Also, I don’t think this will fix context bugs in practice. IMO you *need* to use sCU in a Redux app (otherwise single store is impractical), and as soon as you implement your first sCU, context is broken again.

wtgtybhertgeghgtwtg:
> I'm feeling like this should wait until "React 16 - Now With FIber™" lands so that we can see how it affects performance. It's possible that R16 could change the way we get maximum performance and we end up wanting to keep sCU.

So, now that `react@16` has landed, will this get another look?

markerikson:
@wtgtybhertgeghgtwtg : maybe.  Dan and Andrew have both said that the async React behavior they're currently prototyping and experimenting with will likely require changes in how other libs interact with React.  Andrew has also said that at some point he'd like to have a discussion with the various state management lib maintainers on this topic.

React 16.0 deliberately maintains as much backwards compatibility with 15.x as possible.  Later iterations of 16 may start changing that.  

So, it's still too early to know what sorts of changes React itself is going to have and how those will affect the ecosystem, and therefore too early to know what's going to need to change with React-Redux.

gaearon:
>So, it's still too early to know what sorts of changes React itself is going to have and how those will affect the ecosystem, and therefore too early to know what's going to need to change with React-Redux.

👍 

Let's hold on on any big changes until next summer. I think a lot will be clearer by then.

tyscorp:
Will the new context API (facebook/react#11818) mean that this change will no longer be neccessary?

timdorr:
@tyscorp Only for libraries blocked by `connect()` that use it. It doesn't automatically apply to older libraries.

timdorr:
Closing this out in favor of moving to the new context API.

PR #626, opened 2017-02-18
Feature/merge props factory

neeharv:
This replaces #523. The older branch was forked from the `next` branch ages ago, and there were too many junk commits. I'm opening this against a new branch with just the cherry picked commits.


neeharv:
Any update on if this is on the roadmap to be merged in? If not, will close it so there's less noise. Thanks!

markerikson:
Can you recap what the actual intended use case is here, and the expected benefits?  We've already got a bunch of overloads for `connect`, and we generally view `mergeProps` as an escape hatch (and mostly discourage people from using it).  I'm hesitant to add more complexity around this.

mattyork:
@markerikson 

Use case: an action needs part of the State tree, but that part of the State tree is only used in the action and is not used in the React components.

#### Current workaround 1: pipe the State through the component anyway

This is what most people do today.

```js
const ContainerComp = connect(
 state => ({unusedInComp: state.a, usedInComp: state.b}),
 dispatch => ({ 
   fireAction: (unusedInComp, usedInComp) => dispatch(createAction(unusedInComponent, usedInComp)) 
 })
)(PresentationalComp);
```

downside: the component needs to worry about app state it shouldn't need to worry about. This becomes a problem when working with large teams and complex apps, where we need proper encapsulation for safety and to isolate complexity. Leaking the unnecessary state into the components confuses things. 

We can mitigate this by enforcing that every Container wraps a "StateBinder" component whose only purpose is to bind the unused State props to dispatch props using memoized function factories, then forwarding those down to PresentationalComponent underneath, but that is a messy way of accomplishing the goal. 

```js
const ContainerComp = connect(
 state => ({ unusedInComp: state.a, usedInComp: state.b }),
 dispatch => ({ 
   fireAction: (unusedInComp, usedInComp) => dispatch(createAction(unusedInComponent, usedInComp)) 
 })
)(StateBinderComp);

const memoizedCreateFireAction = (fireAction, unusedInComp) => {
  return usedInComp => dispatch(createAction(unusedInComponent, usedInComp)) 
}
const StateBinderComp = ({ unusedInComp, usedInComp, fireAction }) => (
  <PresentationalComp
    usedInComp={usedInComp}
    fireAction={memoizedCreateFireAction(fireAction, unusedInComp)}
  />
)
```

#### Current workaround 2: use mergeProps

downside: because mergeProps does not take a factory function like mapStateToProps does (hence this issue), the `fireAction` prop is a new lambda every time, preventing `shouldComponentUpdate` optimizations for PureRender components. We could work around that by writing our own `shouldComponentUpdate` function in the child component, but that's error prone and a bit messy.

```js
const ContainerComp = connect(
 state => ({unusedInComp: state.a, usedInComp: state.b}),
 dispatch => ({ 
   fireAction: (unusedInComp, usedInComp) => dispatch(createAction(unusedInComponent, usedInComp)) 
 }),
 (stateProps, dispatchProps) => ({ 
   usedInComp: stateProps.usedInComp,
   presentationalFireAction: usedInComp => dispatch(createAction(stateProps.unusedInComponent, usedInComp)) 
 })
)(PresentationalComp);
```

#### Current workaround 3: use connectAdvanced

```js
const ContainerComponent = connectAdvanced(dispatch => {
  const memoizedCreateFireAction = memoize(unusedInComp => {
    return usedInComp => dispatch(createAction(unusedInComp, usedInComp));
  }
  return state => ({
    usedInComp: state.usedInComp,
    presentationalFireAction: memoizedCreateFireAction(state.unusedInComp)
  });
})
```

downside: this does not have the other features of the normal `connect()`, like deciding whether or not to subscribe to the Store. It also does not have built in memoization for the Container like `connect()` does, so this will only be pure if the PresentationalComp is a PureComponent, which is not always the case for us.

#### Conclusion

We are currently in the leaky state world of workaround 1 right now, but will likely start using workaround 3 with our own custom `connect()` function that wraps `connectAdvanced()`, adding memoization and a decision of whether to subscribe to the store. However, it would cleaner for us, and would make the react-redux API more consistent, if the `mergeProps` argument of `connect()` was changed to accept a factory function just like `mapStateToProps` does. Then we could do this:

```js
const makeMergeProps = 
const ContainerComp = connect(
 state => ({unusedInComp: state.a, usedInComp: state.b}),
 dispatch => ({ 
   fireAction: (unusedInComp, usedInComp) => dispatch(createAction(unusedInComponent, usedInComp)) 
 }),
 () => {
  const memoizedCreateFireAction = (fireAction, unusedInComp) => {
    return usedInComp => dispatch(createAction(unusedInComponent, usedInComp)) 
  };
  return (stateProps, dispatchProps) => ({ 
    usedInComp: stateProps.usedInComp,
    presentationalFireAction: memoizedCreateFireAction(dispatchProps.fireAction, stateProps.unusedInComp)
  })
}
)(PresentationalComp);
```

timdorr:
Yeah, after looking at the use cases, I don't think this is necessary. You're moving too much of your store behavior out of the store/middleware. So, a dispatch from a connected component might not behave like a call elsewhere. I'd argue that's an anti-pattern. If you think about testing, you should be able to use your store entirely without a UI layer. In this case, you wouldn't be able to.

PR #629, opened 2017-02-22
Add React 16 as a valid peer dep

gaearon:
We need this since we recently synced React 16 alpha changes to React Native.

### DIFF
--- a/package.json
+++ b/package.json
@@ -103,7 +103,7 @@
     "loose-envify": "^1.1.0"
   },
   "peerDependencies": {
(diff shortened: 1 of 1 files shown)


gaearon:
@jimbolla @timdorr If you plan to cut a new release, could you include this one with it? I could cut it separately myself too if you don't have the time. Thanks!

timdorr:
Shouldn't be a problem. We have a couple outstanding patch level changes to push, so this will be rolled up with those.

gaearon:
Thank you!

PR #630, opened 2017-02-22
Fixing Grammar

gs-akhan:
A small typo was spotted in mergeProps


markerikson:
That's not actually a typo.  `mergeProps` should be a function, and if you provide it, it is called with the arguments listed.  So, "it is passed" should be valid grammar, although maybe a bit awkward.

gs-akhan:
oh ok.. closing then.. :)

PR #660, opened 2017-03-30
Allow React 16 as peerDependency for 4.x

gaearon:
This is currently causing warnings in our React Native builds at Facebook.
Could you please merge this and release a bump?

I could do this myself but I haven't released in a while and don't want to screw up 😛


timdorr:
Yep, no problem.

timdorr:
Oh wait, this is for 4.0. Did those teams run into problems moving to 5.0? I know that might be a bigger ask, but BC should be maintained despite the major bump.

timdorr:
Pushed as 4.4.7

gaearon:
I don't think we're using 5.x anywhere yet, will raise issues if we find problems.

PR #663, opened 2017-04-08
Fix deprecation warnings for react >= 15.5.0

Calyhre:
Migrate `React.PropTypes` to `PropTypes` from the `prop-types` package, regarding the latest `react` version -> https://facebook.github.io/react/blog/2017/04/07/react-v15.5.0.html#migrating-from-react.proptypes

### DIFF
--- a/package.json
+++ b/package.json
@@ -100,7 +100,8 @@
     "invariant": "^2.0.0",
     "lodash": "^4.2.0",
     "lodash-es": "^4.2.0",
-    "loose-envify": "^1.1.0"
+    "loose-envify": "^1.1.0",
+    "prop-types": "^15.0.0"
   },
   "peerDependencies": {
     "react": "^0.14.0 || ^15.0.0-0 || ^16.0.0-0",
--- a/src/components/Provider.js
+++ b/src/components/Provider.js
@@ -1,4 +1,5 @@
-import { Component, PropTypes, Children } from 'react'
+import { Component, Children } from 'react'
+import PropTypes from 'prop-types'
 import { storeShape, subscriptionShape } from '../utils/PropTypes'
 import warning from '../utils/warning'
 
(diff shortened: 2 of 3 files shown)


jimbolla:
Does this work if someone is still targeting React 14?

wojtekmaj:
It should. PropTypes were not wildly modified since then.

gaearon:
It should work. But I'm not sure if putting it in dependencies would be better.

gaearon:
Yes, lets change this to use dependencies instead.

Calyhre:
Ok :) I moved it from devDependencies to dependencies, and drop the line for peers.

gaearon:
We probably don't need Yarn lock? Seems unrelated to this PR.

Calyhre:
👍  yarn.lock modifications removed

timdorr:
Yes, you do. yarn.lock files are updated with every PR that modifies the dependencies. It's the same for other systems that use lock files (e.g., rubygems)

timdorr:
Thanks, @Calyhre. I've got some follow-on fixes for the tests and to put the lock file back. I'll add those after I merge this in. Thanks!

gaearon:
>Yes, you do. yarn.lock files are updated with every PR that modifies the dependencies

Oops, pardon the brainfart. I don't really understand why I said that 😛

gaearon:
@Calyhre Would you like to also send a PR to backport this to 4.x

timdorr:
Just pushed out 5.0.4 with this. Thanks!

lama-pacos:
old react version will warning
…………

gaearon:
Yes, we’re discussing this in https://github.com/reactjs/react-redux/issues/669.

gaearon:
I released `react@0.14.9` and `react-dom@0.14.9`, they won’t have the warning.

PR #666, opened 2017-04-09
[4.x] Fix deprecation warnings for react >= 15.5.0

Calyhre:
Follow-up of #663, I backported the patch for new deprecation warnings introduced by react 15.5.0

`react-addons-test-utils` seems to have been moved to `react-dom/test-utils` only on `react >= 15.5.0`, in order to use it we'll have to bump the devDependencies to at least that version.

As it's used only during test phase, I don't think it will be that much of a problem, what do you think?

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,9 +81,8 @@
     "istanbul": "^0.3.17",
     "jsdom": "~5.4.3",
     "mocha": "^2.2.5",
-    "react": "^0.14.0",
-    "react-addons-test-utils": "^0.14.0",
-    "react-dom": "^0.14.0",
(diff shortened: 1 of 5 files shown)


timdorr:
I think this is fine. I guess my only concern is this: does `prop-types` work with React 0.14?

Calyhre:
If you look at the travis build for 1e749f2, prior of the `react` bump, `prop-types` seems to works with 0.14

gaearon:
Yea it works as it’s completely standalone.

timdorr:
Pushed as 4.4.8

PR #674, opened 2017-04-13
Make `Provider` not need to wrap root component

mfeniseycopes:
Provider only needs to wrap an ancestor component to pass store to `connect`. I've updated the api docs to reflect this. 

Also, made the `connectAdvanced` remarks more explicit (the "it"s were very vague sounding).


timdorr:
Thanks!

PR #675, opened 2017-04-13
`Provider` can wrap non-root components

mfeniseycopes:
Provider only needs to wrap an ancestor component to pass store to `connect`. A small reference, but it's always good to use consistent language.


timdorr:
Thanks!

PR #685, opened 2017-04-24
Remove state between connectAdvanced calls in example

colepatrickturner:
When I saw this example I was initially excited by the prospect of storing `state` between `connectAdvanced` calls.

I thought of the uses like storing the deleted todo items, or even just the count. But then it dawned on me that the idiomatic approach to Redux, such as the ability for time travel - is challenged when the previous state is stored. For example if I stored `deletedTodoCount` and then used time travel this figure would be wrong because the previous state would be from the future.

For newbies like myself, suggesting this in documentation can be misleading and might encourage the wrong practice of depending on the difference between states for props when really should come from the store.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -361,7 +361,6 @@ import * as actionCreators from './actionCreators'
 import { bindActionCreators } from 'redux'
 
 function selectorFactory(dispatch) {
-  let state = {}
   let ownProps = {}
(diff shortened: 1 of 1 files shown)


timdorr:
Well, it's not used there anyways, so that's OK to remove. Thanks!

PR #691, opened 2017-04-27
Added a new troubleshoot

DanZeuss:
Accessing PropTypes via the main React package is deprecated. Use the prop-types package from npm instead.


timdorr:
Thanks!

PR #695, opened 2017-05-10
Add a way to create a Provider using a specific store key

### LINKED ISSUE #693: custom storeKey in connect() and Provider conflict
`connect` allows the user to set a custom `storeKey`, which is interesting when we have multiple store.

But the `<Provider>` component doesn't allow to set a custom store key. So does `connect()` only work when using the store as prop instead of context?

Will it make sense to add a function to create a custom provider using a specific store key as well?

tleunen:
This is related and fix #693. It offers an easy way to create a custom provider using a specific store key.

`connect` already supported that feature, but the `Provider` was missing it.


markerikson:
Hmm.  Just from a maintenance standpoint, does `createProvider()` actually need to be off in its own file?  I think I'd prefer to see it in the existing file so that we can minimize the history churn.

Other than that, this seems reasonable at first glance in terms of concept and implementation.

tleunen:
Sure. I changed it. Let me know if you'd like something else.
No rush, take your time to review.

src/components/Provider.js
```diff
@@ -19,38 +19,43 @@ function warnAboutReceivingStore() {
...
-  store: storeShape.isRequired,
-  children: PropTypes.element.isRequired
-}
-Provider.childContextTypes = {
-  store: storeShape.isRequired,
-  storeSubscription: subscriptionShape
+    Provider.propTypes = {
+        store: storeShape.isRequired,
+        children: PropTypes.element.isRequired,
+    };
```
timdorr: Nit: no semicolons

markerikson:
I assume if we squash this, it wouldn't reflect the back-and-forth between files, right?

tleunen:
It would reflect this: https://github.com/reactjs/react-redux/pull/695/files
I can manually squash if you'd like (after removing the extra semicolon)

timdorr:
No need, we can squash on GitHub now. I do it every time, since it provides a nice link back to this PR.

markerikson:
:+1: 

Hit it!

PR #702, opened 2017-05-18
Update api.md

stratigos:
* Fix reference to function name in `mapDispatchToProps` documentation.


stratigos:
Original:

> If you don't want to subscribe to store updates, pass `null` or `undefined` in place of `mapStateToProps`.

My change:

>  If you don't want to subscribe to store updates, pass `null` or `undefined` in place of `mapDispatchToProps`. 

Alternatively, it may have been intended to be expressed as follows:

> If you don't want to subscribe to store updates, pass `null` or `undefined` as the second argument of `connect`, just after `mapStateToProps`, e.g., `connect(mapStateToProps, null)`.

timdorr:
Unless I'm wrong, that's correct as-is. If you don't want to subscribe to state, you shouldn't map state.

stratigos:
The original statement is indeed true as-is. However, it kind of jarred me out of context, when reading about it _within_ the documentation for `mapDispatchToProps`. It felt to me like it was referring to passing `null` or `undefined` in as the _second_ argument being discussed, and `mapStateToProps` seemed like an IDE-autocomplete typo. 

Simply close this if you disagree, and Ill try and correct perception on my end :).

markerikson:
To be honest, the current wording feels like the result of the `mapState` paragraph being copied and edited, rather than being written from scratch.  I think that sentence should just be removed.

Issue #709, opened 2017-05-29
documentation: inject complete state into component, re-render


ghost23:
Hi,

in the documentation:

https://github.com/reactjs/react-redux/blob/master/docs/api.md#inject-dispatch-and-every-field-in-the-global-state

It says, that injecting the complete state into a component is a performance killer, since the component would re-render on every action.

My understanding is, that the component would only re-render on a state change, since the default guards (like areStatesEqual and areStatePropsEqual) would still work. If my understanding is correct and since some actions might only result in a side effect (when using redux-saga, for example), I think the documentation should rather say:

_"Don’t do this! It kills any performance optimizations because TodoApp will rerender after every **state change**."_

timdorr:
Feel free to submit a PR to get this changed. Just click the little pen icon at the top right of the file :)

PR #710, opened 2017-05-29
Update api.md

ghost23:
Injecting the state into a component will rerender the component on every state change instead of every action.

Also see issue #709


timdorr:
Thanks!

PR #714, opened 2017-06-06
Fix documentation of connect pure option

testerez:
If we trust the actual documentation `mapStateToProps` would not be called on state changes if not relevant. This is wrong as this is actually `mapStateToProps` that selects the relevant state slices.


jimbolla:
The wording of the original can probably be improved but your edits are incorrect.

For example, calls to `mapStateToProps` will be avoided if `areStatesEqual` and `areOwnPropsEqual` both return true. This is significant in relation to when pure is false, because then `mapStateToProps` is still called even if those functions would otherwise short-circuit it.

timdorr:
Yeah, sorry, but Jim's right on this and these changes don't hold true.

testerez:
Ok, thanks for fixing my understanding.

PR #715, opened 2017-06-08
Persist Nested Subscriptions Through Hot Reload

### LINKED ISSUE #636: react-redux 5.0.3 breaks hot-reloading
### Description

After updating from `react-redux` `5.0.2` to `5.0.3` hot reloading breaks the redux state.

### Expected behavior

hot reloading should not break the redux state.

### Actual behavior

with `react-redux` `5.0.3` the redux state breaks after a hot reload has happened.

### Environment

`react-hot-loader` `3.0.0-beta.6`

### Reproducible Demo

Any project with `react-hot-loader` will do, just upgrade `react-redux` to `5.0.3` and experience the issue.

### LINKED ISSUE #670: HMR: connected component gets disconnected from store
### Description

Connected component gets disconnected from store when code changed and applied by HMR.

Steps:
* fetch demo repo: https://github.com/hoschi/bare-minimum-react-hot-rr4-redux/tree/disconnected-component
* start dev server and open site in browser: http://localhost:8080/page2
* use 'Increment' button to increment counter state to 6

### Expected behavior

Remove [`<p>test</p>`](https://github.com/hoschi/bare-minimum-react-hot-rr4-redux/blob/08740ca8bac9a2d2d7e7527346f0ae36dde3234e/src/components/AppBarDesktop.jsx#L9 ) and ['Increment' component](https://github.com/hoschi/bare-minimum-react-hot-rr4-redux/blob/08740ca8bac9a2d2d7e7527346f0ae36dde3234e/src/components/Increment.jsx#L9 ) should receive changed state when clicked on one of the buttons.

### Actual behavior

When I remove `<p>test</p>` HMR does its code updating which leads to the problem that 'Increment' component doesn't get new store state. You can click on the buttons and the counter stays the same. I can see the actions dispatched in Redux devtools.

I found out that this is the case when the component above in the hierarchy is a container component which subscribed to state which is not subscribed in other components in the tree. You don't get the problem when you change
https://github.com/hoschi/bare-minimum-react-hot-rr4-redux/blob/08740ca8bac9a2d2d7e7527346f0ae36dde3234e/src/components/AppBarDesktopContainer.jsx
to:
```js
import { connect } from 'react-redux';
import AppBarDesktop from './AppBarDesktop';

export default connect((state) => ({
    newsData:state.counter
}))(AppBarDesktop);
```
and do the same steps described in the Description part.

### Environment

react-redux 5.0.4, I tried also 5.0.2 and 5.0.3 without effect
https://github.com/hoschi/bare-minimum-react-hot-rr4-redux/blob/08740ca8bac9a2d2d7e7527346f0ae36dde3234e/package.json

Run these commands in the project folder and fill in their results:

1. `node -v`: 6.3.1
2. `npm -v`: 3.10.3

Then, specify:

1. Operating system: Linux (Arch)
2. Browser and version: Chrome, 56.0.2924.87 (64-bit)

### Reproducible Demo

https://github.com/hoschi/bare-minimum-react-hot-rr4-redux/tree/disconnected-component

### Probably related issues

* https://github.com/gaearon/react-hot-loader/issues/492
* https://github.com/reactjs/react-redux/issues/636
* https://github.com/gaearon/react-hot-loader/issues/539

fun fact: You have to delete the tag below, deleting one above will trigger [this bug](https://github.com/gaearon/react-hot-loader/issues/546 ) and 'fixes' the one I'm reporting here.

dsgkirkby:
This resolves #670.
This likely also resolves #636 but I encounter errors installing the linked repo on my machine.

## Cause Of Issue
When a connected component hot reloads, it unsubscribes and then resubscribes to store updates. As part of unsubscription, all child listeners (also referred to as `nestedSubs`) are removed. For child components that also hot-reload, this is fine, as they will resubscribe during the hot reload. However, if any of the hot reloading parent's descendants are connected but do **not** hot reload, they are never resubscribed to store updates.

## Solution
To resolve this, we persist the previous component's listeners to the reloaded version of the component, so that descendants that did not reload are still subscribed properly to store updates.

## Downsides
This does mean that the old versions of hot reloaded components are still subscribed, but when they unmount as part of the reload, their `selector.shouldComponentUpdate`, `selector.run`, and `notifyNestedSubs` are all set to no-ops, making their `onStateChange` always a no-op itself.

### DIFF
--- a/src/components/connectAdvanced.js
+++ b/src/components/connectAdvanced.js
@@ -271,9 +271,22 @@ export default function connectAdvanced(
           this.version = version
           this.initSelector()
 
-          if (this.subscription) this.subscription.tryUnsubscribe()
+          // If any connected descendants don't hot reload (and resubscribe in the process), their
+          // listeners will be lost when we unsubscribe. Unfortunately, by copying over all
+          // listeners, this does mean that the old versions of connected descendants will still be
+          // notified of state changes; however, their onStateChange function is a no-op so this
+          // isn't a huge deal.
+          let oldListeners = [];
+
+          if (this.subscription) {
+            oldListeners = this.subscription.listeners.get()
+            this.subscription.tryUnsubscribe()
+          }
           this.initSubscription()
-          if (shouldHandleStateChanges) this.subscription.trySubscribe()
+          if (shouldHandleStateChanges) {
+            this.subscription.trySubscribe()
+            oldListeners.forEach(listener => this.subscription.listeners.subscribe(listener))
+          }
         }
       }
     }
--- a/src/utils/Subscription.js
+++ b/src/utils/Subscription.js
@@ -24,6 +24,10 @@ function createListenerCollection() {
       }
     },
 
+    get() {
+      return next
+    },
+
     subscribe(listener) {
       let isSubscribed = true
       if (next === current) next = current.slice()
--- a/test/components/connect.spec.js
+++ b/test/components/connect.spec.js
@@ -60,6 +60,19 @@ describe('React', () => {
         : prev
     }
 
+    function imitateHotReloading(TargetClass, SourceClass, container) {
+      // Crude imitation of hot reloading that does the job
+      Object.getOwnPropertyNames(SourceClass.prototype).filter(key =>
+        typeof SourceClass.prototype[key] === 'function'
+      ).forEach(key => {
+        if (key !== 'render' && key !== 'constructor') {
+          TargetClass.prototype[key] = SourceClass.prototype[key]
+        }
+      })
+
+      container.forceUpdate()
+    }
+
     it('should receive the store in the context', () => {
       const store = createStore(() => ({}))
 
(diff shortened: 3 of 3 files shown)


markerikson:
Sweet, thanks!  We'll definitely want @jimbolla to take a look at this, and I'll try to find time to look over it myself as well.

jimbolla:
At first glance it makes sense, but I'll have to think awhile about any foreseeable issues.

jimbolla:
At any rate, any issues would only affect hot-reloading, which is currently broken, so this is almost certainly an improvement.

GeKorm:
Is something blocking this? Is there anything we can do to help?

adkasp:
it would make a life much easier for many devs, please pull

jimbolla:
I think this is OK to merge.

timdorr:
Cool. Here we go...

3upzorz:
Any idea when this will be published?

olegstepura:
Hi!

This PR should fix an issue #636 which makes our work very unconfortable due to hot reload issues. Personally I tried to downgrade to latest `v4` as a workaround, but this broke rounting with `react-router v4`, so it was not an option for me.

I think I will speak for many developers waiting if I'll kindly ask you to please release this fix as soon as possible.

Thaтks a lot!

timdorr:
OK, pushed out 5.0.6.

pugnascotia:
I ran into this just the other day, so awesome to get it fixed! Thanks everyone!

olegstepura:
This seem to fix #636

PR #724, opened 2017-06-20
MapStateToProps shorthand syntax: Supporting factory selectors

josepot:
Like #723 but supporting factory selectors... Because I'm of the opinion that the object-shorthand syntax should also [support factory selectors](https://github.com/reactjs/react-redux/pull/724/files#diff-b5fd7a42ea1b8efa382416b4a323003cR1953) like @jimbolla [suggested in the past](https://github.com/reactjs/react-redux/pull/323#issuecomment-272251764). 😄 

cc: @gaearon, @jimbolla , @markerikson , @slorber, @timdorr

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -52,9 +52,11 @@ It does not modify the component class passed to it; instead, it *returns* a new
 <a id="connect-arguments"></a>
 #### Arguments
 
-* [`mapStateToProps(state, [ownProps]): stateProps`] \(*Function*): If this argument is specified, the new component will subscribe to Redux store updates. This means that any time the store is updated, `mapStateToProps` will be called. The results of `mapStateToProps` must be a plain object, which will be merged into the component’s props. If you don't want to subscribe to store updates, pass `null` or `undefined` in place of `mapStateToProps`. 
+* [`mapStateToProps(state, [ownProps]): stateProps`] \(*Function* or *Object*): If this argument is specified, the new component will subscribe to Redux store updates. This means that any time the store is updated, `mapStateToProps` will be called. The results of `mapStateToProps` must be a plain object, which will be merged into the component’s props. If you don't want to subscribe to store updates, pass `null` or `undefined` in place of `mapStateToProps`. 
 
   If your `mapStateToProps` function is declared as taking two parameters, it will be called with the store state as the first parameter and the props passed to the connected component as the second parameter, and will also be re-invoked whenever the connected component receives new props as determined by shallow equality comparisons.  (The second parameter is normally referred to as `ownProps` by convention.)
+
+  If an object is passed, each function inside it is assumed to be a Redux selector (or -for advanced scenarios- a Factory function, see below). An object with the same keys, but with the result of invoking each one of its values like the normal `mapStateToProps` function, will be merged into the component’s props.
 
   >Note: in advanced scenarios where you need more control over the rendering performance, `mapStateToProps()` can also return a function. In this case, *that* function will be used as `mapStateToProps()` for a particular component instance. This allows you to do per-instance memoization. You can refer to [#279](https://github.com/reactjs/react-redux/pull/279) and the tests it adds for more details. Most apps never need this.
 
@@ -222,8 +224,8 @@ export default connect(mapStateToProps, mapDispatchToProps)(TodoApp)
 ```js
 import { addTodo, deleteTodo } from './actionCreators'
 
-function mapStateToProps(state) {
-  return { todos: state.todos }
+const mapStateToProps = {
+  todos: state => state.todos
 }
 
 const mapDispatchToProps = {
@@ -355,6 +357,35 @@ function mapDispatchToPropsFactory(initialState, initialProps) {
 
 
 export default connect(mapStateToPropsFactory, mapDispatchToPropsFactory)(TodoApp)
+```
+
+The object shorthand of `stateToProps` also accepts Factory functions
+
+```js
+import { addTodo } from './actionCreators'
+
+const always = x => () => x;
+
+const mapStateToProps = {
+  anotherProperty: (initialState, initialProps) =>
+    always(200 + initialState[initialProps.another]),
+  someProperty: (initialState, initilProps) => createSelector(...),
+  todos: state => state.todos,
+}
+
+function mapDispatchToPropsFactory(initialState, initialProps) {
+  function goToSomeLink(){
+    initialProps.history.push('some/link');
+  }
+  return function(dispatch){
+    return {
+      addTodo
+    }
+  }
+}
+
+
+export default connect(mapStateToProps, mapDispatchToPropsFactory)(TodoApp)
 ```
 
 <a id="connectAdvanced"></a>
--- a/package-lock.json
+++ b/package-lock.json
@@ -2637,14 +2637,6 @@
             }
           }
         },
-        "string_decoder": {
-          "version": "1.0.1",
-          "bundled": true,
-          "dev": true,
-          "requires": {
-            "safe-buffer": "5.0.1"
-          }
-        },
         "string-width": {
           "version": "1.0.2",
           "bundled": true,
@@ -2653,6 +2645,14 @@
             "code-point-at": "1.1.0",
             "is-fullwidth-code-point": "1.0.0",
             "strip-ansi": "3.0.1"
+          }
+        },
+        "string_decoder": {
+          "version": "1.0.1",
+          "bundled": true,
+          "dev": true,
+          "requires": {
+            "safe-buffer": "5.0.1"
           }
         },
         "stringstream": {
@@ -6620,15 +6620,6 @@
       "integrity": "sha1-NbCYdbT/SfJqd35QmzCQoyJr8ks=",
       "dev": true
     },
-    "string_decoder": {
-      "version": "1.0.3",
-      "resolved": "https://registry.npmjs.org/string_decoder/-/string_decoder-1.0.3.tgz",
-      "integrity": "sha512-4AH6Z5fzNNBcH+6XDMfA/BTt87skxqJlO0lAh3Dker5zThcAxG6mKz+iGu308UKoPPQ8Dcqx/4JhujzltRa+hQ==",
-      "dev": true,
-      "requires": {
-        "safe-buffer": "5.1.1"
-      }
-    },
     "string-width": {
       "version": "2.1.1",
       "resolved": "https://registry.npmjs.org/string-width/-/string-width-2.1.1.tgz",
@@ -6654,6 +6645,15 @@
             "ansi-regex": "3.0.0"
           }
         }
+      }
+    },
+    "string_decoder": {
+      "version": "1.0.3",
+      "resolved": "https://registry.npmjs.org/string_decoder/-/string_decoder-1.0.3.tgz",
+      "integrity": "sha512-4AH6Z5fzNNBcH+6XDMfA/BTt87skxqJlO0lAh3Dker5zThcAxG6mKz+iGu308UKoPPQ8Dcqx/4JhujzltRa+hQ==",
+      "dev": true,
+      "requires": {
+        "safe-buffer": "5.1.1"
       }
     },
     "stringstream": {
--- a/src/connect/mapStateToProps.js
+++ b/src/connect/mapStateToProps.js
@@ -1,9 +1,26 @@
-import { wrapMapToPropsConstant, wrapMapToPropsFunc } from './wrapMapToProps'
+import {
+  wrapMapToPropsConstant,
+  wrapMapToPropsFunc,
+  wrapMapToPropsObject
+} from './wrapMapToProps'
 
 export function whenMapStateToPropsIsFunction(mapStateToProps) {
   return (typeof mapStateToProps === 'function')
     ? wrapMapToPropsFunc(mapStateToProps, 'mapStateToProps')
     : undefined
+}
+
+function isValidmapStateToPropsObj(mapStateToProps) {
+  return typeof mapStateToProps === 'object' && Object
+    .keys(mapStateToProps)
+    .map(function (key) { return mapStateToProps[key] })
+    .every(function (val) { return typeof val === 'function' })
+}
+
+export function whenMapStateToPropsIsObject(mapStateToProps) {
+  return (isValidmapStateToPropsObj(mapStateToProps)) ?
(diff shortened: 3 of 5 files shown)


docs/api.md
```diff
@@ -56,6 +56,8 @@ It does not modify the component class passed to it; instead, it *returns* a new
 
   If your `mapStateToProps` function is declared as taking two parameters, it will be called with the store state as the first parameter and the props passed to the connected component as the second parameter, and will also be re-invoked whenever the connected component receives new props as determined by shallow equality comparisons.  (The second parameter is normally referred to as `ownProps` by convention.)
 
+  If an object is passed, each function inside it is assumed to be a Redux selector or a Factory function.
```
pcwinters: don't forget on line 55 to document the argument as `[`mapStateToProps(state, [ownProps]): stateProps`] \(*Object* or *Function*)`

pcwinters:
Am I right to assume that, using this shorthand, a selector could not return a function that wasn't intended to be a factory?

```
@connect({
  getterFnForFooState: (state) => () => state.foo
})
```

I would remove the factory function feature from the MR because I think it would be simpler and more easily justify the inclusion of the shorthand. If you're looking to construct selectors per component instance, you can use [@connectAdvanced](https://github.com/reactjs/react-redux/blob/master/docs/api.md#connectadvancedselectorfactory-connectoptions).

markerikson:
Yeah, I'm afraid I've been focusing on other tasks and not actively trying to get this (and other long-standing PRs) pushed through to completion.

I can't promise a specific deadline, but I will put reviewing this as a high priority and will do my best to get to it within the next week-ish.

josepot:
> Am I right to assume that, using this shorthand, a selector could not return a function that wasn't intended to be a factory?

Yes.

> I would remove the factory function feature from the PR because I think it would be simpler and more easily justify the inclusion of the shorthand. If you're looking to construct selectors per component instance, you can use @connectAdvanced

That's fair, but you could also argue that in that case you could just use the normal connector:

```js
@connect(state => ({
  getterFnForFooState: state => () => state.foo
}))
```

Not that I've actually ever run into a situation where I need to do that...

On the other hand I use very often factory selectors. I would find it very handy to be able to have some properties that come from a factory selector and some properties that come from a normal selector, that's a situation that I run pretty often into.

> I bet it's going to be here for awhile with such a substantial change to the API.

I don't think that this is a substantial change to the API at all, since all this changes are backwards compatible. This is just a small addition to the API, just an _arguably_ handy shorthand... The API remains the same... Notice that the changes into the documentation are minimal, that itself indicates that the API doesn't change much.

cc @pcwinters

jmar777:
It would be great to see this get landed - I'd love to help if there's anything remaining to be done.

I did want to weigh in _against_ the factory properties, though:

1. It seems like a POLA violation, as functions are perfectly valid values for properties.
2. Relatedly, it's inconsistent with the existing API, which allows function values in the object returned from `mapStateToProps`.
3. The current `mapStateToProps` implementation already lends itself well to nested factory functions, if desired.

markerikson:
Well, I clearly didn't put reviewing this as a high priority, unfortunately.  Sorry :(

But, I am actually trying to review this now.  I'm not 100% familiar with the current React-Redux v5 implementation to begin with, so I'll need some time to understand how this actually changes things.

markerikson:
Initial thoughts:

1. Not immediately related to this PR: while I understand what React-Redux v5 does at a high level, I'll admit that trying to trace what's going on through the maze of internal factories is kinda painful.  Fortunately the code shouldn't be changing much going forward, but it does make it more difficult to figure out how this change affects things. 

2. I don't think we actually have _any_ tests related to the existing object shorthand capability for `mapDispatch`.  That seems like a problem.  It would be great if you could add some in this PR, or another separate PR.

3. I would like to see some additional tests added that cover the object shorthand case for `mapState` in more detail.  This PR does add a test to check the behavior of per-field factory functions in an object, but I'd appreciate having some more tests to cover the "basic" selector object behavior as well.

4. The documentation update needs to add more detail.  Please describe what will happen if that object of selectors is passed in, what arguments they will be called with, and how factory selectors interact with that.  We also need some examples added further down the page to demonstrate how to correctly create and pass an object of selectors as `mapState`.

5. What will this do to Flow and TypeScript type defs?  I don't use either of those myself, so I generally stay out of any of the related discussions / issues / PRs.  (To be honest, I don't even know whether the "official" TS and Flow typedefs are actually in the repo here, or in DefinitelyTyped / FlowTyped.)  

6. @jmar777 , can you clarify what you mean by "inconsistent with the current API"?  A `mapState` function shouldn't be used to return functions, just data.  I suppose it's technically valid to include functions in the output of `mapState`, but that's not something we've encouraged, and I don't particularly remember seeing people do that.  But, even if someone _is_ doing that, how are the factory properties "inconsistent" with the API?  The factory selectors would still be ultimately creating a "real" selector, which would then presumably be returning an actual value at that key.

I'm not ready to merge this in yet, but I think it's on the right path.  I think if we can add some more tests around both the existing behavior and the new behavior, and improve the docs appropriately, we can get this in.

Poking @jimbolla to get his thoughts on the PR.

jmar777:
@markerikson 

> can you clarify what you mean by "inconsistent with the current API"

I probably should have worded this more clearly. What I mean is that I would expect this:

```js
connect(state => ({
  foo: 'bar',
  hello: () => console.log('world')
}))
```

...to be ~consistent~ equivalent to the following shorthand:

```js
connect({
  foo: 'bar',
  hello: () => console.log('world')
})
```

In other words, I would expect `connect({ ...someProps })` to always be equivalent to `connect(() => ({ ...someProps }))`. Unless I'm mistaken, I believe that's how the shorthand for `mapDispatchToProps()` behaves as well.

I see your point that function property values aren't an intended use case, but FWIW, I've personally shoved function properties in there in cases where I didn't actually need direct access to `dispatch()`.

Aside from introducing what I'd argue is unexpected behavior, I don't really see the benefit of property-level factory functions. From a verbosity perspective, they're essentially a tie when you have a single property:

```js
// 54 chars
connect(state => ({
  foo: selectors.getFoo(state)
}))

// vs...

// 53 chars
connect({
  foo: state => selectors.getFoo(state)
})
```

...and once you get past a single property, the new shorthand syntax starts to lose badly on that front.

**TL;DR**: it might be a corner case, but the property level factory methods seem surprising to me, and they render the "shorthand" syntax more verbose if people were to actually use them.

markerikson:
@jmar777 : if you want to pass in a function, why not just pass it in as a normal prop?  `<ConnectedComponent hello={() => console.log("World")} />`.

The use case for the property-level factory functions here, I think, is the same as the overall factory function capability already supported by `connect()`: being able to create unique selector instances per component instance.  Here's the test that's added by this PR:

```js
    it('should allow providing a factory function to mapStateToProps object', () => {
      let updatedCount = 0
      let memoizedReturnCount = 0
      const store = createStore(() => ({ value: 1 }))

      const mapStateFactory = () => {
        let lastVal, lastProp
        return (state, props) => {
          if (props.name === lastProp && lastVal === state.value) {
            memoizedReturnCount++
            return lastProp
          }
          lastVal = state.value
          return lastProp = props.name
        }
      }

      @connect({ name: mapStateFactory })
      class Container extends Component {
        componentWillUpdate() {
          updatedCount++
        }
        render() {
          return <Passthrough {...this.props} />
        }
      }

      TestUtils.renderIntoDocument(
        <ProviderMock store={store}>
          <div>
            <Container name="a" />
            <Container name="b" />
          </div>
        </ProviderMock>
      )

      store.dispatch({ type: 'test' })
      expect(updatedCount).toBe(0)
      expect(memoizedReturnCount).toBe(2)
    })
```

So, `{name : mapStateFactory}` is a selector factory function, and both instances of `<Container />` get a unique selector instance.  That way, they each return the exact same values as before, and the components don't re-render.  (The more realistic use case would be looking up something along the lines of `return state.items[ownProps.itemId]`.)

slorber:
```javascript
connect({
  foo: 'bar',
  hello: () => console.log('world')
})
```

This seems not a good idea because on every state change, the connect component would always re-render because it gets a new hello function everytime. Not sure it's

markerikson:
@slorber : to be honest, I'm not even sure exactly what that syntax would do.  I _think_, given this PR, it would be interpreted as a selector, and wind up returning a prop of `{hello : undefined}`.

jmar777:
@markerikson 

>  if you want to pass in a function, why not just pass it in as a normal prop?

Code organization / separation of concerns, primarily. Obviously depends on the specific semantics of the component and the function in question.

@slorber 

> This seems not a good idea because on every state change, the connect component would always re-render because it gets a new hello function everytime.

Right - this was just a minimal example that would exhibit what I believe is surprisingly behavior. The same issue would be exhibited by, e.g.,:

```js
const logWorld = () => console.log('world');

connect({ foo: 'bar', hello: logWorld });
```

@markerikson 

>  I think, given this PR, it would be interpreted as a selector, and wind up returning a prop of {hello : undefined}.

Correct, and that's what I think would be surprising, as returning that same object from the factory function produces different behavior:

```js
const obj = { onClick: someHandler };

// this:
connect(obj); 

// ...behaves differently than this:

connect(() => obj);
```

----

At any rate, I seem to be in the minority here in expecting that objects supplied for the shorthand syntax should behave the same as objects returned from factory functions. While I appreciate the utility of property-level factory functions, inasmuch as they don't enable anything that the existing factory functions already enable, IMHO this just seems more likely to introduce unexpected behavior, and I'd advocate for the least surprisingly implementation.

/2-cents

slorber:
agree with @jmar777 

A simple solution that solves 90% of usecases can be merged right now without risk, until we figure out if we can support more advanced API. Honestly it's just reducing boilerplate so I'm not even sure it's so interesting to support more advanced usecases as it seems not so easy and those usecases are already enabled by more verbose syntax.

The initial simple PR I made was ready 1.5 year ago. 
I feel like this feature will never get merged because the scope is too broad.

Why not reducing the scope and merging something simple now? 
Is there any chance the next steps wouldn't be retrocompatible?

markerikson:
@slorber : I already said I'm pretty much okay with the implementation and scope of this PR, but I want to see more docs and tests added around this.  If you'd like to see this merged, you're welcome to help out with those.

josepot:
@markerikson : sorry that it took me a while to get back to you on this. I've been quite busy these days. I will start working on this tomorrow... I'm hopping to address your comments by the end of this week. Thanks for your feedback!

markusjwetzel:
@josepot : any updates on this?

josepot:
Hi and sorry @markerikson @markusjwetzel for leaving this PR a bit stale... I've been beyond busy these last months.

So, I just did a rebase and I've added a new test. About the improvements that @markerikson said that were necessary before merging this PR. This is what I think:

> I don't think we actually have any tests related to the existing object shorthand capability for mapDispatch. That seems like a problem. It would be great if you could add some in this PR, or another separate PR.

IMO that should be done in a separate PR. I would be very happy to help with that, but first I would like to know what exactly is that we want to test and to what extend. That being said, I just added an extra test that testes the basic functionality of both shorthands.

> I would like to see some additional tests added that cover the object shorthand case for mapState in more detail. This PR does add a test to check the behavior of per-field factory functions in an object, but I'd appreciate having some more tests to cover the "basic" selector object behavior as well.

I'm not sure what's exactly that you want... I mean, in the original PR I made sure that **everything** that worked with the normal `mapState` also worked with its shorthand counterpart... That's why I created the HOF `withMapStateToProps`... I mean, I don't mind adding more tests, the more the merrier, for sure 😄 . That's why I added a new test... is that enough? If not: if you give me the descriptions of the tests that you think that are missing I will gladly implement them.

> The documentation update needs to add more detail. Please describe what will happen if that object of selectors is passed in, what arguments they will be called with, and how factory selectors interact with that. We also need some examples added further down the page to demonstrate how to correctly create and pass an object of selectors as mapState.

I'm on it. It's possible that by the time you read this comment I've already taken care of that. However, please bear in mind that english is not my first language. So, I will try my best but do not hesitate in correcting me. Thanks in advance.

> What will this do to Flow and TypeScript type defs? I don't use either of those myself, so I generally stay out of any of the related discussions / issues / PRs. (To be honest, I don't even know whether the "official" TS and Flow typedefs are actually in the repo here, or in DefinitelyTyped / FlowTyped.)

I have no idea. But I imagine that since these changes are 100% backwards compatible they shouldn't break anything... I mean, yes, Flow and TS users won't be able to use the shorthand until the definitions are updated. But do they need to be updated in this repo? Where exactly? I wouldn't mind creating a PR to FlowTyped once this is in...

Ok, lets finish this already!

docs/api.md
josepot: Done! Thanks!

josepot:
Hi @markerikson! Just a friendly reminder that I'm waiting for your feedback here... Sorry, if you find this comment annoying, it's just that I don't want this PR to become stale again. Thanks!

markerikson:
Afraid I'm on travel at the moment, and don't have a lot of time free.  I'll try to find time to look at it within the next few days.

josepot:
Hi @markerikson ! Just another friendly reminder that I'm waiting for your feedback. Sorry if you find these comments annoying... I'm aware that you must be busy and that this is probably a low-priority PR, but as I said before: I don't want this PR to become stale again. Thanks!

markerikson:
It's on my radar, but I'm still on business travel and don't have much spare time.  I'll try to take a look at it after I get back from this trip.

cellog:
I have to say, I am not a fan of this PR. the number of possibilities for what can be passed for mapStateToProps and mapDispatchToProps is already a bit like reading C++ operator overloading. Better would be to provide different versions of connect that encapsulate the flexibility, or a fluent interface that makes the expected behavior explicit and errors in development if the passed argument varies from what is expected

markerikson:
I do agree that `connect` has gotten too heavily overloaded, and there are certainly userland ways around this (primarily use of Reselect's `createStructuredSelector` ).

I'll leave this PR open for a bit longer, but at the moment I'm leaning towards closing it.

chrisjallen:
Hi @markerikson, can see feature this has been floating in and out of prs for last couple years.  What is the status on this pr? It would a really beneficial feature, is there anything I can do to see this in?

cellog:
A quick note: in the interim, if you need this functionality, you could write a selector factory that provides it, and pass that to connect options

timdorr:
I'll take the bullet.

PR #726, opened 2017-06-23
add "Caveat" section to readme noting how you can't dispatch during `componentWillMount` on the server

faceyspacey:
NOTE: I would have recommended other packages as well, but there are no popular Redux-specific routing packages other than [Redux First Router](https://medium.com/@faceyspacey/pre-release-redux-first-router-a-step-beyond-redux-little-router-cd2716576aea) that offer server-side data-fetching or route-based data-fetching in general. For example the most popular Redux-specific routing package, *Redux Little Router*, does not have data-fetching built in. Which means many users might still end up using `componentWillMount` while using it.

In the future, a short list should be amended here :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,12 @@ If you are on an older version of React Native, you’ll need to keep using [Rea
 We do a deep dive on how React Redux works in [this readthesource episode](https://www.youtube.com/watch?v=VJ38wSFbM3A).  
 Enjoy!
 
+## Caveat
+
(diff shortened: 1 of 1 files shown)


timdorr:
This seems like odd placement. And the routing stuff is a non-sequitur. Data fetching and dispatching in componentWillMount/constructor isn't a new issue to any version of React Redux and there are many solutions for this. I don't think this adds much value to the Readme.

faceyspacey:
Sure it's easier to do nothing. I've been using react-redux for a long time and never knew this. I just happened to have done the right thing. Newer developers like the one I was talking to im sure fall into this trap all the time. This needs to be written somewhere beyond Issue threads, especially as SSR is going to become even more common.

Id be happy to put it somewhere else and edit it in any way you need.

Is it on the Redux site? The thing is if it is, that site has too much info so it will get buried. Somewhere in this repo seems best to me since this is what it specifically applies to.

PR #733, opened 2017-07-06
Move ref context binding to fix object freezing middleware

### LINKED ISSUE #708: Error after freezing store state
Hello

I'm monkeypatching the store to freeze the result of `store.getState()` during development and I'm getting a weird error in only one connected component. The error is: 

`Uncaught TypeError: Cannot assign to read only property 'setWrappedInstance' of object '#<Connect>'`

and it's happening on [this line in connectAdvanced.js](https://github.com/reactjs/react-redux/blob/master/src/components/connectAdvanced.js#L124). What's the relation between `getState()` and `connectAdvanced` so that error is happening?

It's related for sure to freezing the state and also to connect the component, what I don't know if it's the main cause of it's an underlying problem that's being triggered by it. But if I remove one of both conditions (freezing the state or connecting the component) the problem goes away.

EDIT:

This is the monkeypatching code in case it provides more context:

```javascript
if (process.env.NODE_ENV === 'development') {
  const oldGetState = store.getState();
  store.getState = () => deepFreeze(oldGetState());
}
```

gdub22:
Closes #708 

Middleware that uses `Object.freeze` such as [redux-freeze](https://github.com/buunguyen/redux-freeze) started throwing an error somewhere between versions 5.0.0...5.0.5:

```
Uncaught TypeError: Cannot assign to read only property 'setWrappedInstance' of object '#<Connect>'
```

This is because `Connect` instances are being mutated here: https://github.com/reactjs/react-redux/blob/2726c59fa1e5929659067df2088c7c6e0d736f3a/src/components/connectAdvanced.js#L124

This PR moves the context binding of the ref function to a lazy assignment and doesn't mutate the original function.  IMO the intention of the code is clearer this way as well.

### DIFF
--- a/src/components/connectAdvanced.js
+++ b/src/components/connectAdvanced.js
@@ -121,7 +121,6 @@ export default function connectAdvanced(
         this.renderCount = 0
         this.store = props[storeKey] || context[storeKey]
         this.propsMode = Boolean(props[storeKey])
-        this.setWrappedInstance = this.setWrappedInstance.bind(this)
 
         invariant(this.store,
           `Could not find "${storeKey}" in either the context or props of ` +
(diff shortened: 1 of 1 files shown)


jimbolla:
Except now it's creating a new bound function every render. I don't think we want to do this.

timdorr:
LGTM. Thanks!

timdorr:
Oh crap, I didn't follow that code all the way through. 

Binding in the constructor is super-common. I have a hard time believing redux-freeze can't handle this. I'm not sure why it's touching a component instance anyways. That was never made clear to me.

gdub22:
Ok. Here you go: https://github.com/reactjs/react-redux/pull/735

PR #736, opened 2017-07-07
Add additional optional function to the connect function that allows using react-redux without the Provider class

challenger532:


challenger532:
is there any thing I can do to accept this pull request ?

jimbolla:
You can already use connect without Provider by passing the store as a prop. You can leverage https://github.com/acdlite/recompose to build the functionality you want in userland. Basically this:

customconnect.js:
```js
import { compose, withPropsOnChange } from 'recompose'

const customConnect = (mapState, mapDispatch, options, storeProvider) => compose(
  withPropsOnChange(
    () => false,
    () => ({ store: storeProvider.getStore() })
  ),
  connect(mapState, mapDispatch, options)
);

export default customConnect;
```

markerikson:
Yeah, I'm not clear on what this actually does, or what the real benefit is.

timdorr:
Yeah, we've already got this functionality another way. Thanks anyways!

PR #741, opened 2017-07-11
Update api.md

is0sceles:
You may wish to override areStatesEqual to always return false (areStatesEqual: () => false) if you have impure reducers that mutate your store state. (This would likely impact the other equality checks is well, depending on your mapStateToProps function.)

This paragraph's last line said 'is well'


timdorr:
Thanks!

PR #752, opened 2017-07-22
Add createProvider Documentation.

azizhk:
Fixes https://github.com/reactjs/react-redux/issues/738

Open to Grammar, Capitalization & Style changes. Also can squash, rebase etc.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -384,3 +384,40 @@ function selectorFactory(dispatch) {
 export default connectAdvanced(selectorFactory)(TodoApp)
 ```
 
+<a id="createProvider"></a>
+### `createProvider([storeKey])`
+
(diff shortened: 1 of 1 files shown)


docs/api.md
```diff
@@ -384,3 +385,41 @@ function selectorFactory(dispatch) {
 export default connectAdvanced(selectorFactory)(TodoApp)
 ```
 
+<a id="createProvider"></a>
+### `createProvider([storeKey])`
```
azizhk: I wanted to add subKey as well, but then over here:
https://github.com/reactjs/react-redux/blob/a9bb9d3b074eb04f4d7922e22fce733beb148650/src/components/connectAdvanced.js#L78
It is hardcoded to be 
```js
const subscriptionKey = storeKey + 'Subscription'
```
So I don't think I could pass it in `connect` so it would have broken anyways

timdorr: Probably should fix that...

timdorr:
LGTM. Thanks!

PR #807, opened 2017-10-15
feat(connect): return false to skip update

AndersDJohnson:
This adds a new feature allowing us to short-circuit `mapStateToProps`, `mapDispatchToProps`, and `mergeProps` functions by returning `false` to fallback to existing/previous props.

The motivation here is performance optimization. Because these functions may be called with every change to Redux state or own props, they can have a negative impact on performance. Though we can use something like `reselect` memoization, it has implications for memory consumption, an overheard that may not be desired.

In certain cases, we may be able to quickly determine upfront if anything relevant has changed or if the state has any data to which the component needs to respond, and if not, then there's no need to continue computing the props mapping. This is analogous to a React component's `shouldComponentUpdate` lifecycle hook.

Here's an example. Suppose we have a UI that renders cards that can update with real-time data. But it allows cards to be paused, either manually or due to being in the background (minimized, zoomed out, off-screen, etc.). If a card is paused, then we don't need it to update in real-time. With this new feature, we can now detect if `card.paused` is `true`, and if so `return false` to indicate that it can just use previous props, rather than to waste resources going on to call `get` for potentially many extracted fields (or in other ways non-trivially transform the data):

```js
import get from 'lodash/get'

const mapStateToProps = (state, ownProps) => {
  const card = state.cards[ownProps.cardId]

  if (card && card.paused) return false

  return {
    title: get(card, 'titles.full'),
    hasFacebookButton: get(card, 'shareButtons.facebook.isEnabled'),
    // ...
  }
}
```

Here's another example:

```js
import get from 'lodash/get'

const mapStateToProps = state => {
  const product = get(state, 'products.product')

  return {
    title: get(product, 'displayName.full'),
    price: get(product, 'priceData.usd.afterTax'),
    // ...
  }
}
```

Above, if `state.products.product` is `undefined`, it's a waste of time to go on to call `get` for potentially many extracted fields, since we know all of those will also return `undefined`, or to in other ways non-trivially transform the data.

With this new feature, we can now do this:

```js
import get from 'lodash/get'

const mapStateToProps = state => {
  const product = get(state, 'products.product')

  if (!product) return false

  return {
    title: get(product, 'displayName.full'),
    price: get(product, 'priceData.usd.afterTax'),
    // ...
  }
}
```

Now nothing below the `return false` will be executed, and the component will use its previous values for any props.

Ignore the workaround of returning `{}` early here - that may not work for less trivial examples, e.g., where two possibly-`undefined` objects are selected from the state, without introducing more complex branching logic and object merging.

This example may be a bit contrived, since in a real-world scenario in most cases you probably wouldn't want to retain stale data in props from a previous product if the product data were ever deleted from the state. But if you know your app never deletes product data once defined, then this would increase performance for all the action dispatches until the product was defined.

### DIFF
--- a/src/connect/mergeProps.js
+++ b/src/connect/mergeProps.js
@@ -23,7 +23,7 @@ export function wrapMergePropsFunc(mergeProps) {
         mergedProps = nextMergedProps
 
         if (process.env.NODE_ENV !== 'production')
-          verifyPlainObject(mergedProps, displayName, 'mergeProps')
+          mergedProps === false || verifyPlainObject(mergedProps, displayName, 'mergeProps')
       }
 
       return mergedProps
--- a/src/connect/selectorFactory.js
+++ b/src/connect/selectorFactory.js
@@ -1,5 +1,5 @@
 import verifySubselectors from './verifySubselectors'
-  
+
 export function impureFinalPropsSelectorFactory(
   mapStateToProps,
   mapDispatchToProps,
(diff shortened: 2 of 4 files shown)


markerikson:
Hmm.  This is an interesting idea, and I can see the intent behind it, but I'm not sure of the necessity. `mapState` functions should ideally be written to be lightweight overall.

Do you happen to have any example benchmarks that demonstrate some differences in performance?

AndersDJohnson:
@markerikson I don't have any benchmarks, but I assume this is the intent behind the existing `areStatesEqual` option? But I find this implementation more ergonomic.

gaearon:
I don't think it's a great idea to make APIs more loose in terms of typing. Allowing to return `false` introduces the concept of an "update" into a function that's otherwise completely agnostic of it. And it's a lossy abstraction: for example, what would returning `false` on the initial render mean? It's also not friendly to type systems.

So I'd say we won't take it, but I appreciate the effort.

PR #812, opened 2017-10-24
Add TypeScript typings corresponding to redux@next

pelotom:
Add built-in TypeScript typings copied from DefinitelyTyped but modified to accommodate upcoming revamped `redux` typings on the `next` branch (see [redux #2563](https://github.com/reactjs/redux/pull/2563)).

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,295 @@
+import {
+  Component,
+  ComponentClass,
+  ComponentType,
+  ReactNode,
+  StatelessComponent,
+} from 'react'
+
+import {
+  Action,
+  Store,
+  Dispatch,
+  ActionCreator,
+} from 'redux'
+
+// Diff / Omit taken from https://github.com/Microsoft/TypeScript/issues/12215#issuecomment-311923766
+type Diff<T extends string, U extends string> = ({ [P in T]: P } & { [P in U]: never } & { [x: string]: never })[T];
+type Omit<T, K extends keyof T> = Pick<T, Diff<keyof T, K>>;
+
+export interface DispatchProp<S> {
+  dispatch?: Dispatch<S>;
+}
+
+interface AdvancedComponentDecorator<TProps, TOwnProps> {
+    (component: ComponentType<TProps>): ComponentClass<TOwnProps>;
+}
+
+// Injects props and removes them from the prop requirements.
+// Will not pass through the injected props if they are passed in during
+// render. Also adds new prop requirements from TNeedsProps.
+export interface InferableComponentEnhancerWithProps<TInjectedProps, TNeedsProps> {
+    <P extends TInjectedProps>(
+        component: ComponentType<P>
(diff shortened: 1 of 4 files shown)


index.d.ts
```diff
@@ -0,0 +1,295 @@
...
+     * @default false
+     */
+    withRef?: boolean
+}
+
+export interface ProviderProps {
+    /**
+     * The single Redux store in your application.
+     */
+    store?: Store<any, any>;
```
pelotom: This is the main change from what's on DefinitelyTyped, which requires the new `redux` typings. Without it, users will probably run into problems instantiating `Provider` elements in the presence of TypeScript 2.6's `--strictFunctionTypes`.

pelotom:
This is not ready to merge until it can depend on the yet-to-be-released Redux 4 which will have the new typings.

timdorr:
Thanks, @pelotom! Can any TS experts weigh in on this?

index.d.ts
```diff
@@ -0,0 +1,295 @@
...
+  Dispatch,
+  ActionCreator,
+} from 'redux'
+
+// Diff / Omit taken from https://github.com/Microsoft/TypeScript/issues/12215#issuecomment-311923766
+type Diff<T extends string, U extends string> = ({ [P in T]: P } & { [P in U]: never } & { [x: string]: never })[T];
+type Omit<T, K extends keyof T> = Pick<T, Diff<keyof T, K>>;
+
+export interface DispatchProp<S> {
+  dispatch?: Dispatch<S>;
```
aikoven: Isn't the dispatch type parameter an action type now?

aikoven:
Looks good to me.

Tests are missing though.

timdorr:
Ah yes. @pelotom Can you pull in `typescript-definition-tester` to get these tested? Might not work until Redux's `next` branch gets merged.

index.d.ts
```diff
@@ -17,8 +17,8 @@ import {
 type Diff<T extends string, U extends string> = ({ [P in T]: P } & { [P in U]: never } & { [x: string]: never })[T];
 type Omit<T, K extends keyof T> = Pick<T, Diff<keyof T, K>>;
 
-export interface DispatchProp<S> {
-  dispatch?: Dispatch<S>;
+export interface DispatchProp {
+  dispatch?: Dispatch<{}>;
```
pelotom: No need to have a type parameter at all here since we were only instantiating it with `any`. Hard-coding it to `{}` instead of `any` because it's in a contravariant position.

pelotom:
Actually it looks like these definitions can be compatible with both the current and `next` redux typings.

pelotom:
Hm, I just saw #541, and I'm not sure about the relative merits of those typings vs these. The test cases are undoubtedly more comprehensive there. Is there any reason it seems to have been abandoned? Just no one got around to resolving merge conflicts?

pelotom:
Closing this for now as I think #541 is a better foundation to build the `next` typings off of.

PR #815, opened 2017-10-25
Add TypeScript typings (revisited)

pelotom:
This is a continuation of #541 which seems to have stalled. Fixed the merge conflict and upgraded `@types/react`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,5 +6,6 @@ before_install:
 script:
   - npm run lint
   - npm run test:cov
+  - npm run test:typescript
 after_success:
   - npm run coverage
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,70 @@
+import { ComponentClass, Component, StatelessComponent } from 'react';
+import { Store, Dispatch, ActionCreator } from 'redux';
+
+interface ComponentDecorator<TOriginalProps, TOwnProps> {
+  (component: StatelessComponent<TOriginalProps>|ComponentClass<TOriginalProps>): ComponentClass<TOwnProps>;
+}
+
+/**
+ * Following 5 functions cover all possible ways connect could be invoked
+ *
+ * - State: Redux state interface (the same one used by Store<S>)
+ * - TOwnProps: Props passed to the wrapping component
+ * - TStateProps: Result of MapStateToProps
+ * - TDispatchProps: Result of MapDispatchToProps
+ */
+function connect<State, TOwnProps>(): ComponentDecorator<{ dispatch: Dispatch<State> } & TOwnProps, TOwnProps>;
+
+function connect<State, TOwnProps, TStateProps>(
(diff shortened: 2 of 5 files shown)


timdorr:
What is this based on? There are a ton of old commits here.

pelotom:
It's based on #541, so it has all the same commits as there. I'm not sure why there are so many, but as you can see the actual number of final changes is pretty small. I can rebase and squash if you want, but I was erring on the side of preserving history.

aksharpatel47:
@timdorr Can you give feedback on what more needs to be done to get this PR merged?

timdorr:
Got rid of all the junk from the previous branch. No idea how it got to that state, but it's fixed up now.

pelotom:
It looks like the [DefinitelyTyped typings](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-redux/index.d.ts) have undergone significant changes since these typings were developed... I'm unsure now what the relative merits are now of this PR relative to what's there.

Any thoughts from the authors of the DT typings?

/cc @tkqubo @thasner @kenzierocks @clayne11 @tansongyang @nicholasboll @mdibyo @pdeva

timdorr:
Other than taking on ownership, I also doubt the benefits of keeping them here. It causes pain when TS moves forward and we don't. Releases are held to our schedule, which doesn't usually align with TS's schedule. I'd argue against adding this burden.

abenhamdine:
> Other than taking on ownership, I also doubt the benefits of keeping them here. It causes pain when TS moves forward and we don't. Releases are held to our schedule, which doesn't usually align with TS's schedule. I'd argue against adding this burden.

On the other hand, I think it's easier to keep typings up to date when they are in the same repo.
I realized that while contributing to [node-postgres](https://github.com/brianc/node-postgres).

timdorr:
True. I have no connectivity to the TS world. But at the same time, that's an argument for not keeping them here. None of the maintainers here have any experience with TS. We're not able to update them here or on DT. Moving them will not help with update lag.

aikoven:
Agree with @timdorr. Including typings in the source project repo impedes updates on the typings, particularly it doesn't allow to make breaking changes in the typings without bumping the major version of the source project.

With Redux typings, we weren't able to use TS 2.0 features for a very long time because of this. 

Luckily, Redux API is pretty simple in terms of typings. But the same cannot be said of React Redux, which involves higher-order functions — an area in which TS isn't perfect yet, but moves very fast to make it better. Which means that there could be a lot of changes to the typings.

timdorr:
OK, I'm going to close this out. Good DX is more important than us having iron-fisted control.

thasner:
The best scenario is to write code in TS. Then typings are generated automatically 😄.

Otherwise, assuming `react-redux` doesn't have an interest in TS, it's likely best to keep typings in DT where the community can make changes independently.

PR #831, opened 2017-11-30
Re-export compose from redux

larsbs:
Most of the time, when  I want to use `connect` from `react-redux` package I don't have anything to import from the `redux` package. However, I find myself using `compose` most of the time I use `connect`. 

Re-exporting `compose` from `redux` package inside `react-redux` would allow reduce an extra import line.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -3,3 +3,4 @@ import connectAdvanced from './components/connectAdvanced'
 import connect from './connect/connect'
 
 export { Provider, createProvider, connectAdvanced, connect }
+export { compose } from 'redux'


markerikson:
Appreciate the PR, but I don't see a real reason to expand the public API like this.  If you're using `react-redux`, you're also using `redux`, and having a separate import line is not a major burden.

larsbs:
Is that really expanding the public API? It's not more code to maintain, and in fact, one uses more `compose` with `react-redux` than with "pure" `redux`. So it would even make more sense to have it on the `react-redux` package. 

This is something that it's already done in `react-apollo` where they expose their own `compose` function because it's highly possible that if you're using HOCs you're also gonna need a "compose" function.

PR #833, opened 2017-12-01
Proposal for addition of a modifier of onStateChange

asturur:
**Why**

i m working on a fairly complex application that heavily rely on react-redux combination.
Our mapStateToProps actually run some logic and also the components take their time to re renders.

When dispatching actions everything is fine.

When we dispatch a couple of actions in series, even if they interest 2 different part of the store or simply update something that will make a component change but in 2 different steps, the added overhead of a dispatch is noticeable

**How**

The solution we came with is to debounce the onStateChange of our root connected component with a lodash debounce ( 3 or 4 ms )

This allow dispatched fires in rapid sequence to fire just one onStateChange.

example:

```javascript
const myDebouncer = onStateChange => lodash.debounce(onStateChange, 4);
@connect(state => mapStateToProps, null, null, { onStateChangeDecorator: myDebouncer } )
export default class App extends Component { }
```

I understand this may seems an application need, but i also think is very usefull.
I did not find anyway to obtain this with the exposed options or with a different reduxMiddleware.

Your input on this matter is really appreciated as any help to achieve the same effect with a different solution

### DIFF
--- a/src/components/connectAdvanced.js
+++ b/src/components/connectAdvanced.js
@@ -197,7 +197,11 @@ export default function connectAdvanced(
         // parentSub's source should match where store came from: props vs. context. A component
         // connected to the store via props shouldn't use subscription from context, or vice versa.
         const parentSub = (this.propsMode ? this.props : this.context)[subscriptionKey]
-        this.subscription = new Subscription(this.store, parentSub, this.onStateChange.bind(this))
+        let onStateChange = this.onStateChange.bind(this);
+        if (connectOptions.onStateChangeDecorator) {
+          onStateChange = connectOptions.onStateChangeDecorator(onStateChange)
+        }
+        this.subscription = new Subscription(this.store, parentSub, onStateChange)
 
         // `notifyNestedSubs` is duplicated to handle the case where the component is  unmounted in
         // the middle of the notification loop, where `this.subscription` will then be null. An


asturur:
travis failed, i run npm run test before pushing, should i try something else?

```
  React
    ✓ should pass state consistently to mapState
    Provider
      ✓ should enforce a single child
      ✓ should add the store to the child context
      ✓ should add the store to the child context using a custom store key
      ✓ should warn once when receiving a new store in props
      ✓ should handle subscriptions correctly when there is nested Providers

  React
    connect
      ✓ should receive the store in the context
      ✓ should pass state and props to the given component
      ✓ should subscribe class components to the store changes
      ✓ should subscribe pure function components to the store changes
      ✓ should retain the store's context
      ✓ should handle dispatches before componentDidMount
      ✓ should handle additional prop changes in addition to slice
      ✓ should handle unexpected prop changes with forceUpdate()
      ✓ should remove undefined props
      ✓ should remove undefined props without mapDispatch
      ✓ should ignore deep mutations in props
      ✓ should allow for merge to incorporate state and prop changes
      ✓ should merge actionProps into WrappedComponent
      ✓ should not invoke mapState when props change if it only has one argument
      ✓ should invoke mapState every time props are changed if it has zero arguments
      ✓ should invoke mapState every time props are changed if it has a second argument
      ✓ should not invoke mapDispatch when props change if it only has one argument
      ✓ should invoke mapDispatch every time props are changed if it has zero arguments
      ✓ should invoke mapDispatch every time props are changed if it has a second argument
      ✓ should pass dispatch and avoid subscription if arguments are falsy
      ✓ should unsubscribe before unmounting
      ✓ should not attempt to set state after unmounting
      ✓ should not attempt to notify unmounted child of state change
      ✓ should not attempt to set state after unmounting nested components
      ✓ should not attempt to set state when dispatching in componentWillUnmount
      ✓ should shallowly compare the selected state to prevent unnecessary updates
      ✓ should shallowly compare the merged state to prevent unnecessary updates
      ✓ should throw an error if a component is not passed to the function returned by connect
      ✓ should throw an error if mapState, mapDispatch, or mergeProps returns anything but a plain object
      ✓ should recalculate the state and rebind the actions on hot update
      ✓ should persist listeners through hot update
      ✓ should set the displayName correctly
      ✓ should expose the wrapped component as WrappedComponent
      ✓ should hoist non-react statics from wrapped component
      ✓ should use the store from the props instead of from the context if present
      ✓ should throw an error if the store is not in the props or context
      ✓ should throw when trying to access the wrapped instance if withRef is not specified
      ✓ should return the instance of the wrapped component for use in calling child methods
      ✓ should wrap impure components without supressing updates
      ✓ calls mapState and mapDispatch for impure components
      ✓ should pass state consistently to mapState
      ✓ should not render the wrapped component when mapState does not produce change
      ✓ should bail out early if mapState does not depend on props
      ✓ should not swallow errors when bailing out early
      ✓ should allow providing a factory function to mapStateToProps
      ✓ should allow a mapStateToProps factory consuming just state to return a function that gets ownProps
      ✓ should allow providing a factory function to mapDispatchToProps
      ✓ should not call update if mergeProps return value has not changed
      ✓ should update impure components with custom mergeProps
      ✓ should allow to clean up child state in parent componentWillUnmount
      ✓ should allow custom displayName
      ✓ should update impure components whenever the state of the store changes
      ✓ should throw a helpful error for invalid mapStateToProps arguments
      ✓ should throw a helpful error for invalid mapDispatchToProps arguments
      ✓ should throw a helpful error for invalid mergeProps arguments
      ✓ should notify nested components through a blocking component
      ✓ should subscribe properly when a middle connected component does not subscribe
      ✓ should subscribe properly when a new store is provided via props

  Utils
    shallowEqual
      ✓ should return true if arguments fields are equal
      ✓ should return false if arguments fields are different function identities
      ✓ should return false if first argument has too many keys
      ✓ should return false if second argument has too many keys
      ✓ should return false if arguments have different keys

  Utils
    wrapActionCreators
      ✓ should return a function that wraps argument in a call to bindActionCreators


  70 passing (963ms)
```

timdorr:
Rebase against master. Travis updated to node 9 by default and that's not working. I just locked us back to node 8 on master.

markerikson:
I'm not sure this is the right approach.

If you need to cut down on the number of store update notifications, you can do it several ways: batched action dispatches, middleware, store enhancers that accept multiple actions at once, or store enhancers that debounce subscriber notifications.  See https://redux.js.org/docs/faq/Performance.html#performance-update-events .

asturur:
Well the pont is that i do not want to batch everything, is just react i need to debounce a bit while i need the other dispatches to run normally and get notifed immediately.

If you can point me at some resource for just react batching notification i would be glad.

to be honest i thought react-redux is the middleware that couples redux with react.

asturur:
All the resources i see at that link point me in a direction of batching the things at redux level. am i wrong?

```javascript
import { createStore } from 'redux';
import { batchedSubscribe } from 'redux-batched-subscribe';
import debounce from 'lodash.debounce';

const debounceNotify = debounce(notify => notify());
// Note: passing batchedSubscribe as the last argument to createStore requires redux@>=3.1.0
const store = createStore(reducer, intialState, batchedSubscribe(debounceNotify));
```

this example looks like the one that fit most, but i m worried it would debounce all the middleWare queue. while i want just react to be debounced on its own.

markerikson:
@asturur : it's a "bindings layer", not "middleware" (terms matter here).  But yes, it's the coupling between Redux and React.

Can you give more examples about your use case and what the code looks like?  What sort of logic is in your `mapState` functions?  Why do you say you need to debounce "just the React" part, and not the subscriber notifications in general, since it's likely that all the subscribers _are_ instances of connected components?

sirrodgepodge:
@markerikson your point is accurate, most users' subscribers will all be react notifications, however, I think it's perhaps better to do this at the React level as this allows one to still safely subscribe to the store with non-React subscribers.

The alternative would be adding a store decorator on the redux end that inspects subscriber functions to determine if they are React subscribers, and runs logic accordingly.  It seems more appropriate to put this logic in the library connecting React and Redux than to inspect inside of the consumed library.

markerikson:
Per your prior question: using a store enhancer that alters the subscriber notification behavior (by either accepting a batched/array of actions to dispatch at once, or debouncing the notification behavior) would _not_ interfere with middleware behavior.  Middleware form a pipeline that leads up to the core store `dispatch()` method.  Subscriber notifications are the last thing that happens inside of `dispatch()`.  

My response here is that the "right" way to do this is at the Redux store level, primarily with a store enhancer.

asturur:
So maybe i may actually be wrong, and surely you can drive me better.

i have a big state, that can be up to some megabyte of json objects.
I want to sync to computer on the network over this state.

The network part work on the action sequence, because the state is big to send.
Or also i could have a logger on the subscribers to record the actions for later inspections.
My react UI depends not on the actions but just on the final state of the state's data.

So i have side effects that depends on the action sequence, and side effects that depends on the final state of the state.

The UI takes time to render ( a complex widget with many sub components can take 20ms and upper to change ). If a widget has to change on 2 different action in rapid sequence, i would like to be able to relax the `react binding layer` to react a little bit later and react just once.

Both the actions are valid alone, and they can be fired alone or togheter, but would be nice to get one react render for them.

This clarify a bit?

If you have an idea on how to achieve this please share.

markerikson:
@asturur : my initial reaction is that the action sequencing / network behavior should live in middleware (which would see every action flowing through the system), and minimizing UI updates should be done by a combination of a store enhancer that alters the subscription notification behavior and good use of memoized selectors inside of your `mapState` functions to avoid unnecessary re-renders.

asturur:
ok so today i went over this again and while i found a different workaround, i just came here to add other informations

```javascript
trySubscribe() {
    if (!this.unsubscribe) {
      this.unsubscribe = this.parentSub
        ? this.parentSub.addNestedSub(this.onStateChange)
        : this.store.subscribe(this.onStateChange)
 
      this.listeners = createListenerCollection()
    }
  }
```

this trySubscribe from react-redux using  this.store.subscribe over `this.onStateChange`.

So i created my own store-enhancer that replace  store.subscribe, pickup the oncoming listener by a bit, and i'm done.

I wanted just to clarify that  a subscribe function has no clue  on what is incoming.
For my enhancer a listener is just a listener, i have no way to figure out if is the `onStateChange` coming from react redux, and this add to my original point of taking care of this functionality here.

with a store enhancer all subscriptions are created equal, unless i'm missing something, while side effects of listeners may possibly wanted to be handled differently

jimbolla:
The problem with debouncing a high-level component is that it effectively debounces all of its descendant components as well. This can lead to issues such as the textbox cursor issue that manifests when Redux doesn't update the components sooner than React wants it to.

I don't think it'd be wise to add this functionality to React Redux now, but if we make serious design changes as a reaction (hur hur) to fancy stuff enabled by React Fiber, then maybe this is something we could have an extension point for.

markerikson:
Yeah.  On that note, I'm going to close this for now, but we'll keep this in mind when it comes time to deal with the React Fiber-related changes.

asturur:
Well the point of debouncing is debouncing just your top ( or all your top level ) connected components.
All the other are managed by react redux with that parentSub mechanism.

in my tests, debouncing of 1 2 or even just the next tick ms was enough to avoid having more than one render ( virtualdom or dom ) in a chain of 3 rapid fired dispatches.

we are talking of debouncing that stay inside the same screen fram.

@jimbolla do you have any suggestion for something i can try over my app to see if i'm breaking functionalities  that i may not see in the first place?

PR #836, opened 2017-12-07
Export createConnect from index.js

Mosho1:


markerikson:
Pull requests without explanations are unlikely to be merged.

Can you explain why you need this function exposed, and what use case you're trying to solve?

Mosho1:
Sure, I want to use my own implementation of one of the dependencies. There's a comment [here](https://github.com/reactjs/react-redux/blob/master/src/connect/connect.js#L39): "different options opens up some testing and extensibility scenarios", so it has been considered. I think it should be a part of the module exposed from index.js for the same reasons `createProvider` is exposed.

I assume anyone who uses `createConnect` today just imports it from the `react-redux/connect/connect` directly. That's what I'm doing now, though I also have to provide a type definition, which I see aren't provided in this repository so it would have to be added to the definitelyTyped repository.

markerikson:
@jimbolla , any thoughts?

timdorr:
I'm still not sure the point of this. `createConnect` is just a wrapper around our default usage of `connectAdvanced`. It's a configuration. If you need to customize it, you should use `connectAdvanced` directly.

timdorr:
And note that we don't use this export at all. I'm not even sure why it's being exported.

jimbolla:
@Mosho1 Can you explain your use case for using `createConnect`? There's additional support cost in exposing more in the official API; at the minimum more documentation and tests, plus having to support it forever or treat changes as breaking changes. The value in the use case would have to outweigh those costs.

@timdorr It's exposed the way it is so someone like @Mosho1 can experiment with it and show us some way genius new way to use it that we didn't think of.

Mosho1:
@jimbolla I'm using an abstraction over redux for an API that is higher-level and better for my needs. I 
implemented corresponding abstractions over `Provider` and `connect`. My `context[storeKey]` that I pass in to `Provider` is not a redux store, and what I return from `getState` is mutable. So, I had to implement slightly different change detection for `selectorFactory` to be used in my `connect`. I could have used `connectAdvanced` instead, but the rest of the code in `connect/connect.js` would be copy pasted, so I prefer to (and do) use `createConnect` directly.

That might not outweigh the cost, and I agree completely about the consequences of exposing it officially. But it's a useful way to customize for my usage and other might find it useful too.

timdorr:
If we ever make a change to our default configuration, then we would be likely to break your use case unintentionally. There are no explicit tests of this API to guarantee behavior. It would be better for you to copy in the code as-is and "lock in" that configuration, regardless of how the library changes down the line.

I'm going to close this out. I would be interested in seeing that abstraction, though!

PR #857, opened 2018-01-20
Doc update | Update api.md

yabhis:
Changes
rootEl → document.getElementById('root')


timdorr:
Thanks, but this should stay as-is. We're just filling out the existing React API and should keep things as simple as possible here.

Issue #781, opened 2017-09-12
Request: pass own props to `areStatesEqual`


TiddoLangerak:
I want to implement `areStatesEqual` in such a way that it only compares the relevant state for a container. However, this relevant state might be dependent on the container's own props. E.g. suppose my state looks like this:

```js
{
  users : {
    1 : { /* user1 */ },
    2 : { /* user2 */ }
  }
}
```

and I have a container `<User id={userId}>`. Currently, I can't access the id from the `areStatesEqual` function, and as such the best I can do is compare the entire `users` object. This gives a lot more false positives though, and is generally more expensive as well. If I had access to the properties of the container then I could make a much more efficient `areStatesEqual` function.

timdorr:
I think you're looking too high up in the tree. `areStatesEqual` is to allow for alternative state containers, such as immutable.js. I believe you want to use `areStatePropsEqual`, which can utilize props.

TiddoLangerak:
`areStatePropsEqual` is too low in the tree though, since it concerns the component props, not the container's own props. `areStatePropsEqual` is only executed after `mapStateToProps` is executed, so if `mapStateToProps` is expensive then this isn't ideal. The [documentation itself recommends](https://github.com/reactjs/react-redux/blob/master/docs/api.md#optimizing-connect-when-optionspure-is-true) to override `areStatesEqual` when `mapStateToProps` is expensive, but without access to the container's own props it's not always possible to properly implement this function. To give a more complete example (with flow types for better illustration):


```js
/*
const state = {
  users : {
    1 : { ... },
    2 : { ... }
  }
}
*/


type UserComponentProps = {
  user : User
};
function UserComponent({ user } : UserComponentProps) {
   return <div>
     {/* snip */}
  </div>
}

type UserContainerProps = {
  id : number
}
function mapStateToProps(state, ownProps : UserContainerProps) {
   // assume `getUserFromState` is expensive
   return getUserFromState(state, ownProps.id);
}

const UserContainer = connect(mapStateToProps)(UserComponent);
```

Now suppose I have a `<UserContainer id={1}>` component. I only want this component to update when the user with id=1 changes in the store. Currently these are my choices:

- `areStatesEquals: (oldState, newState) => isEqual(oldState.users, newState.users)`. This only prevents updates when other store slices are changed, but it still updates `<UserContainer id={1}>` when user with id=2 is updated.
[...]

PR #865, opened 2018-02-01
pass ownProps to areStatesEqual

z-vr:
We have a use case when we need to use `areStatesEqual` and access `ownProps` at the same time. This PR solves the problem.

```js
function compareStates(paths = [], a, b) {
    return paths.filter((path) => {
        const valueA = get(a, path);
        const valueB = get(b, path);
        return valueA !== valueB;
    });
}

const mapStateToProps = ({
    data: {
        campaign,
    },
}, { data: { id } }) => {
    const campaigns = getEntities(campaign, 'client_id', id);
    return {
        campaigns,
    };
};

export default connect(mapStateToProps, null, null, {
    areStatesEqual(prevState, nextState, { id }) {
        const paths = ['data.campaign.by.client_id.${id}.ids'];
        const differentPaths = compareStates(paths, prevState, nextState);
        const areEqual = differentPaths.length === 0;
        if (!areEqual) {
            console.log('[!] %s are not equal', differentPaths.join());
        }
        return areEqual;
    },
})(Client);
```

### DIFF
--- a/src/connect/selectorFactory.js
+++ b/src/connect/selectorFactory.js
@@ -1,5 +1,5 @@
 import verifySubselectors from './verifySubselectors'
-  
+
 export function impureFinalPropsSelectorFactory(
   mapStateToProps,
   mapDispatchToProps,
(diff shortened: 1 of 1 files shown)


timdorr:
This was already discussed in #781

z-vr:
@timdorr it doesn't make any sense to me why you would not allow to access own properties in this method. The use case is exactly the same as the author you linked, and you're forcing me to run `mapStateToProp` which I do want to avoid. If your users are asking for it why is it so difficult? What's the argument against it?

PR #881, opened 2018-02-17
Fix project links in package.json

simonkberg:
Some links were still referencing the old repo name (`gaearon/react-redux`). Not a huge issue since the old URL redirects, but might as well use the correct URL since some tools display these as-is to end users (eg. yarn via `yarn outdated`) which can potentially be confusing.


markerikson:
Hah, that's been around for a while :)  Nice catch - thanks!

PR #882, opened 2018-02-18
React-redux documentation: Fix links to Redux documentation

afuh:


timdorr:
We're working through issues with switching docs systems right now. These should be able to remain the same, but we definitely need the domain switch. Hopefully we can revert the paths soon!

PR #884, opened 2018-02-22
Probably more careful args check

murashki:
Previous, missed args in mapStateToProps was cause this method called when store updated or component props updated. But whether this behavior is necessary? What is the reason to re-invoket the function if it does not depend on the state or ownProps?


jimbolla:
Libraries such as Reselect can create selector functions with an arity of 0, because they access JS function arguments dynamically.

PR #887, opened 2018-03-02
Update downloads badge to point to graph of downloads over time 📈 instead of duplicating link to npm

cheapsteak:
Hi! I noticed that your "npm" badge and "downloads" badge currently both point to the same page for the project on npmjs.org  

I was wondering if you might prefer having the download badge point to a [download chart](https://npmcharts.com/compare/react-redux?minimal=true) instead?

If not, feel free to close and apologies for the drive-by PR 😄


timdorr:
Thanks, but I'd like to keep it pointed at npmjs.org.

Issue #890, opened 2018-03-05
Discussion: React-Redux and React async rendering


markerikson:
I'll open this up as a general forum for discussing anything related to how React-Redux is going to interact with React's new "Suspense" async rendering capability and time-slicing behavior.

At the moment, we don't really have a good idea of what that means for React-Redux.  We're going to need a better understanding of how React's async rendering behavior works, how this suggested "caching" concept fits into things, and what the final React APIs are going to look like.

lmatteis:
The general React API will change for suspense, but the main concept I believe is that `this.deferSetState({ foo: 'bar' })` will try to render as much as the tree as possible, but only flush the actual changes when all the thrown promises in the tree have completed (`movieFetcher.read()` throws a promise). The actual update could occur after several seconds if the promises are HTTP requests.

If you call `deferSetState` again the old tree is discarded to avoid race-conditions.

There are other bells and whistles, like Placeholder and Loading which allow to have finer-grained access to this process, but this is the overall idea I believe.

In the demo the fetchers seem to use a cache. Throwing an HTTP promise in the render is akin to "start this HTTP operation and re-render me when there's a response", so the cache seems to exist for the render to understand when the response arrived.

It's unclear how this can be used with Redux, and we should probably keep an eye on this PR https://github.com/reactjs/react-redux/pull/856

From a high-level, the cache that render's use to understand whether they need to start an async operation could very well be a redux store. 

I imagine redux should be notified about when such async operations occur:

```js
function MovieDetailPage({ movieId }) {
  const movie = movieDetailFetcher.read(movieId) // throws a promise
  ...
```

Instead, the `connect` HOC could pass to the component the actual fetcher so it knows about the promise:

```js
function MovieDetailPage({ movieId, movieDetailFetcher }) {
  const movie = movieDetailFetcher(movieId) // throws a promise, and let's redux know about it
  ...
```

Just my two cents.

timdorr:
You can throw a Promise both from `render` and from `getDerivedStateFromProps`. The latter looks like it would be a good way to handle things. But it would require us handling side effects explicitly. That's normally done in the middleware, so it's just a matter of this library providing its own "middleware" of sorts and expecting a particular return from middleware (this would be new). 

We would wrap the store.dispatch in a function that checks for a promise return. If it gets one, we store it, fire a setState to trigger an update, and during the render we would throw that promise to React. When it resolves, we would get a re-render from React and should pass through with the resolved state. 

[...]

PR #898, opened 2018-03-09
React 16 experiment: rewrite React-Redux to use new context API

markerikson:
**FOR DISCUSSION ONLY - DO NOT MERGE**

Per discussion in #890 , React-Redux will need to change to support React's upcoming async rendering behavior.

I spent this afternoon experimenting with an update of our current v5 codebase to use the new `React.createContext()` API that will be out in React 16.  To my very pleasant surprise, it actually seems to be working out better than I thought it would.

### Summary of Changes

- Previously, every single connected component instance subscribed to the store separately, as implemented in `connectAdvanced`.  Now, `<Provider>` subscribes to the store, and calls `this.setState({storeState})` in the subscription callback (sorta similar to how the v4 `connect` implementation did things).
- React-Redux calls `React.createContext()` once internally to create a single `ReactReduxContext` provider/consumer pair.  Both `<Provider>` and `connectAdvanced` use that pair.  All references to `contextTypes` have been removed.
- Since the `connectAdvanced` wrappers now directly use the context pair, the entire Subscription concept has been removed
- `connectAdvanced.componentWillReceiveProps` was renamed to `UNSAFE_componentWillReceiveProps`.  (Note that I'm not sure if this is still needed.)
- The selector init/update logic was modified to take the current store state as an argument, rather than closing over the store directly.
- The main call to the memoized selector was moved inside of the new `renderChild` render prop callback.
- Connected components no longer accept `store` as a prop, because they don't/can't subscribe to it anyway.  There could be use cases where this is still desirable in some way (particularly for testing?).  We may need to investigate alternatives, such as exporting a `createReactReduxContext()` function, or possibly allowing `<Provider contextProvider={myContext.Provider}>` somehow.

##### Tests
I made a few quick edits to our unit test suite, primarily changing `ProviderMock` to point to the actual `Provider` class instead.  Right now we've got 25 failing unit tests, but I think that most of them are likely to no longer be relevant (such as "should unsubscribe before unmounting").

##### Main Testing Setup
I cloned and built [`facebook/react@fcc4f52`](https://github.com/facebook/react/tree/fcc4f52cdded62e3b8e95d18e3e6adf17b981eed).  I then used `yarn link` with the built copies of `react` and `react-dom` inside of `build/node_modules`, and used `yarn link` on my local clone of `react-redux`.

I then opened up the local repo for my [Project Mini-Mek sample application](https://github.com/markerikson/project-minimek/tree/532a6fdec1e9d773ec0b7cae33b05431751973c3), and used `yarn link` to load in `react`, `react-dom`, and `react-redux`.

##### Main Testing Observations
When I start and load Project Mini-Mek, the application appears to correctly load and run as expected.  I can click through the tabs, load the sample data, select pilot entries, and edit a pilot entry, with no obvious errors in the console or the UI.

If I view which components are updating and re-rendering with a combination of turning on "Highlight Updates" in the React DevTools, and adding console logging to `connectAdvanced.renderChild()`), I see only the components that I expect to re-render updating.  In particular, the list items in the current HEAD of Project Mini-Mek are not currently using memoized selectors.  I tried optimizing them via the [`makeMapState()` technique](http://blog.isquaredsoftware.com/2017/12/idiomatic-redux-using-reselect-selectors/), and saw that they no longer are re-rendering on every store state update.

##### Other Testing Observations

Before modifying React-Redux v5, I threw together a quick tiny version of the "Provider subscribes" approach separately, and whipped up a tiny Redux counter example in a CRA app.  I then compared behavior between dispatching 1 `"INCREMENT"` action and dispatching 5 `"INCREMENT" actions in a row, in both React click event handlers and non-React DOM element click handlers.

I observed that dispatching 5 actions in a row in a React event handler only caused the Provider and the connected component to both re-render once, as React was batching up its `setState()` calls.  However, dispatching five times in a row from a DOM element click handler caused five separate `Provider -> connect -> component` update cycles, as the `setState()` calls were not batched.

Finally, I tried adding a test key to the Project Mini-Mek Redux store, had a component subscribe to that key, and then created a second Redux store instance with the same key but a different value.  I then had my app render a nested `<Provider store={secondStore}>`, and had a test component dispatch an action.  The nested test component correctly picked up the data from the second store, and the dispatched action updated the second store.

### Conclusion

Given that I've only hacked on this for the last few hours, this actually appears to be working out _way_ better than I had anticipated.  I'm sure there's all kinds of edge cases I haven't run into yet, but this looks like a viable first attempt at a React-Redux v6 that would be compatible with React 16.3+.

@gaearon, @acdlite, @bvaughn , @timdorr, @jimbolla : so, how many ways is this broken right now? :)


markerikson:
Also, while I haven't tried this yet: I'm seriously considering the idea of leveraging the `calculateChangedBits / observedBits` aspect of the new context API to implement as a first-approximation performance optimization.

The idea would be to have `Provider` do a shallow equality check and determine which top-level keys had changed, and return a bitmask from `calculateChangedBits` using a "minimal perfect hash" to determine how the key names map to the bits in the bitmask.  Meanwhile, `connectAdvanced` would need to do something fancy internally to determine which top-level keys its `mapState` function actually accesses.  I'm thinking this could be done by wrapping the `state` value in a Proxy when `mapState` is called, and tracking which keys are accessed.  I've seen at least two or three examples of the "use a proxy to track what's accessed" concept lately, so it seems feasible.  Not sure what the runtime perf benefits or negatives would be on that.

I briefly played with the "minimal perfect hashing" idea in a sandbox a couple days ago, and it seemed that I could get string keys to consistently map to the same output bits, so that seems like it'd be feasible too.

jimbolla:
So the whole need to subscribe every component was in case an intermediate component implements shouldComponentUpdate() and returns false.

markerikson:
Other side note:  it appears that `Context.Provider` and `Context.Consumer` now appear in the React DevTools component tree.  We are _really_ going to need an option to hide those, because the depth of a Redux-connected component tree just grew considerably and it's annoying :(

(Existing open issue: https://github.com/facebook/react-devtools/issues/604 )

timdorr:
I'd check out my PR on React Router: ReactTraining/react-router#5908

Some things of note that I'd recommend doing:

- First, tab size should be 2 spaces. And install Prettier 😛 
- You can [polyfill the new context API](https://github.com/ReactTraining/react-router/blob/6f7e912be28fcfc1871ee9349e48847a12ebaf25/packages/react-router/modules/RouterContext.js) on older Reacts (16.0-16.2) with create-react-context.
- Leave the older context API in place. Less breakage that way. Then just pass the return of that function into your Provider's `value`. Makes it a little easier to implement that way.
- Rebase this on #856 so we can take advantage of those improvements. Less work is needed here as a result too.

markerikson:
I don't think I can meaningfully rebase this on top of 856 - some of the changes I made directly conflict with the ones in 856.  856 is still trying to directly subscribe to the store in `connect`, 898 isn't.  Also, 856 removes `makeSelectorStateful`, I kept it and modified it.

 I don't think I can maintain the old context API in here either, because `connect` isn't getting the store at all.

theKashey:
Are you going to export context adapters? Someone could need the direct access to the “current” store, or reproviding it(and not only me). 

About hashing in areStatesEqual - first you need to “perfect hash” the changes, next you can use the “change masks” from connected components to early reject the changes. 
But - how many collisions you gonna get in in only 32 possible masks?
32 top level keys are good enough, but hashing could reduce it to 6 values in the real case with easy. 
I’ll better keep the track of top level keys(usually they never change), and count and use the order number as a bit mask. Close to zero collisions in result 

As a early rejection itself - I’ve already skipped idea in my [beautiful-react-redux](https://github.com/theKashey/beautiful-react-redux/issues/2), as long I have that proxy underneath. Sounds promising. 
The only 2 concerns - Android(no proxy support, needs polyfill), and the amount of time a could spent in areStatesEqual. 
Especially having in mind changes in this PR, when you are already inside react render loop and early rejection will not save some “React cycles”.

faceyspacey:
Looks awesome. @markerikson I had a hunch you were gonna work on this when I saw your tweets.

I think this is a really good first draft that tries to preserve a lot of the old stuff (i.e. for backwards compatibility), *but I think if we were able to start fresh* and if React were still to require so many little workarounds to beam global state efficiently to individual components, the new Context API has missed the mark.

`react-redux` really should be a use-case where the new Context API shines; and not just making this more performant, but also simplifying the implementation. It's basically black magic to anyone that doesn't spend hours analyzing it, and that shouldn't be the case for something doing relatively little. 

I wonder what the ***start fresh*** approach would look like, if the primary goal was to harness new APIs as much as possible...

----

...after about an hour of tinkering, I've come up with the following basic format, which seems congruent with how the new context api is best put to use. We can bring back required additional functionality once a basic structure is agreed upon. Hopefully somehow this time around thanks to the new context API + `getDerivedStateFromProps` API we don't need as many workarounds.

```js
export default function connectAdvanced(selectorFactory, options) {
    class Inner extends Component {
      constructor(props) {
        super(props)
        this.selector = selectorFactory(props.dispatch, options)
      }

      static getDerivedStateFromProps(nextProps) {
        const { dispatch, state, ...props } = nextProps
        return this.selector(state, props) // returns `null` if no changes
      }

      addExtraProps(props) {/* the usual */}

      render() {
        return createElement(WrappedComponent, this.addExtraProps(this.state)) // <--
      }
    }

    function Connect(props) {
      return (
        <ReactReduxContext.Consumer>
          {({ dispatch, state }) => <Inner dispatch={dispatch} state={state} {...props} />}
        </ReactReduxContext.Consumer>
      )
    }

    Connect.WrappedComponent = WrappedComponent
    return hoistStatics(Connect, WrappedComponent)
  }
}
```
> The hallmark of this approach is 3 components, rather than 2, and a transfer of `context` "state" to component `state`, and ultimately on to the `props` of the `WrappedComponent`.

faceyspacey:
Here's the alternative context API in action:

```js

export default function connectAdvanced(selectorFactory, options) {
    class Connect extends Component {
      static contextConsumers = [ReactReduxContext.Consumer]
    
      static getDerivedStateFromProps(nextProps, [context], prevState) {
        const { dispatch, state } = context
        return selectorFactory(dispatch, options)(state, nextProps, prevState)
      }

      addExtraProps(props) {/* the usual */}

      render() {
        return createElement(WrappedComponent, this.addExtraProps(this.state))
      }
    }

    Connect.WrappedComponent = WrappedComponent
    return hoistStatics(Connect, WrappedComponent)
  }
}
```

Ideally this task is as simple as that, minus things like HMR, etc.

pierr:
This is great

markerikson:
@faceyspacey : yeah, I agree that an approach that requires 3 components is not preferable.  As it is, I'm already annoyed at all the extra `Context.Consumer` entries that are going to show up in the component tree (per my comment earlier).

I'm definitely curious about the performance aspects.  Since we're no longer subscribing in every component, there's certainly fewer actual subscribers to the store.  I haven't implemented any perf optimizations yet other than shuffling around the existing `this.selector` usage, but we could actually put the initial "has the root state changed?' check directly in `<Provider>`'s subscription callback:

```js
store.subscribe( () => {
    const storeState = store.getState();
    if(storeState !== this.state.storeState) {
        this.setState({storeState});
    }
});
```

That would avoid even triggering the consumers if the state hasn't changed.

theKashey:
Btw: so why the context format should be changed? Why it could not hold the “old” store and let components “subscribe” in an old way?
Using context to transfer the state just moves logic from redux to react, but not changing the idea. 
And the all you get - less control, and less speed, as long internally context is far more complex thing than simple redux’s subscribe.

faceyspacey:
@theKashey we have to jump through considerable hoops to make sure subscriptions cascade efficiently. In older versions of react-redux performance was a major issue as a tree would have many unnecessary re-renders as connected parents re-rendered connected children multiple times. So perf isn’t just about doing it outside of React—because is if our work doesn’t parallel the React’s rendering algorithm, we will in fact trigger more renders.

So similarly we would hope that the new context api finds the absolute most performantly and intelligent algorithm to re-render all children using the same context the least number of times, among other optimizations we can’t do outside of React. By relying on React to perform this task, we can insure this stays performant as React evolves, while being able to keep this library minimal and easy to reason about.

markerikson:
@theKashey : because the main concern here is not actually performance, but rather ensuring that the entire React component tree consistently sees the same "current Redux state" as components re-render at slightly different times.  In order to do that, we need to have the current _state_ being passed down from the `<Provider>`, rather than having the _store_ get passed down and each component subscribing to the store separately.  That way, we'll hopefully avoid the "tearing" situation that Andrew has talked about.

If you read the discussion in #890 , you'll see that Dan and Andrew have been talking about actually "reimplementing the Redux store API on top of React", in order to fully make use of React's internal scheduling behavior.  I'm hoping that this approach will prove to be an acceptable alternative to that.

faceyspacey:
@markerikson just went through all of https://github.com/reactjs/react-redux/issues/890 , and it really doesn't seem like the strategy to "reimplement the Redux store API on top of React" makes much sense anymore once you discovered passing down the state as `context`. Basically, between that and what I wrote above, it's painstakingly clear to me exactly how this is *supposed* to work. I don't see how the `storeRef` stuff @gaearon pointed out is actually needed. However, it does meet @gaearon 's stated goal here:

> I want to warn against coming up with something complicated. Our goal is to make React Redux work roughly the way it was conceived three years ago. Not put another super complicated system like we have now on top.

As far as suspense and "tearing"--which I assume means when different parts of the page use stale state--it shouldn't really matter if part of the page is out of sync while something loads, as when putting all state in context it's React Async's job to make sure they use the context state that existed previously. I.e. rather than update a a suspended leaf because the context it depended on updated in the middle.

In general, the suspense stuff seems unrelated, though I think @timdorr came up with a powerful connector concept for global state stores and React Async:
> Here's what I'm thinking: We would wrap the store.dispatch in a function that checks for a Promise return. If it gets one, we store it, fire a setState to trigger an update, and during the render we would throw that promise to React. When it resolves, we would get a re-render from React and should pass through with the resolved state.

However that's basically a different Redux library, not just in regards to built-in promise support, but one that has automatic "loading" state handling. I think there's something to the idea, but no library can sweepingly make the decision for you that you should start showing spinners, i.e. *suspend* all connected components, or suspsend a top level component containing most of your app. There's many variations to this though where middleware could target components via state. I mean basically redux-coupled components can check if a specified state is available, and throw if not, triggering `<Placeholder />`. I see this as something that happens in userland or component libraries, but not the concern of Redux. 

In short, I think what we're trying to do here revolves around the new Context API + `getDerivedStateFromProps` and if used correctly, it's React Async's job to render it all properly.

faceyspacey:
@theKashey the funny thing about your amazing [memoize-state](https://github.com/theKashey/memoize-state) lib is it seems we could get built-in memoization and never have to use reselect again:


```js
class Inner extends Component {
      constructor(props) {
        super(props)
        this.selector = memoize(selectorFactory(props.dispatch, options))
      }

     static getDerivedStateFromProps(nextProps, prevState) {
        const { dispatch, state, ...props } = nextProps
        return this.selector(state, props, prevState)
      }
```

That's how many people first looking at `react-redux` think they can just use it for best performance. Thanks to your lib, we may be able to give them their dream.

markerikson:
@faceyspacey : well, there _are_ more reasons for using selectors than _just_ performance - they're also beneficial for state shape encapsulation, and as a place to put state derivation logic.

Couple other quick thoughts:

- I know I've seen comments that there's still several meaningful environments out there that don't support ES6 Proxies.  Thus far, React-Redux hasn't used anything that couldn't be compiled back to ES5, so bumping that up would be something we'd have to discuss.
- Your `getDerivedStateFromProps` examples appear to have an issue in that they refer to `this.selector`, which won't work because it's a `static` function and therefore doesn't have access to `this.selector`.

faceyspacey:
@markerikson , the one based on the "hypothetical/additional" static context api didn't use `this`, and as commented in the code could be cached. Here's what a cached selector would look like in the current render props context api:


```js
export default function connectAdvanced(selectorFactory, options) {
    class Inner extends Component {
      constructor(props) {
        super(props)
        this.selector = selectorFactory(props.dispatch, options)
      }

      static getDerivedStateFromProps(nextProps, prevState) {
        const { dispatch, state, ...props } = nextProps
        const selector = prevState.selector || selectorFactory(dispatch, options)
        const nextState = selector(state, props, prevState) 

        if (nextState) nextState.selector = selector // no `nextState` will be returned if no changes
        return nextState
      }

      addExtraProps(props) {/* the usual */}

      render() {
        return createElement(WrappedComponent, this.addExtraProps(this.state)) // <-- feels right
      }
    }

    function Connect(props) {
      return (
        <ReactReduxContext.Consumer>
          {({ dispatch, state }) => <Inner dispatch={dispatch} state={state} {...props} />}
        </ReactReduxContext.Consumer>
      )
    }

    Connect.WrappedComponent = WrappedComponent
    return hoistStatics(Connect, WrappedComponent)
  }
}
```

So essentially, on the first run, we use the selector factory, but thereafter, we use the cached version.

faceyspacey:
Here's a demo of the above stuff working with the Redux-First Router demo on codesandbox:

https://codesandbox.io/s/3rjkzjr9p?module=%2Fsrc%2Fconnect%2Findex.js

So you'll notice I used @theKashey 's `memoize-state` library. I primarily did it just to simplify making a quick demo and seeing how this structure would pan out, and guess what? It worked perfectly!

Currently, `connect` receives a single argument that should both map the state + dispatch to props, with no autobinding of actions.

The primary connected components to checkout are:
- `/components/Sidebar.js`
- `/components/Switcher.js`
- `/components/List.js`
- `/components/Video.js`

In Sidebar, I bind `dispatch` myself in the "selector," and in the others I just pass `dispatch` to the component to do in `onClick` handlers.

----
There is only one potentially unrelated problem: when you press the iframe browser's back/next buttons the transition group messes up the animation. I thought this had to do with using React Async (16.4.0-alpha.0911da33). So I went back to the original Redux-First Router codesandbox demo and upgraded the versions there, and didn't see the issue.

In short, there is something different in how the state changes are propagating. It could also have to do with my `<TransitionGroup />` component's internals. It could be an issue that occurs as result of the combination of both the new Context API + React Async. My hunch is it's probably usage that should be deemed *"incorrect"* in my `transition-group`library, which is now outdated.

faceyspacey:
After doing some logging, the memoization aspect unfortunately is not working. @theKashey what you think the issue is? 

Otherwise, that's not really the main thing anyway, as ultimately the goal for this repo will be to preserve the same API etc. For that, "standard" context API + `getDerivedStateFromProps` usage seems to be working as advertised for react-redux's purposes.

theKashey:
@faceyspacey - usually then you want to Map something - you use mapStateToProps, and it should be memoize. If you want to dispatch - mapDispatch, and it could be memoized, but usually - not. If you want to mix them - you use mergeProps. 
Divide and Conquer.

theKashey:
@markerikson  is absolutely right about "structuring" the access using selectors - cascades could "share" memorization, and just beautiful, but sometimes it is not easy to use them - for example in "adapters"
Lets imagine you have some complex structure, and have to extract the "meta data" from it - it is loading, in error, displayable and so on. To make memoization work you have to "select" all "spare parts", which is not a problem, but changes the way you shall write the function. That's the problem - "need of mind shift".
Actually this is why I love redux - it forces you to structure things.

> I know I've seen comments that there's still several meaningful environments out there that don't support ES6 Proxies. 

I've just tested - ES5 polyfill (ie `describe property +get`) is just 5% slower.

What else - even proxies and magic stuff around in production is discussable thing, it is still possible to do another trick - use memoize state in dev mode, to double check that user have mapStateToProps properly memoized. Ie if result got changed, but should not - _console.error('Please check the function')_ (or more [complex thing](https://github.com/theKashey/beautiful-react-redux/blob/master/src/check.js) - "purity check")

ericanderson:
@markerikson I think you could very easily extend this to support multiple stores; Instead of creating one Provider/Context, you create a map of `storeKey`: `{Provider,Context}`.

faceyspacey:
@markerikson you check out the demo? What you think--you think we'll be able to get rid of a lot of what made this library so intimidating in the past?

I'm also curious as to whether this eliminates the additional work regarding @gaearon 's `storeRef` stuff, tearing, etc. Unless there's something I'm not privy to, it seems like it *just works.*

Here's the link again:
https://codesandbox.io/s/64mzx7vjmz?module=%2Fsrc%2Fconnect%2Findex.js

coryb08:
@markerikson re the last point on summary of changes:

> Connected components no longer accept store as a prop, because they don't/can't subscribe to it anyway. There could be use cases where this is still desirable in some way (particularly for testing?). We may need to investigate alternatives, such as exporting a createReactReduxContext() function, or possibly allowing <Provider contextProvider={myContext.Provider}> somehow.

For testing, couldnt we just manually pass `store` to connected components as a prop? or is that way too cumbersome/out of the question..?

markerikson:
@coryb08 : I'm saying that in this PR, that capability no longer exists.  Based on some Twitter feedback, sounds like it's getting some active use, so I'll see if I can re-implement the capability in this PR.

@faceyspacey : uh... what do you mean by "what made this library so intimidating" ?

And yes, the idea behind this PR is that it would hopefully work correctly with async rendering without the tearing issues.  I'm hoping to talk to the React team this week to get some feedback on the approach.

faceyspacey:
@markerikson the library's *implementation* has never been what it should be brother. That's not the fault of the absence of high quality contribution, but rather a result of previous versions of React not being a good impedance match for a global state store. I was curious at first, but now I'm sure of it--the context API makes it so react-redux can be written the way it should always have been written: *less code, less workarounds, easy for the common developer to grok.*

The light is finally shining for `react-redux` my friend. You removed a lot of cruft, but let's take this all the way. I've pointed out and demoed several ideas. I guess I was just looking for some confirmation that you saw them. As you said, you'll be in contact with the React team, hopefully you bring these to the table as well. There's really a specific way this sort of HOC is supposed to look. There's a formula too it. It's presented herein.

theKashey:
Let’s me add 5 cents - react-redux is 
- store transportation service, aka Provider
- Selecting and focusing framework, aka mapStateToProps
- And state changes marshal, aka redux itself 

First part is completely replaced by a new context API. 
Second part always gonna be the same. 

Middle part?

It is not bound to redux, and could make pure context api a bit more usable. 

One can Provide redux State in form of state + Symbol.dispatch, thus making it consumable by any “common render prop”, as everybody consumes context. Thus will me connect method compatible with any data you can pass via context. 
It can spread better practices. Related : https://twitter.com/alexandereardon/status/976296940836044800?s=21

apostolos:
I took @faceyspacey's demo and converted it to a "render prop" version (hardcoded store for now).

Connect: [https://github.com/LWJGL/lwjgl3-www/blob/master/client/store/Connect.jsx](https://github.com/LWJGL/lwjgl3-www/blob/master/client/store/Connect.jsx)
Provider: [https://github.com/LWJGL/lwjgl3-www/blob/master/client/store/Provider.jsx](https://github.com/LWJGL/lwjgl3-www/blob/master/client/store/Provider.jsx)

It works really really well. Not only I'm observing 100% correct behavior -- I threw everything at it: nested connected components, components mounting/unmounting based on higher-level state, etc, things that were incredibly hard to solve in previous versions -- but also the performance is exceptionally good even on a page with a ridiculous amount of connected components:

perf

One advantage of the render-prop version is that connect does not have to carry and merge all the additional props and, of course, you don't have to worry about an additional shallowEqual in sCU.

It's amazing how well it works. Currently running in production here: [https://www.lwjgl.org/customize](https://www.lwjgl.org/customize)

Only thing that still bothers me is that top-level render to update the Context...

faceyspacey:
@apostolos excellent work! It in fact seems better and more natural with a render prop api since the context api is based on it and because memoize results in less work producing selectors. Ultimately i see the vision as this: precompile all selectors and by virtue of accessing them in the child component re-renders occur. So that means the elimination of even having to pass mapStateToProps. And at that point we might as well remove mapDispatch and pass all actions.

As far as the top level provider render: add a child component with SCU and compare the store state prop passed to determine whether to allow updates. Block on its changes but allow other prop changes through.

apostolos:
@faceyspacey I may be completely wrong, but I believe a ReactContext node is already smart enough to not cause re-renders to its children. At least that's what I'm observing.

React Devtools draws the Provider's render call on each state change (that's totally expected & understandable) but immediate unconnected children are not re-rendered. I'm not doing anything special, no PureComponent, no sCU, nothing...

theKashey:
```js
export const ControlledCheckbox = ({ spec }: Props) => (
  <Connect
    state={state => ({
      label: spec.label,
      checked: spec.checked != null && spec.checked(state),
      disabled: spec.disabled != null && spec.disabled(state),
      hidden: spec.hidden != null && spec.hidden(state),
    })}
    actions={(dispatch: Dispatch<*>) => ({
      handleClick: (value: boolean) => dispatch(spec.action(value)),
    })}
  >
```
JFYI: that connected component will not react on props change, absolutely. And I am not sure that manual mapping of all props you get from a parent, is a good idea. And, if it is, props should be an argument if a state function, to 1) make it possible to extract state function 2) bypass memoization.

markerikson:
@faceyspacey, @thekashey, @apostolos : I do appreciate the research and experimentation you're doing, but skimming your discussions, I don't think those are directions we want to go... _yet_.

At the moment, the immediate goal is to update React-Redux to work better with React's upcoming async behavior and the new context API, with minimal changes to our current public API (and preferably minimal changes to the internal implementation as well).  However, this is likely to be an intermediate step that will be insufficient to actually work "properly" with async React rendering, and we'll need a larger set of changes down the road to fully take advantage of async React behavior.  At that point, public API changes will be more on the table.

We'll try to open up some RFC-style discussion issues in the near future to figure out where we eventually need to go, but for now I think most of the changes I'm seeing discussed are out of scope.

faceyspacey:
@apostolos in that case, we're golden. So were you just concerned about the provider itself re-rendering (in the virtual dom)? If it's just that single component, and not *"immediate children"* I'd say we don't have a problem.

@markerikson what is the problem with async rendering? Any more info from the React team? 

The context API is supposed to *just work* and we can use it as advertised (with nothing special on react-redux's part). If there does turn out to be a problem, it seems like its a problem *not with redux + async rendering*, but with context + async rendering in general.

I'm not sure what conditions would make a global redux store any different from your own global ad hoc store propagated via context.

Thinking logically, it would have to do with our external `dispatch` mechanism rather than `setState` used within the `Provider` component itself, as that's the only difference between a redux `Provider` and a regular ad hoc one. 

So following that thread, the issue--if there is one--would be dispatching during component lifecycle handlers, leading to a ***cycle***, as React has no idea what such arbitrary functions are, whereas with `setState` (passed down as a callback from a Provider) they know it will result in a re-rendering of children. 

Ultimately using lifecycle handlers in global state store apps is an anti-pattern that results from mixing other state stores like React Router. So this may be a problem that only exists for poorly designed apps. For example, when using *Redux-First Router*, you never have to use lifecycle handlers to dispatch. None of my apps do. It's ultimately a different far simpler way of development than what's promoted in the mainstream which is "component everything" which entails lifecycle handlers. On a side note, the *component everything* strategy of development attains a level of modularity that is hard with Redux, but also rarely made full use of, making it a subpar option than further abstractions coupled to your Redux store, e.g. a router, controllers, etc. "Redux modules" *with component-level bindings to a given redux module* would provide the best of both approaches.

Anyway if this does turn out to be a problem (for the many participating in the "component everything" + lifecycle anti-patterns), Dan and Andrew's idea that React has to manage the store is likely the only solution--cuz obviously they can then work their low-level magic to batch `dispatches` etc just like they do with `setState`. But that said, given global state stores are such a huge part of the ecosystem, a straightforward API would have to be in order. Perhaps they will come up with something super simple where we wrap `dispatch` in:

```js
const  dispatch = reactAsyncAware(dispatch, providerInstanceRef)
``` 

Thinking about all this, I'm not worried; React Async basically has to go primetime at the same time they have a solution for global state stores. It matters not to us how the dispatching happens, and whoever has to implement it will--presuming redux/mobx are users are considered--have a fun afternoon :)

faceyspacey:
And actually, here it is:

```js
import React, { Component, Children } from 'react'
import Context from './Context'

export default class Provider extends Component {
  constructor(props) {
    super(props)
    this.state = {
      storeState: props.store.getState(),
      dispatch: (...args) => {
        const res = props.store.dispatch(...args)

        const storeState = props.store.getState()
        this.setState({ storeState })

        return res
      }
    }
  }

  render() {
    return (
      <Context.Provider value={this.state}>
        {Children.only(this.props.children)}
      </Context.Provider>
    )
  }
}
```

...So `dispatch` is really just a wrapper for `setState`.

You can see it in the demo "working":
https://codesandbox.io/s/r59r531l04

There is one problem though--you will notice when you change pages, the spinner spins forever--this is because `dispatches` from outside the component tree are no longer subscribed to.

The solution is obviously that we gotta wrap those dispatches similarly, i.e. by having the `providerInstanceRef` escape the component. And ideally there's a way to do this before the initial render--perhaps by providing a `context` object reference we can both share.

faceyspacey:
And here's a naive implementation to escape `this.setState` for now:

https://codesandbox.io/s/jn8l8ly85?module=%2Fsrc%2Fconnect%2FProvider.js

the spinner no longer loads forever, and `dispatch` outside of components also uses the same strategy with `setState` instead of subscriptions.

faceyspacey:
Not in the example code *if you would just look at it*. It gives control of when the dispatch actually occurs to React. Show it to them brother.

It's working in the demo. There's no difference from regular usage of `setState`. Redux just needs a ref to `providerInstance.setState`, which in the demo is jerry-rigged as a proof of concept.

The actions that are high priority are usually idempotent with respect to low priority actions, eg: quick typing in an autocomplete field vs. incoming async data--they are unrelated and won't change each other's states. As mentioned before, any challenging issues related to React Async aren't just a problem for Redux, but for usage of the context API in general. It sounds like React Async brings *userland* considerations regarding idempotency of actions/stateChanges, order, etc.

...it will hopefully turn out to be a very small deal, but not everyone can move so slow brother. Some of us are building long term projects that will be affected, and it would be really nice to know what the secret challenges are sooner than later. But it's becoming clearer thanks to the info you're collecting. Thank you.

markerikson:
@faceyspacey: Look... I've been busy.  I've been at a conference this week, except for the day I spent talking to the React team.  I've got a lot of other things demanding my attention as well.

I've got a decent grasp on some of this React async stuff, but I don't yet understand all the implications, and I definitely don't fully understand exactly how it impacts React-Redux.  I'm still trying to sort through all that myself.

To be a bit blunt, I'm pretty sure you're throwing out a bunch of exposition and sample code without fully understanding the problem (and I say that because _I_ don't fully understand the problem).  I appreciate the enthusiasm, but I think it would be beneficial to hold off a bit until we maintainers actually have a better understanding of what the actual issues are, and we can open up a larger discussion to get the community's involvement in figuring out the best way forward long-term.

faceyspacey:
The example code has been made to look EXACTLY like standard context API usage of setState, so that *I/we don't have to understand the problem.*

Seeing this link https://codesandbox.io/s/jn8l8ly85?module=%2Fsrc%2Fconnect%2FProvider.js will give the React Async team some help in finalizing it, as they've been talking about how challenging it might be, and they might realize they've been spinning their wheels when they see only a very small integration is needed to make Redux usage identical to usage of the new context API :)

theKashey:
The conversation is a bit incomplete without React core members. Could someone bring them here?
We are trying to solve problems we don’t truly understand. And your problems could be something they should have in mind.  And already have as far I know.

markerikson:
@acdlite, @gaearon, @bvaughn : feel free to chime in here :)

@faceyspacey , @theKashey : as I said, we maintainers _have_ been talking with the React team already, both via messages and in person.  We're still trying to actually understand _exactly_ what the problems are ourselves, so that we can actually talk with the community at large about the best ways to solve those problems.  Throwing out notional code snippets and claiming they "solve the problem" is meaningless right now, because if you don't know what the problems actually are, you can't validly claim to be solving them.

To the best of my understanding, these are the problems that React-Redux faces when trying to work with async React:

1) React's time-slicing means that it's possible for other browser events to occur in between slices of update work.  React might have half of a component tree diff calculation done, pause to let the browser handle some events, and something else might happen during that time (like a Redux action being dispatched).  That could cause different parts of the component tree to read different values from the store, which is known as "tearing", rather than all components rendering based on the same store contents in the same diff calculation.

2) Because of time-slicing, React also has the ability to set aside partially-completed tree diffs if a higher priority update occurs in the middle.  It would fully calculate and apply the changes from the higher-priority change (like a textbox keystroke), _then_ go back to the partially-calculated low-pri diff, modify it based on the completed high-pri update, and finish calculating and applying the low-pri diff.  

In other words, React has the ability to re-order queued updates based on priority, but also will start calculating low-pri updates as soon as they're queued.  That is not how Redux works out of the box.  If I dispatch a "DATA_LOADED" action and then a "TEXT_UPDATED" action, the data load causes a new store state right away.  The UI will eventually look "right", but the sequence of calculating and applying the updates was not optimal, because Redux pushed state updates into React in the order they came in.

3) There's also likely to be issues around working with the React Suspense data caching/loading concept.  I'm not yet clear on much of this aspect myself.

So, hopefully that clarifies what the problem space is we're looking at.  Again, I'm certainly looking to get the community involved in helping solve these problems, but jumping off and throwing together "solutions" without actually knowing what you're trying to solve really doesn't help anything.

faceyspacey:
All is understood. Tearing likely is not an issue when you don't have multiple subscriptions as before--unless they haven't ironed out all the kinks with using context in combination with async.  Sure, there may be different things on the page, but it's quick and temporary and fades away naturally.

Anyway, I just noticed a snippet Dan wrote, and the funny thing is he has something extremely similar where a `ref` is shared between React + Redux:

https://github.com/reactjs/react-redux/issues/890#issuecomment-370521609

So basically the latest demo is an implementation of that concept.

Again, you don't really need to understand every detail (though all your points are crystal) to know what React needs to have *full* control.

Basically, React just needs to be in charge of when the final state change happens (in addition to using context to broadcast state in the first place). Likely most Redux apps will be fine if there is occasional small delays until the dispatch actually occurs--they're gonna have to be. About all React can do is notify the 3rd party lib that there is a queued request to make a state change (and its priority, and perhaps every time it "merges" the changes into the component tree), and then another notification when it occurs. And there's probably nothing Redux even can do with the initial notification(s), but I'm sure future libs would find use for it. 

As far as ordering, we're just gonna have to re-redesign our apps so order doesn't matter for when these state changes occur. It's a context/reactAsync problem + a userland problem, not a Redux problem. That said, it looks like it's primarily focused on back-to-back interactions, which might have common characteristics that allow for these state transitions to inherently be idempotent more often than not.

gaearon:
I'm happy to answer specific very targeted questions but reading the whole discussion is a bit beyond what I can do at the moment.

faceyspacey:
@gaearon no questions. In fact, if people have questions, point them here. A lot of work to find the *likely* answers to those questions, without expending your time, is in this thread.

markerikson:
@gaearon : can you confirm my understanding of the issues from my last comment, and if I'm missing any other major concerns?

gaearon:
@markerikson Your last comment sounds right. I encourage you to approach this from suspense perspective first (rather than treating it as thing to reflect on later) because I find it much easier to think about these issues with suspense rather than time slicing. Because suspense operates on "macro" scale (seconds) rather than "micro" scale.

ematipico:
From the changes I can see this 

>// TODO We're losing the ability to add a store as a prop. Not sure there's anything we can do about that.
//Connect.propTypes = contextTypes

Can we define this as breaking change? There's really no way to inject the store as prop? 
Just asking as I relay sometimes on this feature, especially while testing and I don't want to mount the `Provider`  component

markerikson:
@ematipico : that's one of the reasons why this PR is labeled "experimental" :)

I asked for some feedback on Twitter about people's use cases for this a couple weeks ago: https://twitter.com/acemarke/status/972266211064139776

I think it's probably possible to re-implement this using the new context API, I just haven't had time to give it a shot.  

Part of the question is whether we would want `store={store}` as a prop to a component to still only apply to that _one_ component (the way it does now).  To do that, we'd probably need to modify Provider to accept a different `Context.Provider` as a prop, and have the connected component create a unique provider/consumer pair so that only itself gets the updates.

If we choose to have the connected component's `store` prop apply to all of its descendants, then at that point you might as well just wrap the component in a `<Provider store={store}>` and be done with it.

Personally, I'd be okay with having some breaking changes for basic React async compatibility and calling it v6, and then v7 would be a potential re-think of the whole API if needed.

src/components/connectAdvanced.js
```diff
@@ -185,85 +167,71 @@ export default function connectAdvanced(
...
+
+          if(this.selector) {
+            this.selector.run(this.props, storeState);
+          }
+          else {
+              this.initSelector(dispatch, storeState);
+          }
+
+          if (this.selector.error) {
+              // TODO This will unmount the whole tree now that we're throwing in render. Good idea?
```
derekclair: It could be... But, I would think may need to be accompanied by a `componentDidCatch` to gracefully handle the Error.

ematipico: Shouldn't the user be responsible to handle the errors?

elisherer:
@markerikson , I think because of redux's 1st principle (single source of truth), 
we can agree that there should be only one redux store provider, and because of that, we could save the store provided as prop to the `Provider` as an object in the `react-redux` module (as a singleton). then the consumer/connect function could always rely on the saved store in the module (assume initialized) and have the same API without breaking changes (i.e. have `connect` automatically "connect" to the redux store)

timdorr:
@elisherer There are many projects where that's not necessarily true. 

What might be interesting is to mirror the context API and generate a Provider/connect pair. The default ones would be a singleton:
```js
import { createConnector } from 'react-redux'

const ConnectorA = createConnector()
const ConnectorB = createConnector()

import App from './App'
const ConnectedApp = ConnectorA.connect()(App)


import SubApp from './SubApp'
const ConnectedSubApp = ConnectorB.connect()(SubApp)

ReactDOM.render(
  <ConnectorA.Provider store={store}>
    <ConnectorB.Provider store={substore}>
      <App>
        <SubApp />
      </App>
    </ConnectorB.Provider>
  </ConnectorA.Provider>
)
```

elisherer:
@timdorr, that is exactly the case in `react-waterfall`.
The problem I see with this is to enable third parties create automatically connected components. 
In this "new generation", every component would have to get a copy of the store in order to work (or be connected manually).

Hypothetic example:|
```js
//before
import { reduxForm, Field } from 'redux-form';
...
<Field name="firstName"/>
...
export default reduxForm(config)(MyForm);

//after
import { reduxForm, Field } from 'redux-form';
...
<Field name="firstName" store={store} />
...
export default reduxForm(store, config)(MyForm);
```

Creating a pair of `Provider`/`connect` would fit the Context API but would break the redux "automagic".

theKashey:
Bad idea, as long (reusable/third party) component should just establish a connection to the (user defined) store, but could not obtain a reference to it (read - import '../../createStore.js'), as long isolated.
There should be a straight way to connect to the "default" endpoint.

src/components/connectAdvanced.js
```diff
@@ -113,33 +105,25 @@ export default function connectAdvanced(
...
         this.renderCount = 0
-        this.store = props[storeKey] || context[storeKey]
-        this.propsMode = Boolean(props[storeKey])
+        this.storeState = null;
+
+
         this.setWrappedInstance = this.setWrappedInstance.bind(this)
+        this.renderChild = this.renderChild.bind(this);
 
+        // TODO How do we express the invariant of needing a Provider when it's used in render()?
```
timdorr: The Consumer of createContext will take on the default value when it's not under the paired Provider. So, I think just leaving this invariant in place would do.

timdorr: Example: https://codesandbox.io/s/p5o489mym0

markerikson:
I've spent the last couple months prepping for a workshop, which just wrapped up.  I need a bit of down time to try to recuperate, but after that I want to come back and look at both this PR and my starter kit package, and see about pushing things along.

Hypnosphi:
> You can polyfill the new context API on older Reacts (16.0-16.2) with create-react-context.

Actually, the polyfill works with React 15 as well

0rvar:
What is the status here?

markerikson:
@0rvar : see my roundup in #950 .  I haven't had time to come back to the situation since I posted that.

Kalkut:
@markerikson I ended up [implementing the benchmark with screenshots of the performance profile](https://github.com/Kalkut/redux-data-frequency-benchmark).

The performance drop after `dispatch` calls is real but it is not an inefficiency of Redux.

It is due to the fact that behind a call to `dispatch` there is a call to `this.setState`. and `this.setState` is heavy.  I just think that apps and features like the one in the benchmark should not be made with React in the first place. It is preferable to wrap such an app in a React component after having built it in an other way that allows more mutability.

Now using Redux or something else as a data store is totally orthogonal so `dispatch` does not cause any inefficiency per se. It does when coupled with React.

markerikson:
This PR has been superseded by the discussions and changes in #950, #995, and #1000 , so I'm closing it.

PR #908, opened 2018-03-27
Fix dependency on loadsh-es

thomaswelton:
Recent versions of lodash-es were affected by a bug in NPM where by it set the timestamps of files to sometime in the 1970's which causes a problem when trying to create a zip file of the `node_modules`

See the released issues here.
https://github.com/yarnpkg/yarn/issues/5577
https://github.com/lodash/lodash/issues/3714

Lodash was republished with a newer version of NPM which fixes this issue.
This PR update just contains the result of `npm update loads-es`


jimbolla:
Shouldn't package.json be updated as well then?

thomaswelton:
It's not required, the package.json contains `^4.17.5`
If you're running an old version of npm which ignores the lock file then you'll just get the latest minor release of `4.17` which right now would be `4.17.8`
But for new versions of npm you can only get when is in the lock file

timdorr:
Thanks!

PR #918, opened 2018-04-02
Failing test for <StrictMode>

### LINKED ISSUE #897: Fix <StrictMode> warnings
Is it possible to replace componentWillReceiveProps because is deprecated to remove React 16+ warning.

timdorr:
Doesn't fix #897. But lets us know what's going to break in future versions of React:
```
Warning: Unsafe lifecycle methods were found within a strict-mode tree:

    componentWillReceiveProps: Please update the following components to use static getDerivedStateFromProps instead: Connect(Container)

    componentWillUpdate: Please update the following components to use componentDidUpdate instead: Connect(Container)

    Learn more about this warning here:
    https://fb.me/react-strict-mode-warnings
```
If others want to fix this, you can use this branch as a base. Feel free to try it out!

### DIFF
--- a/test/components/connect.spec.js
+++ b/test/components/connect.spec.js
@@ -1218,14 +1218,13 @@ describe('React', () => {
       const store = createStore(() => ({}))
 
       function makeContainer(mapState, mapDispatch, mergeProps) {
-        return React.createElement(
-          @connect(mapState, mapDispatch, mergeProps)
-          class Container extends Component {
-            render() {
-              return <Passthrough />
-            }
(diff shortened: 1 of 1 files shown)


theKashey:
In terms of React-Hot-Loader `componentWillUpdate` could _not_ be replaced by `componentDidRender`, as long RHL needs to being able render "a new" Connected component, not a cached version :(
Something major should be changed.

timdorr:
@theKashey We may be able to use the new `getSnapshotBeforeUpdate` and cDU while keeping `oldListeners` around between the two (or going up to the top module level).

PR #919, opened 2018-04-03
 Remove usages of async-unsafe lifecycle methods

### LINKED ISSUE #897: Fix <StrictMode> warnings

Hypnosphi:
Fixes #897


test/components/connect.spec.js
```diff
@@ -2121,7 +2120,7 @@ describe('React', () => {
 
       @connect(null)
       class Parent extends React.Component {
-        componentWillMount() {
+        UNSAFE_componentWillMount() {
```
Hypnosphi: When I tried to change it to `componentDidMount` it broke the test. Maybe it should be changed to `constructor`, I'm not sure

src/components/connectAdvanced.js
```diff
@@ -248,6 +247,10 @@ export default function connectAdvanced(
 
       render() {
         const selector = this.selector
+
+        // Handle forceUpdate
+        if (!selector.shouldComponentUpdate) selector.run(this.props)
```
timdorr: Shouldn't this be handled by sCU above?

test/components/connect.spec.js
timdorr: What error did you get with cDM? The cWM function is not consequential here anyways; [this is testing for unmount cleanups](https://github.com/reactjs/react-redux/commit/ea53d6fb076359a864a1d543568d951d4b3eab3d). So, I'd be fine with removing it and putting a dispatch outside of the ReactDOM.render.

gaearon:
Btw let us know when you'll need a final review for whether it's async-safe.
I'm not actively watching right now but we could take a look when PR is ready.

src/components/connectAdvanced.js
Hypnosphi: sCU doesn't run when forceUpdating

test/components/connect.spec.js
Hypnosphi: `data` was null

src/components/connectAdvanced.js
```diff
@@ -156,11 +156,10 @@ export default function connectAdvanced(
         if (this.selector.shouldComponentUpdate) this.forceUpdate()
       }
 
-      componentWillReceiveProps(nextProps) {
-        this.selector.run(nextProps)
-      }
-
-      shouldComponentUpdate() {
+      shouldComponentUpdate(nextProps) {
```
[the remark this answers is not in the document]
Hypnosphi: Will try that. Is it called before sCU?

test/components/connect.spec.js
Hypnosphi: I think I'll just change the initial state

src/components/connectAdvanced.js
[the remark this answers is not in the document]
Hypnosphi: Looks like it's not: https://reactjs.org/docs/react-component.html#updating

I can try to move `selector` from field on instance to state field, then it will be available in `static getDerivedStateFromProps`

Hypnosphi:
@gaearon I think it's ready for review

test/components/connect.spec.js
```diff
@@ -2009,7 +2008,7 @@ describe('React', () => {
 
       @connect(null, mapDispatchFactory, mergeParentDispatch)
       class Passthrough extends Component {
-        componentWillUpdate() {
+        componentDIdUpdate() {
```
pke: There is a type here? componentD*I*dUpdate

package.json
```diff
@@ -53,7 +53,8 @@
     "lodash": "^4.17.5",
     "lodash-es": "^4.17.5",
     "loose-envify": "^1.1.0",
-    "prop-types": "^15.6.1"
+    "prop-types": "^15.6.1",
+    "react-lifecycles-compat": "^1.1.4"
```
pke: Why do you need the polyfill when you changed all the deprecated functions?

timdorr: He's using `getDerivedStateFromProps`.

pke: I am using that too without a polyfill. No more warnings (in my own code).

timdorr: It's for use with older versions of React. It's not to deal with deprecation warnings, it's to polyfill those lifecycle methods for React <= 16.2.

pke: Ah right, didn't think it that way around.
That brings me to a question for my own lib. What if I want to support the new and old Context API I guess I could just check if `React.createContext` is a function and then use it if avail?

markerikson: @pke : there's a distinct difference in how you declare and use the old and new context APIs.  See my PR in #898 for an example.

jimbolla: New context API is probably going to require breaking changes to React Redux's public API for supporting multiple stores, making it a major version bump.

test/components/connect.spec.js
Hypnosphi: Will fix & add test

src/components/connectAdvanced.js
```diff
@@ -129,7 +138,9 @@ export default function connectAdvanced(
           `or explicitly pass "${storeKey}" as a prop to "${displayName}".`
         )
 
-        this.initSelector()
+        this.state = {
+          selector: this.createSelector()
```
markerikson: I haven't looked at what StrictMode warns about yet.  Does it complain about instance variables?  Is that why we're putting the selector into state?

Hypnosphi: No it doesn't.

This is because `gDSFP` is static, so one can't access instances from it

markerikson: Ahhh, I see it now.  Got it.

src/components/connectAdvanced.js
```diff
@@ -86,6 +91,11 @@ export default function connectAdvanced(
     [subscriptionKey]: subscriptionShape,
   }
 
+  function getDerivedStateFromProps(nextProps, prevState) {
+    prevState.selector.run(nextProps)
```
gaearon: This looks mutative. What does it actually do?

I'm worried that this won't be safe and then avoiding the warning just gives a false sense of security.

Hypnosphi: It's not pure but it's idempotent.
I can try tomake it pure by moving all the fields from `selector` to  state though.

gaearon: Let's try that!

Hypnosphi: Did that. It required some minor changes in tests: https://github.com/reactjs/react-redux/pull/919/commits/a7ba2981b4e4482830da85b105d86600811c7938#diff-b5fd7a42ea1b8efa382416b4a323003c

It's quite clear why `setState` calls count changed, but as for `mapStateToPropsCalls`, I'm not so confident

markerikson:
I _think_ my changes in #898 supersede #856 .

We might want to consider how these would fit together.

peacechen:
I'm testing with this fork and components subscribed to redux state using `mapStateToProps` don't trigger `componentDidUpdate` on state changes.  Is there something else I need to set up?

Hypnosphi:
@peacechen it works OK on my project. Can you share some parts of your setup (maybe as a reproduction repo), so that I could debug it?

peacechen:
It's a pretty large app.  I'll try to create a small sample project to recreate the behavior.

markerikson:
Yeah, I suspect there's a lot of overlap between the two.  I assume this would probably merit a v6.0 major version?  So, notionally v6 would be "compat with React 16.3+", and v7 would be "super-duper React async rendering compat" .

timdorr:
Anyone else want to second this? I can push a 5.1.0-beta.1 to give it a whirl elsewhere.

ematipico:
I could try to integrate it into a small project of mine and leave some feedback, once the beta is out on npm.

timdorr:
OK, I'm going to merge this in and push a 5.1.0-test.1 release. Please give this a whirl! We can look into some of the BC-breaking stuff next.

timdorr:
Found a big bug with the gDSFP change in 16.4: https://reactjs.org/blog/2018/05/23/react-v-16-4.html#bugfix-for-getderivedstatefromprops

Because it's getting called for any render, it's getting double-called for most updates. That leads to it setting the sCU flag to true and then false on the 2nd call.

I've got to do a deeper dive, but I don't have time at the moment.

timdorr:
Well, I think (hope) it's just an issue with our test suite. So, `react-redux@next` is now on npm to try out.

PR #936, opened 2018-04-24
Simplify object check and eliminate lodash depdency

matthargett:
Mirrors https://github.com/reactjs/redux/pull/2599 . One tweak was to cache the original object prototype instead of calling the function twice.


timdorr:
This doesn't work because we don't have a `isPlainObject.js` locally.

matthargett:
Of course I have to make a dumb mistake on my first contribution to this repo. This is what I get for pushing right before I'm rushing out the door to get my child 😆 I did run lint and test before pushing, I swear.

Fixed.

timdorr:
I swapped out the modified versions with their originals from Redux. Given the relative controversy over replacing lodash's version, I'd rather not stir the pot anymore beyond that 😄

markerikson:
@ajoshi0 : whenever we have a reason to push out another release. This doesn't have any real user-facing impact, so I don't see what the urgency is to get out a release that has this change.

ajoshi0:
@markerikson I just asked because My app bundle has 60+ KB bloat due to lodash, lodash-es which will be removed once you do a release but I will wait thanks

markerikson:
Whatever's going on with your app, that is _not_ due to React-Redux using Lodash-ES.  Please check your own app's build setup.

src/utils/verifyPlainObject.js
```diff
@@ -1,4 +1,4 @@
-import isPlainObject from 'lodash/isPlainObject'
+import isPlainObject from './isPlainObject'
```
lencioni: This looks good to me and I'm definitely a fan of removing the lodash dependency, but I wonder if instead of rolling our own isPlainObject we should bring in https://github.com/jonschlinkert/is-plain-object ?

timdorr: Also, I rejected `is-plain-object` because it uses the `toString()` method, which is relatively expensive.

PR #944, opened 2018-05-04
Add subkey option to connect

theJian:
This PR added an option `subkey` to `connectAdvanced`.

Currently, the `createProvider` accepted the second optional argument `subKey`, It seems like has no real usage for now. Providing the custom subscription key only to `Provider` is not sufficient for using this feature, `connect` components need to know the custom subscription key to access the subscription corresponding to custom store. By adding a new option `subKey`, `connect` can be able to access the specific subscription.

- Add an option `subKey` to `connectAdvanced`
- Add tests to `Provider`
- Delete extra whitespaces / Identing lines in codebase.

### DIFF
--- a/src/components/connectAdvanced.js
+++ b/src/components/connectAdvanced.js
@@ -68,6 +68,9 @@ export default function connectAdvanced(
     // the key of props/context to get the store
     storeKey = 'store',
 
+    // the key of context to get the subscription
+    subKey,
+
     // if true, the wrapped element is exposed by this HOC via the getWrappedInstance() function.
     withRef = false,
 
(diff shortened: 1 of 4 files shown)


theJian:
Btw I would like to know if it's possilbe to add `.editorconfig` and `.prettierrc.json` to this project. For a consistent code style.

timdorr:
I'd argue we can just get rid of `subKey`. It doesn't get you anything. You know the subscription key because it follows a specific pattern. It's just extra complexity without any benefits.

theJian:
@timdorr That's a good point. I will send a pr to remove the `subKey` another day. We can close this one now.

jimbolla:
Yeah I don't like subKey, the sub and the store are supposed to go together. I would've made them a single thing if there was a way to do it that was backwards compatible.

PR #945, opened 2018-05-05
Remove subkey

theJian:
This pr removed `subKey` option from `createProvider`.

The `subKey` isn't really useful because `connect` components doesn't know the custom subscription key.
And according to discussion #944 , removing the `subKey` makes things simpler.

- Remove subkey
- Add tests to `Provider`


timdorr:
Imma just merge these in. Thanks!

PR #971, opened 2018-07-11
update validation of component passed to the function returned by connect()

### LINKED ISSUE #914: connect() does not work with React.forwardRef
Specifically, if I compose `connect` with another HOC that uses `forwardRef`, the result of `forwardRef` (an object) is passed as the `WrappedComponent`, which throws an error claiming it should only be a function.

Example
```js
const MyComponent = React.forwardRef(() => null);
connect()(MyComponent);
```

pgarciacamou:
Fixes #914 

This aims to remove the strict function check and use react-is to validate the component that is passed to the function returned by connect(). I first encountered this issue when I created an HOC that forwarded a reference to the composed component, as discussed in the issue referenced above.

### DIFF
--- a/src/components/connectAdvanced.js
+++ b/src/components/connectAdvanced.js
@@ -1,6 +1,7 @@
 import hoistStatics from 'hoist-non-react-statics'
 import invariant from 'invariant'
 import { Component, createElement } from 'react'
+import * as ReactIs from 'react-is'
 
 import Subscription from '../utils/Subscription'
 import { storeShape, subscriptionShape } from '../utils/PropTypes'
@@ -88,7 +89,7 @@ export default function connectAdvanced(
 
   return function wrapWithConnect(WrappedComponent) {
     invariant(
-      typeof WrappedComponent == 'function',
+      ReactIs.isValidElementType(WrappedComponent),
       `You must pass a component to the function returned by ` +
       `${methodName}. Instead received ${JSON.stringify(WrappedComponent)}`
     )
--- a/package.json
+++ b/package.json
@@ -48,6 +48,7 @@
     "invariant": "^2.2.4",
     "loose-envify": "^1.1.0",
     "prop-types": "^15.6.1",
+    "react-is": "^16.4.1",
     "react-lifecycles-compat": "^3.0.0"
   },
   "devDependencies": {
--- a/test/components/connect.spec.js
+++ b/test/components/connect.spec.js
@@ -1206,6 +1206,16 @@ describe('React', () => {
       )
     })
 
+    it('should not throw an error if a object created with React.forwardRef is passed to the function returned by connect', () => {
+      const ForwardRefComponent = React.forwardRef((props, ref) =>
+        <div {...props} ref={ref} />
+      );
+
+      expect(() => {
+        connect()(ForwardRefComponent)
+      }).not.toThrow()
+    });
+
     it('should throw an error if mapState, mapDispatch, or mergeProps returns anything but a plain object', () => {
       const store = createStore(() => ({}))
 
@@ -1544,7 +1554,7 @@ describe('React', () => {
 
     it('should throw an error if the store is not in the props or context', () => {
       const spy = jest.spyOn(console, 'error').mockImplementation(() => {})
-      
+
       class Container extends Component {
         render() {
           return <Passthrough />


src/components/connectAdvanced.js
```diff
@@ -2,6 +2,7 @@ import hoistStatics from 'hoist-non-react-statics'
 import invariant from 'invariant'
 import { Component, createElement } from 'react'
 import { polyfill } from 'react-lifecycles-compat'
+import * as ReactIs from 'react-is'
```
timdorr: For the purposes of tree shaking, you should import only the `isValidElementType` function.

pgarciacamou: @timdorr I did try, it fails with: 

```bash
src/index.js → dist/react-redux.js...
[!] Error: 'isValidElementType' is not exported by node_modules/react-is/index.js
https://github.com/rollup/rollup/wiki/Troubleshooting#name-is-not-exported-by-module
src/components/connectAdvanced.js (5:9)
3: import { Component, createElement } from 'react'
4: import { polyfill } from 'react-lifecycles-compat'
5: import { isValidElementType } from 'react-is'
            ^
```

pgarciacamou: any clues? here is the package: https://github.com/facebook/react/tree/master/packages/react-is

markerikson: For reference, here's the actual built file from the NPM package: https://unpkg.com/react-is@16.4.1/cjs/react-is.development.js .

The entire built `react-is` package is CJS, so I'm not sure there's any tree-shaking that's going to happen.

timdorr: AFAIK, Rollup and Webpack are working on tree-shaking/DCE for CJS bundles (well, all code, really).

But `react-is` is built exactly like `react`, so I'm not sure why that doesn't work. Maybe our build tools need an upgrade?

pgarciacamou: @timdorr to be more specific, it happens when `npm run build:umd` is executed (consequently `build:umd:min` also fails).

pgarciacamou: @timdorr @markerikson I updated every dependency and the issue is still there. I don't think we are going to be able to do that. You are welcome to try.

screen shot 2018-07-12 at 11 41 42 am

pgarciacamou: Just a comment, by doing that, the tests fail. I only did it because I tried updating a single dependency at the time and it didn't work.

pgarciacamou: @timdorr I don't want this PR to get stale, is there anything you recommend me doing? E.g. maybe creating a PR (or issue) to react-is so that it exports individual modules to enable tree-shaking

cellog: as mentioned in https://github.com/reduxjs/react-redux/issues/914#issuecomment-377406113 it's not yet possible to import specific portions of react-is, but if you fix the build of react-is, it would then become possible

pgarciacamou: @timdorr @cellog ok, here is my attempt https://github.com/facebook/react/pull/13250

timdorr: Cool. Let's see if they can get that out soon enough. I'm not particularly worried about this aging quickly, as this part of the code isn't often touched.

pgarciacamou: I closed it because it was not the expected solution.

markerikson:
Looks like there's some movement in https://github.com/facebook/react/issues/13272 , so maybe we can wait for that to get worked out.

pgarciacamou:
@markerikson 

> So do we still want to wait on react-is to be updated with an ESM build, or is it better to just go with the existing version?

that is a good question, but at this point, I feel like it hasn't moved forward because we should wait. Although no one has literally mentioned it.

It looks like @TrySound is working on the ESM feature. We are rooting for you TrySound!

TrySound:
I do my best. For now you can use namedExports option of commonjs plugin.

pgarciacamou:
I'm waiting on https://github.com/facebook/react/pull/13321 to update this PR. As soon as that get's merged I'll update the code.

TrySound:
I'm waiting this one https://github.com/facebook/react/pull/13356

pgarciacamou:
Thanks, @TrySound!

I didn't mean to make it sound like I was pressuring you, I just wanted to comment to let other developers know that this PR is not stalled.

pgarciacamou:
Rebased.

pgarciacamou:
Closed in favor of https://github.com/reduxjs/react-redux/pull/1000

PR #980, opened 2018-07-22
Fully fix #965

### LINKED ISSUE #965: [5.1.0-test.1] Connected components don't re render after state update
I gave a first try to the new beta and I can see some problems (unless I should make some changes).

The components that are connected to the store via `connect` don't get updated (they don't re render). While debugging I can see that the function `mapStateToProps` gets called correctly (and I can see the changes) after the action _*but*_ the component doesn't re-render hence it doesn't show the new values. 

I hope it could help

cellog:
Success!

Using getDerivedStateFromProps properly, this fix passes all existing tests and is a perfect intermediate version prior to react-redux 6.x and 7.x

Fixes #965


package.json
```diff
@@ -39,15 +39,14 @@
     "coverage": "codecov"
   },
   "peerDependencies": {
-    "react": "^0.14.0 || ^15.0.0-0 || ^16.0.0-0",
+    "react": "^16.4.0-0",
```
timdorr: We're not going to break React compat in this version yet. We need to still support <=16.2, otherwise this is a major revision.

cellog: https://github.com/reduxjs/react-redux/issues/965#issuecomment-406864460 Could you and @markerikson decide what you want and let me know?

cellog: (point of clarification) Because gDSFP changed in React 16.4, the only way to support both is to have parallel implementations, and a runtime check when connectAdvanced is loaded, something like this pseudo-code:

```js
import React from 'react'

const parts = React.version.split('.')
if (parts[0] + 0 < 16) {
  module.exports = require('old/connectAdvanced')
} else if (parts[0] === '16' && parts[1] + 0 <= 3) {
  module.exports = require('new5.1/connectAdvanced')
} else if (parts[0] === '16' && parts[1] + 0 >= 4 || parts[0] + 0 > 16) {
  module.exports = require('new5.2/connectAdvanced')
}
```

Or you could use the 5.1 (current test.1 version) with the polyfill for the older React versions.

timdorr: gDSFP changed in 16.4 to fire more often (specifically on any update, not just ones from props changes), so it should be compatible with 16.3 and react-lifecycle-compat. If it's not, then this still suffers from a buggy or improper implementation.

One thing that might of use is in #856, which makes use of a setState callback that returns null to cancel the update. I might not take that PR entirely verbatim, but the concept may be helpful to the implementation here.

cellog: Because gDSFP has no way of knowing whether an update comes from a state change or from a props change, there is no way to support receiving both props and state (redux changes) that works in both 16.3 and 16.4.

I'm baffled as to why you would think it necessary to explain how the gDSFP API changed as if it wasn't the core of the original bug report this PR and my prior version addresses. Perhaps you're not reading the PR carefully enough?

timdorr: You certainly can know and you've implemented such a check. `this.state.props` is essentially `prevProps`, which you can check against `nextProps` to determine the source.

Or, if you wanted to do it differently, you can add something to the setState (which comes through with `prevState`) to signal that this is a state-based update.

cellog: I'm sorry, but I need to make a pointed remark. Please don't take it personally, I am merely pointing out an inefficiency in our communication which I would love to see remedied so we can actually solve this problem.

I don't believe you've actually taken any substantive time to look at this pull request before dismissing it, because it literally does everything you have just described. For instance, here is the source code to gDSFP:

```js
      static getDerivedStateFromProps(props, state) {
        if (connectOptions.pure && shallowEqual(props, state.props) || state.error) return null
        const nextChildProps = Connect.getChildPropsState(props, state)
        return nextChildProps
      }
```

Note the comparison of props to state.props.

*I* misspoke and did not check to see if the patch worked in 16.0-16.3, which would have saved some time as well.

So, I just tried the pull request in React 16.3, and React 16.0 and the latest React 15. All tests except the strict mode test (which only exists in 16.3+) pass in all React 16 versions. In React 15, all the tests fail because the test renderer isn't working to return the root, and that is with the polyfill. I haven't done the research to see why that is happening, but it is an issue with the tests. I also haven't taken the time yet to see if the thing works in a test project even though all the tests fail.

So the main issue is that the peer dependency in package.json is too restrictive, and can be relaxed to include React 16+, and possibly 15, if the issue with the tests can be determined to be only in the tests.

timdorr:
(Moving this up to the top level in case another commit hides our thread above)

No worries. Text-based communication is always lossy.

If the polyfill works in 16.0-16.2 (where it's actually active; it's not in 16.3), then I'd have confidence in it working in older versions of React as well. 

As for the PR contents itself, I still have concerns about passing props as a hack around the signature of gDSFP (essentially to make it function like cWRP did before). Something there doesn't smell right. At the very least, there's a bug with `this.state.props`, as they're never updated past instantiation.

cellog:
You're right, the fact props aren't updated is an issue, and they should be returned from gDSFP, no idea yet if that will affect test failures. I had to be quick this morning, so I may have tested things without polyfill, or with, but it was working no failures 16.0-.4

I'll have time to breathe a bit later, I can check again.

cellog:
OK, The last commit relaxes the peer dependency, and adds back the storing of props. It would only affect performance on updates, so didn't affect behavior, which is why tests weren't broken. Take a look now?

Storing props in state would normally be a problem, if we were using it to generate derived props, they could be stale. But it is only used to detect changes in props in order to make derived props generation less expensive, so I don't think we will have a problem storing them this way.

src/components/connectAdvanced.js
```diff
@@ -124,22 +99,52 @@ export default function connectAdvanced(
...
+          const nextProps = state.childPropsSelector(state.store.getState(), props)
+          if (nextProps === state.childProps) return null
+          return { childProps: nextProps }
+        } catch (error) {
+          return { error }
+        }
+      }
+
+      static getDerivedStateFromProps(props, state) {
+        if (connectOptions.pure && shallowEqual(props, state.props) || state.error) return null
```
markerikson: Maybe add parentheses around the `pure && shallowEqual()` bit just to clarify the comparison?

src/components/connectAdvanced.js
```diff
@@ -159,11 +164,14 @@ export default function connectAdvanced(
         // dispatching an action in its componentWillMount, we have to re-run the select and maybe
         // re-render.
         this.subscription.trySubscribe()
-        this.runUpdater()
+        this.updateChildPropsFromReduxStore(false)
       }
 
       shouldComponentUpdate(_, nextState) {
-        return nextState.shouldComponentUpdate
+        if (!connectOptions['pure']) {
```
markerikson: Any particular reason for the bracketed field access here?  Pretty sure you should be able to dot-access it - that would only be an issue if `connectOptions` was undefined, which it won't be.

src/components/connectAdvanced.js
```diff
@@ -186,17 +194,35 @@ export default function connectAdvanced(
...
+            })
+          const ret = {
+            storeState: nextState,
+            ...childPropsState
+          }
+          if (notify && childPropsState === null) {
+            this.notifyNestedSubs()
+          }
+          return ret
+        }, () => {
```
markerikson: Could change this to `notify ? this.notifyNestedSubs : noop` .

src/components/connectAdvanced.js
```diff
@@ -217,7 +243,7 @@ export default function connectAdvanced(
       }
 
       onStateChange() {
```
markerikson: If all this does is call `updateChildPropsFromReduxStore()`, we could just pass that to the `Subscription` instead.

markerikson:
I left a few nitpicks stylistically, but code-wise this seems plausible.

I pulled and ran the tests, and it does look like everything's passing.  Sweet!

I'd say it's worth merging and publishing as `5.1.0-test.2` or something.  Let's get it out where people can try it.

@cellog , thanks a lot for collaborating with us on this!  I think we had some nice back-and-forth sharing of ideas here.

Assuming this works, we can start turning our attention to a reworked version of #898 that uses new context.

cellog:
I had the same thought about `onStateChange` yesterday afternoon when I couldn't get to a machine, and of course forgot about it by evening. I took a moment to fix the nitpicks.

I think before we merge, it is important to add some more tests, specifically to check for how more than 1 update affects its performance and its correctness, since I had forgotten to save props in gDSFP and adding it in didn't affect any tests.

However, I have a larger issue I'd like to bring up relating to tests. Is there any reason the repo doesn't use enzyme? Also, do you think it would be feasible to make the test suite run against multiple React versions? In my gut, I feel uncomfortable publishing when every test fails in 15.x because the suite can't properly run on my machine (the issue is that react-test-renderer API changed dramatically at 16.x). If the tests were switched to use enzyme, then the change is as simple as:

```js
import React from 'react'
import { configure } from 'enzyme'
import Adapter14 from 'enzyme-adapter-react-14'
import Adapter15 from 'enzyme-adapter-react-15'
import Adapter16 from 'enzyme-adapter-react-16'

if (+React.version.split('.')[0] === 0) {
  configure({ adapter: new Adapter14 })
} else if (+React.version.split('.')[0] === 15) {
  configure({ adapter: new Adapter15 })
} else {
  configure({ adapter: new Adapter16 })
}
```

If there is no blocker to making this change, I'd like to tackle it.

Also, I use wallaby in my IDE to run the tests instantly (even faster than jest, about .75 seconds to run the whole test suite), could I add my config file (it's generic) to the repo for others who might use it? It doesn't affect anyone who doesn't use wallaby, and could be ignored by npm with `.npmignore`

If you're cool with the test change, I'll open an issue for tracking the switch to enzyme and copy/paste the 2nd portion of this comment

cellog:
P.S. enzyme doesn't support the 16.3 context yet, but it can be easily mocked out

markerikson:
I don't know of any specific reason why we've never pulled in Enzyme.  And yes, I'd love to see more tests added!  We've already seen the benefit of our existing test suite - adding more tests to help clarify expected behavior would be extremely valuable.  (It would also be great if we had some comments explaining why some of these current tests exist, but that can be added later.)

I have no idea about the feasibility of targeting multiple React versions with the tests, but it seems like a neat idea if we can do it.

I do find it a bit ironic that the test suite entirely relies on using `connect` as a decorator, yet we (I?) generally discourage it being used that way.

timdorr:
We haven't pulled in enzyme because we haven't needed it. We would have to rewrite all of our tests to really make use of it. And since the plan is to focus 6.0 on React 16.4+, I don't think we need to worry about targeting multiple React versions. It's a lot of effort that wouldn't have any real advantages.

cellog:
I don't think converting the tests to enzyme will be hard, I did a similar thing in my own project, with 1200 tests. It took me about 1.5 hours, so I don't anticipate this will take much longer :). In any case, it will allow running the current suite against 14 and 15, which currently doesn't happen. I'll make a PR and you can decide if it's worth it.

markerikson:
Yeah, I see both trains of thought here.  On the one hand, 6.0 would presumably obsolete any checks against React <16.5.  But, wouldn't hurt to have that around for the 5.x branch.  And hey, if @cellog wants to contribute that, I'm not going to argue against free help :)

cellog:
Incidentally, I think we will find this implementation leaks when an error is thrown in mapStateToProps, because if the error boundary simply discards the tree, we will have subscribed in the constructor.

So I'd like to move all the subscription stuff to cDM and cDU. I think it may even be dangerous to do it with a setState callback because we can't know if the component was unmounted in the mean time, since an error won't call lifecycle methods.

I'm not sure how this affects existing tests, but we'll find out.

gaearon:
I recommend against converting tests to Enzyme. It's been lagging behind React quite a lot.

cellog:
Oh man, I did not mean to merge these pull requests, I was trying to merge it into a different branch

cellog:
OK I successfully removed the accidental merge of #984 into this PR so they can be considered separately

cellog:
The last commit moves subscription out of the constructor, which is async-unsafe, and into componentDidMount.

We need to do some SSR checks to see if react-redux still works there, since subscriptions will never trigger in this design. Perhaps add a check for SSR somehow and do the sub in constructor if so?

cellog:
fixed nested subs with last commit

src/utils/Subscription.js
```diff
@@ -44,13 +44,37 @@ function createListenerCollection() {
...
     this.unsubscribe = null
     this.listeners = nullListeners
+    this.loaded = false
+    const resolve = resolve => {
+      this.resolve = () => {
+        this.loaded = true
+        resolve()
+      }
+    }
+    this.ready = new Promise(resolve)
```
markerikson: Per your question: I'm not sure if we've ever actually defined what our target JS environment is.  Then again, we've always back-compiled to plain ES5.

What exactly is going on here with this `resolve()` business?

Is there another way to implement this without using a Promise?

Kovensky: You could try using a really minimal fake polyfill, but without being able to use the microtask loop.

```js
const FakePromise = typeof Promise === 'function' ? Promise : class FakePromise {
    constructor (resolver) {
      this._resolved = false
      this._cbs = []
      resolver(this._resolve)
    }

    then (cb) {
      this._cbs.push(cb)
      this._notify()
    }

    _resolve () {
      this._resolved = true
      this._notify()
    }

    _notify () {
      if (!this._resolved) { return }
      const cbs = this._cbs.splice(0)
      for (const cb of cbs) {
        cb()
      }
    }
  }
}
```

src/components/connectAdvanced.js
```diff
@@ -140,7 +146,11 @@ export default function connectAdvanced(
 
       static getDerivedStateFromProps(props, state) {
         let ret = null
-        if (state.lastNotify !== state.notifyNestedSubs) {
+        // this next check only should trigger if gDSFP reacts to state changes
+        // in React 16.3, it doesn't. react-lifecycles-compat matches React 16.3 behavior
+        // if a future version of react-lifecycles-compat DOES trigger on state changes
+        // the isOldReact() call must be removed
+        if (!isOldReact() && state.lastNotify !== state.notifyNestedSubs) {
```
markerikson: No reason to do the string splitting every time this runs.  We can just do it once when `connect()` is called, or even once at the top of the file.

src/components/connectAdvanced.js
```diff
@@ -124,54 +105,126 @@ export default function connectAdvanced(
...
         this.state = {
-          updater: this.createUpdater()
+          ...this.state,
+          ...Connect.getChildPropsState(props, this.state)
+        }
+      }
+
+      static getChildPropsState(props, state) {
+        try {
+          const nextProps = state.childPropsSelector(state.store.getState(), props)
```
markerikson: Reading the store state here is still my biggest concern, async-wise.  But, I guess we can go with it until we get more examples from the React team for what proper patterns look like.

src/components/connectAdvanced.js
```diff
@@ -262,29 +312,31 @@ export default function connectAdvanced(
...
           // listeners, this does mean that the old versions of connected descendants will still be
-          // notified of state changes; however, their onStateChange function is a no-op so this
+          // notified of state changes; however, their updateChildPropsFromReduxStore function is a no-op so this
           // isn't a huge deal.
           let oldListeners = [];
 
-          if (this.subscription) {
-            oldListeners = this.subscription.listeners.get()
-            this.subscription.tryUnsubscribe()
+          if (this.state.subscription) {
```
markerikson: I can't wait to make all this stuff hopefully go away in 6.0.

markerikson:
So, we appear to have a PR that passes all existing tests as-is, other than the tiny tweak to where we're checking for a `store` instance.

I'm going to see if I can start putting together a perf benchmark harness, but this may at least be ready for release as a `5.1.0.test-2`.

src/components/connectAdvanced.js
cellog: duh. fixing :)

cellog:
P.S. I do have one concern. Right now, subscription contains parentSub at creation time, and I think this counts as a side effect. However, I think the best path going forward is to get 6.x out as soon as possible, which will remove any possible side effect subscription issues. The subscription to redux is handled properly since we use setState to trigger re-renders on new actions.

markerikson:
I don't _think_ that's too much of a concern.  If a component is created and then thrown away before mounting, its subscription would ultimately get GC'd, and it hasn't produced any lasting changes outside of itself (as far as I understand things atm).  But yes, totally agree that we ought to start aggressively working on 6.0 ASAP.

markerikson:
After discussion with both @cellog and @timdorr , we're leaning towards _not_ putting out a 5.1 release.  Trying to manage lifecycle behavior across multiple React versions is difficult (and apparently the `react-lifecycles-compat` polyfill [uses the 16.3 behavior, not 16.4](https://github.com/reactjs/react-lifecycles-compat/pull/27) , which doesn't help).

I'm currently trying to work on putting together a [performance benchmark harness](#983) so that we can start comparing comparing behavior between different releases.  From there, I plan to pick up work on #898 , or at least a variation of that approach, and start serious development on a 6.0 release that would probably be 16.5+ only.

cellog:
@deecewan the problem we encountered is that fixing the issue in `gDSFP` is incredibly complex, and results in a build about 4 times slower than 5.0.7 when benchmarked. If you stick to 16.3 or lower, the 5.1.0-test.1 will be stable, but it fails spectacularly on React 16.4

timdorr:
@deecewan We're doing a 6.0 release that will target 16.4+ (technically 16.3 too, but I consider that a broken release), so that's essentially what 5.1.0 will become.

deecewan:
@timdorr is there an issue list for the 6.0 release? anything I can help out with?

cellog:
@deecewan #995 #1000 but especially #983 and the discussion in #890 is also relevant

markerikson:
@deecewan : we don't have a formal issues list for 6.0 yet - we probably ought to start coming up with one.

As @cellog said, right now we've got two different PRs that try to rework `connect` to use `React.createContext()`, with varying approaches.  I'm at React Rally this week, and hope to talk to Brian Vaughn and get his feedback on the two implementations.  From there we can hopefully get a better picture of what 6.0 needs to include.

I've also been trying to put together a perf benchmark harness over in #983 .  I'd certainly appreciate some help trying to flesh that out, and officially add it to the React-Redux repo.

deecewan:
I'm on 16.4.1. No `StrictMode` yet - but I still get warnings in the console that I'm using both `gDSFP` and `cWU` (`cWU` coming from `react-redux`).

cellog:
@deecewan that seems unlikely to be a result of using react-redux 5.0, since the warning from react comes from a single class definition, and react-redux 5.0 connectAdvanced does not define gDSFP. Can you elaborate on the warning (or paste a copy) to help understand what's causing it? Is it this warning?

https://github.com/facebook/react/pull/12419

markerikson:
@deecewan : our end goal is that if you're on React 16.4 and React-Redux 5, you should hopefully be able to do a near drop-in upgrade when we finally have v6 ready.  There will be at least a few API differences, but hopefully they'll be small (things like replacing the `withRef` option with use of React's forwarded refs, etc).

PR #996, opened 2018-08-12
Convert tests to react-testing-library, remove enzyme and test-renderer

cellog:
This drastically simplifies testing against future versions, as the only requirement is ReactDOM.render()

@markerikson


cellog:
(Travis build fails because HEAD doesn't work in React 16.4. The output is expected)

test/components/Provider.spec.js
```diff
@@ -75,47 +81,52 @@ describe('React', () => {
 
     it('should add the store to the child context', () => {
       const store = createStore(() => ({}))
+      store.mine = 'hi'
```
timdorr: I'm not sure this is the best approach for identifying a store. It would be better to provide canary values in the initialState for each store. Setting arbitrary properties is technically possible, but it's not idiomatic Redux and might be confusing to those working with these tests in the future.

test/components/Provider.spec.js
```diff
@@ -5,22 +5,28 @@ import PropTypes from 'prop-types'
 import semver from 'semver'
 import { createStore } from 'redux'
 import { Provider, createProvider, connect } from '../../src/index.js'
-import { TestRenderer, enzyme } from '../getTestDeps.js'
+import * as rtl from 'react-testing-library'
```
timdorr: Can we have explicit imports here (`import { render, fireEvent }`)? I tend to dislike `import * as foo` syntax out of both habit and as a very implicit signal that this is TS code (as it's a standard over there, I've noticed. Obviously it doesn't matter at all in the testing context, but it's mainly about enforcing good patterns for others in future PRs. Folks tend to copy what you do in their code.

test/components/Provider.spec.js
```diff
@@ -5,22 +5,28 @@ import PropTypes from 'prop-types'
...
+import 'jest-dom/extend-expect'
```
timdorr: This should probably be in our test setup file. (Do we have one? We should, if not.)

wallaby.js
```diff
@@ -0,0 +1,23 @@
+module.exports = function(wallaby) {
```
timdorr: I don't want to get into a philosophical debate, but can we please remove this file? I don't want to support proprietary, closed-source software in an OSS project. It's fine that you use it locally, but it's not a requirement to develop for the library nor do I really want to promote them from this project.

markerikson:
@timdorr : I think Greg's comment is that our current HEAD is the `5.1.0-test1` code, and that doesn't work right with 16.4.  Did you mean that we should actually drop 16.3 from the test matrix?

timdorr:
Well, I consider 16.3 a "broken" release anyways, so yeah, that can go too. 

Or we can keep them and try returning to the previous code prior to #919 (which is the last code change anyways).

markerikson:
Yeah.  Tonight I plan to revert the actual code on master to where it was for 5.0.7, then create a `5.x` branch (similar to our existing `4.x` branch).  Then I'm going to sort out what's going on with tests.

markerikson:
@timdorr : when you said "It's not going to get fixed in this version", what "it" are you referring to?  The `<StrictMode>` test?

markerikson:
Closed, superseded by #998 .

test/components/Provider.spec.js
cellog: I'm fixing this in the 16.x PR I'm making. Fully agree, and an easy fix

markerikson: I already took care of this in the #998 branch - I put some canary values in the store as suggested.

Issue #802, opened 2017-10-06
Error handling in connectAdvanced can swallow errors


markerikson:
Reported in Reactiflux chat by user "Bahamut", who was not in a position to file a report himself.

Description:

There was a function being defined inside of a `mapDispatch` function, like:

```js
function mapDispatch(dispatch) {
    return {
        someFunction: () => dispatch({type : "SOME_ACTION", string : `some template literal`})
    };
}
```

There was an error in the template literal, and that threw an error when the `mapDispatch` function was checked by `connectAdvanced()` at [connectAdvanced.js#L15-25](https://github.com/reactjs/react-redux/blob/f1367df6de7fc5aed4e2434d68038a517208f6da/src/components/connectAdvanced.js#L15-L25).  

However, the error was swallowed by the `try/catch` block there, and no visible error was reported in the console.

So, the important issue here appears to be that we have a `try/catch` that swallows errors, and doesn't report them in any way.

Would it be a good idea to use the `warning()` util to log errors here?

hoschi:
We found the same problem with when our selector function had an error (accessing prop 'x' from undefined) which was swallowed.

rick-li:
Same here, it swallows all the run time errors in render, it can waste days debugging it.... :(

markerikson:
@rick-li : I wouldn't expect this to swallow errors from _rendering_, just errors that occur inside of a `mapState` or `mapDispatch` function.

mjniday:
I just had this happen with `mapState` where errors were not surfaced. I was trying to access a property on something inside its `ownProps` which was temporarily undefined. Less certain still about this part, but the browser kept crashing when trying to render that component. Fixing the error in the `mapState` function resolved that issue for me.

Something like...
```js
const mapState = (state, ownProps) => {
  const itemId = ownProps.collection.first().itemId // ownProps.collection.first() was initially undefined
  return {
    itemId
  }
}
```

Seems possible that the error handling here was blowing up the call stack which resulted in the browser dying.

jimbolla:
The error should be rethrown in render() [here](https://github.com/reactjs/react-redux/blob/f1367df6de7fc5aed4e2434d68038a517208f6da/src/components/connectAdvanced.js#L254)

AlexTwine:
This needs to be fix.  I spent a good amount of time trying to figure out why my debugger wasn't getting hit in the mapState function.  The error should show up in the console.

benlakey:
[...]

PR #997, opened 2018-08-13
Use new Context, forwardRef for 6.x (alternate implementation to #995)

cellog:
This PR combines an alternate approach to implementing Context to #995 and also the move to react-testing-library (#996) because testing React context with enzyme isn't yet possible. The PR does the following:

* implements context and subscriptions via React's `React.createContext()`
* adds support for refs directly via `React.forwardRef`
* allows multiple children in the Provider component
* allows changing the store prop in Provider (for hot reloading, or switching out for a new store)
* allow nested Providers, the lowest provider in the tree controls children below it (feature of React context, not react-redux)
* replaces passing store as prop by passing custom context provider to `<Provider>` and context consumer to connected components:

```js
function Comp(props) {
  return <div>{props.hi}</div>
}

const Container = connect(...)(Comp)

const customContext = React.createContext()

function FancyPants() {
  return (
    <Provider store={store1}>
      Primary provider
      <Provider store={store2} context={customContext.Provider}>
        <Container>
          Uses store 1
          <Container>
            Also uses store 1
            <Container consumer={customContext.Consumer}>
              Uses store 2
              <Container /> <!-- this uses store 1 again -->
            </Container>
          </Container
        </Container>
      </Provider>
    </Provider>
  )
}
```
* it reduces code size by almost 25% (drops 80 lines)
* removing subscriptions greatly reduces maintenance complexity
* no usage of `getDerivedStateFromProps`, instead, derived props are memoized in `render()`
* more async-ready because all subscribing is done in `componentDidMount` and `componentDidUpdate` of the `Provider` component, subscriptions are not handled at all in connected components, except to read from React's context consumer. This doesn't address tearing and other async issues that are caused by redux being synchronous in nature. Fixing this will require breaking redux's backwards compatibility, and that's a different discussion.
* probably more hot loader compatible, again because subscriptions are isolated in `Provider` and handled by React context.
* Tests: removes testing in React prior to 16.4, ups dependencies to React 16.4+, adds react-testing-library dependency, removing enzyme and react-test-renderer deps. It preserves the multi-version testing framework, in order to add new React versions for testing as they are released. Several tests are updated to support the backwards-compatibility breaks

Drawbacks to this approach:
* BC break: `withRef` is gone for connected components. The code errors out explaining that the user should simply pass a ref in and it will work. This will cause codebases to break that are relying on withRef. Fortunately, the fix is to remove withRef and update code to use standard reference handling procedures in React. docs need to be updated
* BC break: `store` cannot be passed as a prop to connected components. The code errors if a user attempts to pass store to a connected component. The error explains to use a custom `Provider` (as noted above) to fix this, but will cause breaks in the few programs using this feature. docs need to be updated
* BC break: `storeKey` is irrelevant in `connect()` options. The code errors if a user attempts to pass this option in `connect()`'s options. The error explains to simply use a nested `Provider` or a custom `Provider`. docs need to be updated
* requires 3 wrapper components, `Connect`, the Context consumer, and `PureComponent` wrapper around the component in pure mode (2 in impure mode). This is what makes the simplicity possible. The top-most component retrieves the context containing redux state, calculates derived props, and passes them to the inner component. Rather than try to do all of this before `shouldComponentUpdate`, which leads to tons of complexity and brittle code, we pass it to the underlying component. sCU in `PureComponent` works perfectly for our needs, preventing update if the derivedProps are unchanged.
* in order to support forwardRef, we have to import `react-is`, which does not yet have an esm build, so we gain a little heft in exchange for the reduction in code
* the existing hot loading tests don't work at all, and had to be stripped. The implementation assumes that when hot loading, we may get passed a new store to Provider, which will require unsubscribing from the old and subscribing to the new in componentDidUpdate, but this is not tested in the real world yet. There are tests for this new behavior in `Provider.spec.js`
* because tests didn't work at all, the diff is huge.

@markerikson @timdorr


cellog:
Forgot to mention the Issues this addresses/fixes

This fixes
#914 (implements forwardRef and works with forwardRef element as connected component too)
#802 (no try/catch block - mSP is called directly from render() and is bubbled up naturally)

and addresses
#950 (implements React context)
#890 (uses async-safe subscription to redux in Provider)

Issue #943, opened 2018-05-03
prop-types source is included in dist builds


nkbt:
It would be nice to have source of `prop-types` removed from dist. Especially minified version

This helps: https://www.npmjs.com/package/babel-plugin-transform-react-remove-prop-types

In non-minified build `node_modules/react-redux/dist/react-redux.js`

![image](https://user-images.githubusercontent.com/175264/39558890-ae3d2374-4ed5-11e8-92bf-61bdf4234e8b.png)

In minified build `node_modules/react-redux/dist/react-redux.min.js`

![image](https://user-images.githubusercontent.com/175264/39558900-bb7ea8fa-4ed5-11e8-937a-3123d1ec9b1f.png)

gaearon:
Probably want to throw in https://github.com/oliviertassinari/babel-plugin-transform-react-remove-prop-types in prod mode.

timdorr:
I'm cool with that. I'll check into it for the next release.

nkbt:
Unfortunately `babel-plugin-transform-react-remove-prop-types` does not solve the problem, since it does not remove anything from contextTypes which react-redux uses (and issue is closed now as basically won't fix: https://github.com/oliviertassinari/babel-plugin-transform-react-remove-prop-types/issues/77)

For my library I've done a little hack that works well, but I would be reluctant to recommend it to everyone: 
1. webpack config: https://github.com/in-flux/react-component-router/blob/master/scripts/utils/webpack/common.js#L89-L96
2. actual file: https://github.com/in-flux/react-component-router/blob/master/scripts/utils/webpack/emptyPropTypes.js

timdorr:
This will be fixed once we do our next release.

artem-malko:
Hello @timdorr I know, this is quite old issue, but it should be reopened, as I think. 
Yes, you've removed prop-types from react-redux.min.js, but prop types like:
```
Provider.propTypes = {
  store: PropTypes.shape({
    subscribe: PropTypes.func.isRequired,
    dispatch: PropTypes.func.isRequired,
    getState: PropTypes.func.isRequired
  }),
  context: PropTypes.object,
  children: PropTypes.any
};
```
are still in a file.
Moreover, than I build my app with webpack, all proptypes are in my bundle, cuase they are required in components/Provider.js. And terser or uglifyJs ca not strip it.

What do you  think, if we just wrap prop types to NODE_ENV check like this:
```
if (process.env.NODE_ENV !== 'production') {
[...]

PR #1000, opened 2018-08-14
Use React.createContext() by @cellog

### LINKED ISSUE #914: connect() does not work with React.forwardRef

cellog:
**Supersedes #997, based off of master so no merge conflicts**

**takes into account @timdorr 's request regarding Provider.spec.js not using canary values**
**to see the commit history that led to this, check out #997.**

This fixes
#914 (implements forwardRef and works with forwardRef element as connected component too)
#943 (removes prop-types in production build)
#802 (no try/catch block - mSP is called directly from render() and is bubbled up naturally)

and addresses
#950 (implements React context)
#890 (uses async-safe subscription to redux in Provider)

This PR combines an alternate approach to implementing Context to #995 and also the move to react-testing-library (#996) because testing React context with enzyme isn't yet possible. The PR does the following:

* implements context and subscriptions via React's `React.createContext()`
* adds support for refs directly via `React.forwardRef`
* allows multiple children in the Provider component
* allows changing the store prop in Provider (for hot reloading, or switching out for a new store)
* allow nested Providers, the lowest provider in the tree controls children below it (feature of React context, not react-redux)
* replaces passing store as prop by passing custom context provider to `<Provider>` and context consumer to connected components:

```js
function Comp(props) {
  return <div>{props.hi}</div>
}

const Container = connect(...)(Comp)

const customContext = React.createContext()

function FancyPants() {
  return (
    <Provider store={store1}>
      Primary provider
      <Provider store={store2} context={customContext.Provider}>
        <Container>
          Uses store 1
          <Container>
            Also uses store 1
            <Container consumer={customContext.Consumer}>
              Uses store 2
              <Container /> <!-- this uses store 1 again -->
            </Container>
          </Container
        </Container>
      </Provider>
    </Provider>
  )
}
```
* it reduces code size by almost 25% (drops 80 lines)
* removing subscriptions greatly reduces maintenance complexity
* no usage of `getDerivedStateFromProps`, instead, derived props are memoized in `render()`
* more async-ready because all subscribing is done in `componentDidMount` and `componentDidUpdate` of the `Provider` component, subscriptions are not handled at all in connected components, except to read from React's context consumer. This doesn't address tearing and other async issues that are caused by redux being synchronous in nature. Fixing this will require breaking redux's backwards compatibility, and that's a different discussion.
* probably more hot loader compatible, again because subscriptions are isolated in `Provider` and handled by React context.
* Tests: removes testing in React prior to 16.4, ups dependencies to React 16.4+, adds react-testing-library dependency, removing enzyme and react-test-renderer deps. It preserves the multi-version testing framework, in order to add new React versions for testing as they are released. Several tests are updated to support the backwards-compatibility breaks

Drawbacks to this approach:
* BC break: `withRef` is gone for connected components. The code errors out explaining that the user should simply pass a ref in and it will work. This will cause codebases to break that are relying on withRef. Fortunately, the fix is to remove withRef and update code to use standard reference handling procedures in React. docs need to be updated
* BC break: `store` cannot be passed as a prop to connected components. The code errors if a user attempts to pass store to a connected component. The error explains to use a custom `Provider` (as noted above) to fix this, but will cause breaks in the few programs using this feature. docs need to be updated
* BC break: `storeKey` is irrelevant in `connect()` options. The code errors if a user attempts to pass this option in `connect()`'s options. The error explains to simply use a nested `Provider` or a custom `Provider`. docs need to be updated
* requires 3 wrapper components, `Connect`, the Context consumer, and `PureComponent` wrapper around the component in pure mode (2 in impure mode). This is what makes the simplicity possible. The top-most component retrieves the context containing redux state, calculates derived props, and passes them to the inner component. Rather than try to do all of this before `shouldComponentUpdate`, which leads to tons of complexity and brittle code, we pass it to the underlying component. sCU in `PureComponent` works perfectly for our needs, preventing update if the derivedProps are unchanged.
* in order to support forwardRef, we have to import `react-is`, which does not yet have an esm build, so we gain a little heft in exchange for the reduction in code
* the existing hot loading tests don't work at all, and had to be stripped. The implementation assumes that when hot loading, we may get passed a new store to Provider, which will require unsubscribing from the old and subscribing to the new in componentDidUpdate, but this is not tested in the real world yet. There are tests for this new behavior in `Provider.spec.js`
* because tests didn't work at all, the diff is huge.

@markerikson @timdorr


package.json
```diff
@@ -39,15 +39,15 @@
...
   "peerDependencies": {
-    "react": "^0.14.0 || ^15.0.0-0 || ^16.0.0-0",
+    "react": "^16.4.0-0",
     "redux": "^2.0.0 || ^3.0.0 || ^4.0.0-0"
   },
   "dependencies": {
     "hoist-non-react-statics": "^2.5.5",
     "invariant": "^2.2.4",
     "loose-envify": "^1.1.0",
-    "prop-types": "^15.6.1",
```
timdorr: This needs to remain as a plain dependency for the UMD dev build. 

Let's rip this kind of stuff out of this PR for now. I'd rather keep it to a minimal amount of changes in case there are other PRs merged in in the meantime. We can always do it later. I want to switch us to preset-env and upgrade to Babel 7 anyways.

src/components/Context.js
```diff
@@ -0,0 +1,5 @@
+import React from 'react'
+
+const { Provider, Consumer } = React.createContext(null)
+
+export { Provider, Consumer }
```
timdorr: Can we export the Context as default? That will let us export it from the package by itself, which I'd rather do in case React adds new APIs to the Context (potentially some sort of `.read()`).

src/components/connectAdvanced.js
```diff
@@ -65,30 +43,33 @@ export default function connectAdvanced(
...
 
-  const contextTypes = {
-    [storeKey]: storeShape,
-    [subscriptionKey]: subscriptionShape,
-  }
-  const childContextTypes = {
-    [subscriptionKey]: subscriptionShape,
-  }
+  invariant(storeKey === 'store',
+    'storeKey is deprecated and does not do anything. To use a custom redux store for a single component, ' +
```
timdorr: If it doesn't do anything, it's not deprecated :) I'd just say "The storeKey option has been removed."

src/components/connectAdvanced.js
```diff
@@ -112,190 +106,100 @@ export default function connectAdvanced(
...
-        this.setWrappedInstance = this.setWrappedInstance.bind(this)
-
-        invariant(this.store,
-          `Could not find "${storeKey}" in either the context or props of ` +
-          `"${displayName}". Either wrap the root component in a <Provider>, ` +
-          `or explicitly pass "${storeKey}" as a prop to "${displayName}".`
+      constructor(props) {
+        super(props)
+        invariant(!props[storeKey],
+          'passing redux store in props is deprecated and does not do anything. ' +
```
timdorr: Nit: Capitialization. Also, not "deprecated" if the functionality is gone.

src/components/connectAdvanced.js
```diff
@@ -112,190 +106,100 @@ export default function connectAdvanced(
...
-        }
-      }
+    function forwardRef(props, ref) {
+      return <Connect {...props} forwardRef={ref} />
     }
 
-    return hoistStatics(Connect, WrappedComponent)
+    forwardRef.displayName = Connect.displayName
+    const forwarded = React.forwardRef(forwardRef)
+    forwarded.displayName = Connect.displayName
```
timdorr: Do you need to do this twice? Also, you can just use `displayName` variable that's locally scoped here.

cellog: I removed the first one (commit pending). We need Connect's version, as it puts `Connect()` around the `wrappedComponentName`

package.json
cellog: I'll do this change next.

cellog: Hi, I just checked, and the UMD plain build has PropTypes in it, because we have it as a devDependency, and it is used by rollup to make the development UMD bundle.

In any case, it was trivial to add this, so it will be trivial to add it back in.

markerikson:
I just ran another perf benchmark for 5.0.7, #995, and #1000 , set to 2500 connected components.  Here's the FPS (avg / min-max) and Chrome trace time split results for each one (30-second test runs):

|       | FPS           | Scripting | Painting | Rendering |
|-------|---------------|-----------|----------|-----------|
| 5.0.7 | 28.26 (24-30) | 17326     | 9569     | 2072      |
| #995  | 24.26 (19-28) | 19067     | 8246     | 1767      |
| #1000 | 15.8 (14-17)  | 22401     | 5589     | 1187      |


The implementation of #1000 is conceptually interesting, and I like its relative simplicity (just memoize in render vs splitting across two components), but based on these numbers it seems noticeably slower.

Can anyone else confirm the numbers?  I did fresh builds from both PR branches (including tweaking the Rollup config to set `output.sourcemap : true`, and updating the perf script to copy the sourcemaps to the app build folder to get some better names in the traces).

cellog:
can you link to the perf script/setup you're using again?

cellog:
nvm found it here https://github.com/markerikson/benchmark-react-redux-perf

cellog:
needs a pushed commit with your changes though @markerikson

markerikson:
I'm going to try updating #995 to add in `forwardRef` and stuff so that it's more equivalent to #1000 , then try re-running the comparison.

markerikson:
I pushed the sourcemap usage changes to the perf repo.  For building React-Redux, it's just adding `sourcemap : true` to the `output {}` object in `rollup.config.js`.

I'll throw the average calculation into the perf runner.  It's probably not entirely correct, as the FPS value changes could be happening at different times, and so the "average" should really be calculated using frame timestamps or something, but it's at least _an_ objective comparison.

src/components/connectAdvanced.js
timdorr: Connect doesn't do this, `getDisplayName` does it, which is set to the `displayName` var in this closure. It's a nitpick more than anything, so it's not a necessary change.

package.json
```diff
@@ -39,15 +39,16 @@
     "coverage": "codecov"
   },
   "peerDependencies": {
-    "react": "^0.14.0 || ^15.0.0-0 || ^16.0.0-0",
+    "react": "^16.4.0-0",
     "redux": "^2.0.0 || ^3.0.0 || ^4.0.0-0"
   },
   "dependencies": {
     "hoist-non-react-statics": "^2.5.5",
```
jamesreggio: This will need to be updated to v3.

`React.forwardRef` returns an object, which, without the update to v3, will have its fields hoisted onto the component returned by `connect()`, which will in turn will 'skip' the `connect()` component itself because the hoisted fields make the parent look like the interior of the `forwardRef` invocation.

If none of that makes sense, suffice it to say that an upgrade is needed and an outdated `hoist-non-react-statics` is to blame for weird behavior at the intersection of HOCs and `forwardRef`.

cellog: thanks, next commit updates this issue

timdorr:
The current branch is now published to npm under the `next-1000` tag: https://www.npmjs.com/package/react-redux?activeTab=versions

timdorr:
Oh, and renamed this and #995 so they pair up well in the PR list and browser's tab bar.

markerikson:
I just pushed an update to this branch to use `{forwardRef : true}` instead of `{withRef : "forwardRef"}`, for consistency with #995 .  The `next-1000` tag has been republished (thanks, Tim!).

markerikson:
Hmm. I'd guess a lot of it had to do with not being able to propagate a reference change in `context.store` using legacy context.  In other words, if the `<Provider>` supplied a new store in its context, nested components would never see the new reference because of intervening `shouldComponentUpdate` usages blocking the change (a known flaw in how legacy context worked).

Not only does `createContext` not have that update problem, but we're sticking the store _state_ into context now (and, actually, the store itself too, although we're just grabbing `store.dispatch`).  I think this branch has support for swapping the store?  Don't remember if I added it to the #995 branch.

vladshcherbin:
Since react-router v4.4 is coming with new context, there is [a bug related to react-redux v5](https://github.com/ReactTraining/react-router/issues/6424).

Updating to this pull request fixed it. 🎉

stevemao:
@vladshcherbin is the bug still in v5.1.0?

vladshcherbin:
@stevemao yep

timdorr:
Rebased on master and any conflicts fixed up. Also dealt with some package-lock.json noise I made. Whoops!

markerikson:
Gimme a chance to look things over tomorrow - just got back from the conference.

markerikson:
Reviewing the differences between this and 995 right now.

First question: since we're already forcing a minimum of 16.3 anyway, should we just go straight to 16.6 instead?  I see a couple places where we're using `PureComponent`, and we could replace that with `React.memo()` instead.

Kovensky:
There actually is a really small snag that I encountered with using `React.memo` and the other 16.6 goodies: they'll fail `propTypes` validation for components that expect to receive a component as a prop.

They're not functions, they're special objects ("exotic components" as I'm calling them) so they'll fail `PropTypes.func`.

At worst all that this does is print an angry warning in the console, though.

markerikson:
@Kovensky : hmm.  Interesting point.  Very similar to what we already faced over in #914 / #1062 .  

I'm inclined to say we go ahead with using `React.memo()` anyway.  It's a major version bump, so we've got an excuse, and the ecosystem is going to need to catch up with this being a thing soon anyway.

markerikson:
Suppose if I'm going to review it, I might as well make it an actual review to get my thoughts out there.  Loading...

Kovensky:
```ts
// The PropTypes validator in the Route component does not accept `React.lazy` components
// because their `typeof` is not `'function'` (it's an object handled specially by react fiber).
if (__DEV__ && (Route as React.ComponentType<any>).propTypes !== undefined) {
  delete (Route as React.ComponentType<any>).propTypes!.component
}
```

😇

src/components/connectAdvanced.js
```diff
@@ -100,203 +90,169 @@ export default function connectAdvanced(
...
+        shouldComponentUpdate(nextProps) {
+          return nextProps.derivedProps !== this.props.derivedProps
+        }
+
+        render() {
+          let { forwardRef, derivedProps } = this.props
+          return <WrappedComponent {...derivedProps} ref={forwardRef} />
+        }
+      }
+      PureWrapperRef.propTypes = {
```
markerikson: I don't think we need to bother with proptypes for this internal wrapper.

src/components/connectAdvanced.js
```diff
@@ -100,203 +90,169 @@ export default function connectAdvanced(
 
     const displayName = getDisplayName(wrappedComponentName)
 
+    let PureWrapper
+
+    if (forwardRef) {
+      class PureWrapperRef extends Component {
```
markerikson: Can definitely be replaced with use of `React.memo()`.

src/components/connectAdvanced.js
```diff
@@ -100,203 +90,169 @@ export default function connectAdvanced(
 
     const displayName = getDisplayName(wrappedComponentName)
 
+    let PureWrapper
+
+    if (forwardRef) {
```
markerikson: I'm really not keen on this "if `forwardRef` generate A, else generate B" branches.  I know why they're here, but I think these can be simplified.

src/components/connectAdvanced.js
```diff
@@ -100,203 +90,169 @@ export default function connectAdvanced(
...
-        this.renderCount = 0
-        this.store = props[storeKey] || context[storeKey]
-        this.propsMode = Boolean(props[storeKey])
-        this.setWrappedInstance = this.setWrappedInstance.bind(this)
-
-        invariant(this.store,
-          `Could not find "${storeKey}" in either the context or props of ` +
-          `"${displayName}". Either wrap the root component in a <Provider>, ` +
-          `or explicitly pass "${storeKey}" as a prop to "${displayName}".`
+    const OuterBase = connectOptions.pure ? PureComponent : Component
```
markerikson: More `React.memo()`

src/components/connectAdvanced.js
```diff
@@ -100,203 +90,169 @@ export default function connectAdvanced(
...
+
+    class Connect extends OuterBase {
```
markerikson: Doggonit, I want to use hooks here _so bad_ right now.

src/components/connectAdvanced.js
```diff
@@ -100,203 +90,169 @@ export default function connectAdvanced(
...
+      constructor(props) {
+        super(props)
+        invariant(forwardRef ? !props.props[storeKey] : !props[storeKey],
+          'Passing redux store in props has been removed and does not do anything. ' +
+          'To use a custom redux store for a single component, ' +
+          'create a custom React context with React.createContext() and pass the Provider to react-redux\'s provider ' +
```
markerikson: Error message will need to be updated appropriately.

Also, can we do this _only_ in dev?  Just want to minimize the startup path in prod.

src/components/connectAdvanced.js
```diff
@@ -100,203 +90,169 @@ export default function connectAdvanced(
...
+          return lastDerivedProps
+        }
       }
 
-      getWrappedInstance() {
-        invariant(withRef,
-          `To access the wrapped instance, you need to specify ` +
-          `{ withRef: true } in the options argument of the ${methodName}() call.`
+      renderWrappedComponentWithRef(value) {
+        invariant(value,
```
markerikson: Ditto for this.  I don't know how expensive the call to `invariant()` is, but I don't want to do this in prod.

src/components/connectAdvanced.js
```diff
@@ -100,203 +90,169 @@ export default function connectAdvanced(
...
-        this.componentDidUpdate = undefined
-        this.notifyNestedSubs()
+        return <WrappedComponent {...derivedProps} ref={forwardRef} />
       }
 
-      isSubscribed() {
-        return Boolean(this.subscription) && this.subscription.isSubscribed()
-      }
+      renderWrappedComponent(value) {
+        invariant(value,
```
markerikson: Same as above.

src/components/connectAdvanced.js
```diff
@@ -100,203 +90,169 @@ export default function connectAdvanced(
...
 
       render() {
-        const selector = this.selector
-        selector.shouldComponentUpdate = false
-
-        if (selector.error) {
-          throw selector.error
-        } else {
-          return createElement(WrappedComponent, this.addExtraProps(selector.props))
+        if (this.props.unstable_observedBits) {
```
markerikson: Shouldn't be any need to treat `observedBits` specially - just render it, and it likely is undefined most of the time.

Beyond that, tbh I don't think I want to even bother with that for 6.0.  I want to get this out the door, and we can investigate `observedBits` for an update.

src/components/connectAdvanced.js
```diff
@@ -100,203 +90,169 @@ export default function connectAdvanced(
...
-          if (shouldHandleStateChanges) {
-            this.subscription.trySubscribe()
-            oldListeners.forEach(listener => this.subscription.listeners.subscribe(listener))
-          }
-        }
+    if (forwardRef) {
+      Connect.prototype.renderWrappedComponent = Connect.prototype.renderWrappedComponentWithRef
+      Connect.propTypes = {
+        props: propTypes.object,
+        forwardRef: propTypes.oneOfType([
```
markerikson: If we're doing `forwardRef`, then this isn't even the actual exposed component - I don't think proptypes are necessary in that case.

src/index.js
```diff
@@ -1,5 +1,6 @@
-import Provider, { createProvider } from './components/Provider'
+import Provider from './components/Provider'
 import connectAdvanced from './components/connectAdvanced'
+import Context from './components/Context'
 import connect from './connect/connect'
 
-export { Provider, createProvider, connectAdvanced, connect }
+export { Provider, connectAdvanced, Context, connect }
```
markerikson: If we're going to export it, let's call it `ReactReduxContext` or something like that.

test/components/Provider.spec.js
```diff
@@ -1,47 +1,45 @@
...
           return (
-            <div data-testid="store">
-              {storeKey} - {text}
-            </div>
+            <Consumer>
+              {(value) => {
+                return (
+                  <div>
+                    <div data-testid="store">
+                      {storeKey} - {value && value.store ? value.store.getState().mine : ''}
```
markerikson: Generally not keen on this `.mine` thing in the tests.  Will probably overwrite the test files with the versions from #1000 .

markerikson:
Summarizing the review:

- ^16.6
- We should allow passing an _entire_ context object for customization
- Warning messages should be dev-only
- `PureComponent` should be replaced with `React.memo()`
- Remove mentions of `observedBits`
- Simplify/remove `forwardRef`-specific function declarations
- Probably simplify use of `propTypes`

markerikson:
@cellog told me a couple days ago he's busy with some interviews atm, so I'll tackle this.  I may have a little time to play with this Wednesday evening.  Otherwise, I should be able to put most of my attention to it over this next weekend.  I'll aim to have the PR cleaned up by then.

src/components/connectAdvanced.js
Kovensky: <<resists urge to write work code using 16.7.0-alpha.0>>

src/components/connectAdvanced.js
Kovensky: I thought the point of `invariant` is for it to always be used, even in prod. If it's something you can survive in prod it can be just a `warning`.

cellog: @Kovensky is correct. invariant just ignores itself in production, so these are dev-only warnings

cellog:
> @cellog told me a couple days ago he's busy with some interviews atm, so I'll tackle this. I may have a little time to play with this Wednesday evening. Otherwise, I should be able to put most of my attention to it over this next weekend. I'll aim to have the PR cleaned up by then.

Thank you! I am swamped, and appreciate the effort.

markerikson:
@cellog : Sure!  I obviously don't want to take away any of the credit for working on this.  Sure, it's a collaborative effort, but you've done a lot of work here.  That said, you're busy, I've got tweaks I'd like made, and so I'll go ahead and make them over the next couple days so we can move this forward.

cellog:
I love this, because open source ™ :) keep up the great work. I'll surface for air soonish!

package.json
```diff
@@ -40,17 +40,17 @@
...
-    "hoist-non-react-statics": "^3.0.0",
+    "hoist-non-react-statics": "^3.0.1",
     "invariant": "^2.2.4",
     "loose-envify": "^1.1.0",
-    "prop-types": "^15.6.1",
+    "prop-types": "^15.6.2",
     "react-is": "^16.6.0",
-    "react-lifecycles-compat": "^3.0.0"
+    "shallow-equals": "^1.0.0"
   },
```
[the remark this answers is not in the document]
timdorr: Might have been a side effect of my rebases. If it's gone, it can be removed.

src/index.js
timdorr: Under UMD, this is going to be `ReactRedux.Context`. In ES land, you can rename it yourself if it collides (`import { Context as ReduxContext }`)

test/components/Provider.spec.js
timdorr: Yeah, I had brought this up before too: https://github.com/reduxjs/react-redux/pull/996#discussion_r209653040 I think you had a solution?

src/components/connectAdvanced.js
markerikson: Okay, good to know.

markerikson:
Hmm.  So I'm trying to work on this, but the 16.6 test suite won't run.  Jest fails with this error:

`The --config option requires a JSON string literal, or a file path with a .js or .json extension.`

It sure _looks_ like there's a valid JSON string literal being generated and passed as an argument.  No idea what's going on here.

Fortunately, I can run the tests "manually" from WebStorm, but this seems busted.

markerikson:
I'm looking at this again today, and I'm semi-stuck on the handling of `forwardRef` vs `wrapperProps`.

The current commit has totally alternate code paths for rendering the child component with and without a forward ref.  There's near-duplicate `renderWrappedComponent` and `renderWrappedComponentWithRef` functions, and the logic does:

https://github.com/reduxjs/react-redux/blob/e742caea859d22104e5d40fbf5f3c4ef6b8995dc/src/components/connectAdvanced.js#L234-L235

There's also alternate versions of a `PureWrapper` component for with and without passing down `forwardRef`.

I understand _why_ that's there.  It's because we want to maintain referential integrity of the `wrapperProps` object so that it can be easily checked for changes via a single `===` comparison in that memoization function:

https://github.com/reduxjs/react-redux/blob/e742caea859d22104e5d40fbf5f3c4ef6b8995dc/src/components/connectAdvanced.js#L164-L167

But, the code duplication really bothers me.  And, 99.8% of the time people are going to be using standard `connect`, not `connectAdvanced`, and isn't that what the existing `mapState`-ish selector logic ultimately does anyway?

On the other hand, I could see that if we just went ahead and did `<Connect {...wrapperProps} forwardRef={ref} />` in the forwarding function, and the ref was a callback ref, we could wind up with the combined "wrapper props" actually considered changed on every render.  _If_ `mapState` depends on `ownProps`, we'd be re-running that more often that actually necessary.

Any thoughts?

markerikson:
@cellog : what was the actual performance difference that you saw there?

stevemao:
I'm wondering if https://github.com/facebook/react/issues/13739 is a blocker for perf. @alexreardon @gaearon thoughts?

markerikson:
I don't think it's an actual _blocker_, although yes, anything the React team can do to speed up context updates would be helpful.  What I will say is that I want to start getting people to try this out in actual apps so that we can get feedback on real-world performance, not just the artificial stress-test benchmarks we've been trying to run.

alexreardon:
My use case is fairy extreme as I perform 60 updates per second. There is a
slow down. I can get some numbers on Monday if you like

markerikson:
I've been hacking on this tonight, and I've got something that at least passes the tests after a bunch of cleanup and restructuring.  About to test this against the benchmarks repo.

markerikson:
## Progress Update - Ready for Beta-ish?

I think I've got this branch not just cleaned up, but actually improved quite a bit.  In particular, I've managed to simplify the logic around `pure` and `forwardRef`, _and_ it looks like I've sped up this branch significantly compared to its earlier iterations.  It's still not as fast as 5.0.7, but it's the fastest of the v6 variants we've tried thus far.

**I think this is probably ready for a beta release of some kind so we can start getting real-world feedback.**

### Performance Techniques

The big improvement in these latest commits was due to bringing back a trick from v4: memoizing the actual element reference returned from `Connect.render()`.  It's not particularly documented, but if you return the _exact_ same element reference from a render method as the last time, React will bail out of rendering that child.  Since v4 did all of its real work in `render()`, the end looked like this:

```js
        if (!haveMergedPropsChanged && renderedElement) {
          return renderedElement
        }

        if (withRef) {
          this.renderedElement = createElement(WrappedComponent, {
            ...this.mergedProps,
            ref: 'wrappedInstance'
          })
        } else {
          this.renderedElement = createElement(WrappedComponent,
            this.mergedProps
          )
        }

        return this.renderedElement
```

In v5, that kind of went away, since we do all the memoization work up front, and only call `setState()` once we _know_ the wrapped component needs to re-render.

In v6, we're back to doing more work in `render` again thanks to the use of a `<Context.Consumer>`.  Our previous attempts at handling the "no changes" case involved stuff like `PureComponent` as an extra wrapper for the real wrapped component.

I remembered the memoized element trick while playing with the hooks-ified branch I posted over in #1065 , and was able to reuse that with a handwritten memoization function here.

### Performance Benchmarks

I've updated the benchmarks repo over at https://github.com/reduxjs/react-redux-benchmarks , and ran a new set of benchmarks with version 5.0.7, an earlier commit from this branch (4855c84, which I think was before Tim rebased it), and my new commit 2d06cdf .

The repo currently has three benchmark apps:

- `stockticker`: a flat tree of N stock ticker components, rapidly updating at random
- `tree-view`: a variation of the Redux `tree-view` example app, with N counter components in a seeded-randomized tree, and fake clicks on increment/delete/insert buttons at random
- `twitter-lite`: three lists, with new entries continually inserting into them at random

They're artificial stress tests, but they are at least some kind of objective numbers that we can look at.

Each test is run for 30 seconds twice in Puppeteer.  The first time tries to capture FPS results, the second activates a Chrome trace to capture scripting/rendering/painting times.

In theory, anyone ought to be able to clone the benchmarks repo, build from this branch, drop the UMD build into `$BENCHMARKS_REPO/react-redux-versions`, and roughly replicate these results.

And now, *drumroll please*, the results:

#### `stockticker`
```
+------------------+---------+-----------+-----------+----------+
¦ 5.0.7            ¦ 24.32   ¦ 10325.44  ¦ 14866.93  ¦ 3261.55  ¦
+------------------+---------+-----------+-----------+----------+
¦ 6.0-1000-2d06cdf ¦ 18.40   ¦ 14596.15  ¦ 11756.86  ¦ 2474.33  ¦
+------------------+---------+-----------+-----------+----------+
¦ 6.0-1000-4855c84 ¦ 15.63   ¦ 17865.38  ¦ 9266.14   ¦ 1906.47  ¦
```

#### `tree-view`
```
+------------------+---------+-----------+-----------+----------+
¦ 5.0.7            ¦ 43.41   ¦ 9854.61   ¦ 10257.18  ¦ 704.06   ¦
+------------------+---------+-----------+-----------+----------+
¦ 6.0-1000-2d06cdf ¦ 42.37   ¦ 10476.60  ¦ 12690.35  ¦ 465.57   ¦
+------------------+---------+-----------+-----------+----------+
¦ 6.0-1000-4855c84 ¦ 41.78   ¦ 12019.21  ¦ 12074.83  ¦ 447.20   ¦
```

#### `twitter-lite`
```
+------------------+---------+-----------+-----------+----------+-
¦ 5.0.7            ¦ 40.24   ¦ 21362.26  ¦ 4813.54   ¦ 570.79   ¦ 
+------------------+---------+-----------+-----------+----------+-
¦ 6.0-1000-2d06cdf ¦ 31.90   ¦ 23361.35  ¦ 3659.78   ¦ 473.59   ¦ 
+------------------+---------+-----------+-----------+----------+-
¦ 6.0-1000-4855c84 ¦ 29.39   ¦ 24938.56  ¦ 2910.20   ¦ 419.44   ¦ 
```


### Summary

All of our v6 PRs have been considerably slower than v5 **in these artificial benchmarks**, and I don't think there's a lot we can do about that.  v5 could bail out before React ever got involved, but with v6, React is _always_ involved, and has to traverse the tree to find all of the potential context consumers.

However, this branch is measurably faster than our earlier attempts, and it's time we find out just how well it works in the real world.

Longer term, I'm hoping [the React team will find ways to speed up context updates](https://github.com/facebook/react/issues/13739), which would benefit us considerably.

@timdorr, let's get this published as `react-redux@next` and see what happens! :)

timdorr:
OK, ran things through Prettier (I'll get on top of #1071 after this) and we should be good to go.

markerikson:
Yay!

test/components/connect.spec.js
```diff
@@ -1531,35 +1491,82 @@ describe('React', () => {
...
+            <Decorated context={context} />
+          </ProviderMock>
+        </ProviderMock>
+      )
 
       expect(actualState).toEqual(expectedState)
     })
 
-    it('should throw an error if the store is not in the props or context', () => {
+    it.skip('should throw an error if the store is not in the props or context', () => {
```
jfo84: @markerikson This skip is my only concern in the 6.0.0 work. I can pick up the work if you like

Otherwise, looks great!

markerikson: I'll try to take a look at it tomorrow.  I think I disabled that test when doing some of the earlier work in #995 .  I'll need to review what the behavior is / ought to be.

jfo84: @markerikson @cellog Cool. Let me know. Seems like a simple enough piece of work

And also, the matrices could add back 16.3 and the peer-dep dropped to 16.3 like you guys mentioned

markerikson: 16.3 was effectively broken with respect to `getDerivedStateFromProps`.  Anyone on 16.3 _ought_ to be able to bump to 16.4, so I don't think there's any reason to target it directly.

markerikson: Hmm.  I just pulled master, created a branch, and turned on this test... and it passes.  (Even tried changing the message to verify it fails in that case.)

I guess it was failing at some point in the process and I turned it off, but it works as intended now :)

markerikson: I'm gonna mark this as resolved, and push a change to master that just turns this test on.

jfo84: Sounds good. Glad I could help

PR #1005, opened 2018-08-27
fixed isPlainObject.js

alex89lj:
hi , I can't get the correct result when I use Object.create.  
for example：isPlainObject(Object.create({}))   =>  false


markerikson:
We only use this in a couple places in our codebase, and I can't see any reason why we would need to be concerned with use of `Object.create()` in this case.  Appreciate the PR, but I don't think this is something we need to add.

markerikson:
Because it's not meaningfully needed in this case.

It's used in `verifyPlainObject`, which is called here: https://github.com/reduxjs/react-redux/blob/3e53ff96ed10f71c21346f08823e503df724db35/src/connect/mergeProps.js#L26

and here: https://github.com/reduxjs/react-redux/blob/3e53ff96ed10f71c21346f08823e503df724db35/src/connect/wrapMapToProps.js#L61

Both of those places are exceedingly unlikely to be dealing with an `Object.create()` call.

alex89lj:
Ok.thank you for your reply

PR #1013, opened 2018-09-06
Refactor the for loop with ES6 style

sumleo:
Update the code style to read clearly


timdorr:
Thanks, but this treats arrays as object, so it isn't ideomatic JavaScript.

PR #1015, opened 2018-09-08
Add updated <Provider /> docs

maecapozzi:
I added a new folder structure based on @markerikson's suggestions. Documentation will continue to live under `/docs`, but I've added a folder at `/docs/api` and a `Provider.md` file. 

The goal of these changes is to make the section on the `<Provider />` more readable and approachable for react-redux users. I'm looking for suggestions on grammar, any missing information, and a sanity check on my code examples.


markerikson:
First suggestion: actually save the file with the contents inside and THEN push :)

Issue #1018, opened 2018-09-12
Investigate use of context + observedBits for performance optimization


markerikson:
I've written about this in several places recently, so I might as well consolidate the information as an issue.  I'll copy and paste from the other stuff I've written as appropriate.

Prior discussions and materials for reference:

- For an overview of how `calculateChangedBits` and `unstable_observedBits` work, see the post [Bitmasks and the new React Context API](https://hph.is/coding/bitmasks-react-context)
- A [long Reddit comment I wrote on how React-Redux handles updates, and how bitmasks might help with optimization](https://www.reddit.com/r/reactjs/comments/9ey9co/til_react_context_has_a_secret_observedbits/e5t4z6d/?context=3)
- [My React RFC proposal for making `Context.Provider` accept a `calculateChangedBits` prop](https://github.com/reactjs/rfcs/pull/60)
- Discussion of [hypothetically auto-determining what state fields a component accesses](https://github.com/respond-framework/remixx/issues/1)  (based on [other plans from the "Remixx" project](https://github.com/respond-framework/remixx/blob/75de76d822c4f15f96642bad99ee4d9e369b8e8a/docs/initial-plan.md))

## Summary and Use Case

In React-Redux v5 (the current version), every instance of a connected component is a separate subscriber to the store.  If you have a connected list with 10K connected list items, that's 10,001 separate subscriber callbacks.

Every time an action is dispatched, _every_ subscriber callback is run.  Each connected component checks to see if the root state has changed.  If it has, that component re-runs the `mapState` function it was given, and checks to see if the values it returned have changed.  If any of them have, then it re-renders your real component.

In v6, we're changing it so that there's only one subscriber to the Redux store: the `<Provider>`.  When an action is dispatched, it runs `this.setState({storeState : store.getState()})`, and then the store state is passed down to connected components using the new `React.createContext` API.  When React sees the value given to the `Context.Provider` has changed, it will force all of the associated `Context.Consumer` instances to update, and that's when the connected components will now re-run their `mapState` functions.

[...]

PR #1021, opened 2018-09-16
Use React.createContext() with observed bits

theKashey:
This is extension of https://github.com/reduxjs/react-redux/pull/1000 enhanced with "Observed Bits" specially for #1018

Key findings:
1. It works out of the box!
2. If `changed bits` are zero(nothing changed) context will not propagate anything. Probably this is what we want, but at least some tests are failing due it. I am adding a new prop to connect - `observer`, to let _"observe"_ any changes in state, and setting the first bit in the change always to true. Probably `pure` is better fit for this. Probably we could just remove tests.
3. If `observed` bits are 0, but `mapStateToProps` defined - component is treated as `observer` and will not use observedBits. This is for cases `state => state` (no slice used). Probably it's better scan result in search of `state`, but right now mappedProps are not accessible from `connect`.
3. Normal elements await for changes in bit 2(store), and bit 3(hashing func) changes, rest bits are used to represent changes among _slices_.
4. mapStateToProps is given mocked `state`, observed bits calculated using es5 `getters`. Proxy is not needed here, but it's better to double measure the speed.


theKashey:
How well does this handle the top-level state shape changing at runtime? 
Quite well :) First three bits in the mask are reserved for:
- b1. Any changes. All unpure connects has this bit to be always informed about any changes
- b2. Store updates. I am not sure we need this bit, sounds like it's impossible to change store without changing state share.
- b3. hashingFunction updates. Provider provides this function to the Connect for consistent hash calculations, and __updates this function every time state shape changes__.
> PS: actually I forgot to add shallow check to provider, to actually update hashing function, gonna add these 3 lines right now :)

markerikson:
Awright, ran this through the benchmark suite, and... uh... it performs far _worse_ than #995 or #1000 : 

```
¦ 5.0.7            ¦ 17.76   ¦ 16775.78  ¦ 10057.74  ¦ 1974.30  ¦
¦ 6.0-1000-4855c84 ¦ 15.33   ¦ 18308.03  ¦ 8906.81   ¦ 1719.75  ¦
¦ 6.0-1021-129ad67 ¦ 1.63    ¦ 23771.81  ¦ 1969.22   ¦ 335.66   ¦
¦ 6.0-995-9210282  ¦ 14.61   ¦ 19251.80  ¦ 8157.55   ¦ 1581.22  ¦




¦ 5.0.7            ¦ 58.37   ¦ 20583.79  ¦ 6644.07   ¦ 763.82   ¦ 
¦ 6.0-1000-4855c84 ¦ 55.94   ¦ 22541.77  ¦ 5413.20   ¦ 674.57   ¦ 
¦ 6.0-1021-129ad67 ¦ 48.36   ¦ 23830.79  ¦ 4338.93   ¦ 596.82   ¦ 
¦ 6.0-995-9210282  ¦ 55.17   ¦ 23124.32  ¦ 4980.42   ¦ 648.05   ¦ 
```

Looking at the code, I think I can see a couple possibilities why.

First, this looks like it's going to call `setState()` in `Connect` _every_ time the component re-renders and the state changed in some way:  
https://github.com/reduxjs/react-redux/blob/129ad678ccc36dce74b3e8b0191537c691cdccf0/src/components/connectAdvanced.js#L109-L113 
https://github.com/reduxjs/react-redux/blob/129ad678ccc36dce74b3e8b0191537c691cdccf0/src/components/connectAdvanced.js#L167-L178

Second, it's doing at least a little bit extra work for each memoization check:

https://github.com/reduxjs/react-redux/blob/129ad678ccc36dce74b3e8b0191537c691cdccf0/src/components/connectAdvanced.js#L209-L218

We'd have to dig further and see exactly where the bottleneck is here, I guess.

@theKashey , can you clarify why that component is calling `this.props.setObservedBits()` like that?

theKashey:
Actually - this is quite strange, but both examples are using flat arrays as a store. The logic behind my change not working for arrays at all. Probably slow down was driven my double render of components.

PS: I could not run benchmark - package installation failing. Working only with yarn.

markerikson:
Yeah, you're right about the array state shape being unusual.  

We ought to rework the benchmarks so they use something that looks a bit more typical, even if it's just splitting the root state into four different slices or something.

markerikson:
Tried to do some perf investigations yesterday.  It looks like the calculation of `observedBits` in every call to the memoizer is relatively expensive, although I'm kinda struggling to figure out what else is slowing things down if anything.

I think what I'd like to see for now is a new `connectAdvanced` option that basically says "I confirm this component's data access patterns don't change over time", so that we can precalculate the `observedBits` in the constructor of `Connect`.  

Normally we wouldn't have access to the store state in the constructor there.  However, 16.5 now has `Context.unstable_read()`, which means we can imperatively pull the value out of the context in the constructor, and use that to calculate `observedBits` once on creation.

theKashey:
`Context.unstable_read` will solve half of perf issues, but I dont think it's possible to precalculate used slices. To be more concrete - it is possible if you are using `reselect` - first select, next combine, but with _normal_ code, where select and combine are used together - it's not stable.

Right now realization of _observing_ is not quite cool - two function calls. Will try to improve it.

__!__ What are you thinking about:
- enabling this feature only for modern browsers, with Proxy support, and disabling for the rest (IE11, React Native)
- adding a special function, with _predeclared_ slice usage. Basically it's quite easy to store this information in reselect stats and calculate once on selector creation.

theKashey:
Updated with faster key tracker, I've spiked in https://github.com/theKashey/with-known-usage, and `Context.unstable_read`, which, anyway, is not yet supported :)
@markerikson - could you remeasure speed?

theKashey:
https://github.com/facebook/react/issues/13739 - changing `context` information itself causes whole tree re-render, as long you __render__ parent component.
Probably we have to have a `propagation blocker` next to Provider, or wait for a better solution.

PS: Actually we could use observedBits even with "old" redux.

timdorr:
#1000 merged in, so I'm going to close this out. If we want to try more experiments with bitmasks, we should probably start over.

PR #1033, opened 2018-10-06
Remove duplicated page titles, fix sidebar menu

BTMPL:


BTMPL:
Updated the PR a bit more to generate proper sidebar on API page:

image

PR #1039, opened 2018-10-08
Fix link to getting started doc

pyliaorachel:


markerikson:
Thanks!  That said, we really ought to point the README to the new docs site instead :)

PR #1052, opened 2018-10-18
Declare PropTypes as external

DominicTobias:
Hi, I noticed that you're bundling in the prop-types library into the package which increases bundle size and seems unnecessary?


TrySound:
It's only related to umd. Does bundle size matter?

Issue #1055, opened 2018-10-21
API example error related to ownProps


carloluis:
The API doc. section: [The arity of mapStateToProps and mapDispatchToProps determines whether they receive ownProps](https://github.com/reduxjs/react-redux/blob/master/docs/api.md#the-arity-of-mapstatetoprops-and-mapdispatchtoprops-determines-whether-they-receive-ownprops) has some code examples, but the `console.log` hint on the `ownProps` in second code excerpt is wrong. It says undefined instead of the empty object used as default value for the parameter.

I'll create a PR with this minor update/fix:

```diff
const mapStateToProps = (state, ownProps = {}) => {
  console.log(state); // state
-  console.log(ownProps); // undefined
+  console.log(ownProps); // {}
}
```

markerikson:
Fixed by #1056 .

PR #1056, opened 2018-10-21
API Doc code excerpt: log message hint fix for ownProps on map function

carloluis:
- mapStateToProps with default value for `ownProps` parameter
- hence `ownProps` always be the default value

Related issue #1055


markerikson:
Thanks!

PR #1062, opened 2018-10-25
(feat): support react 16.6

JoviDeCroock:
Tests are provided aswell, i hope i did this the right way :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ env:
   - REACT=16.3
   - REACT=16.4
   - REACT=16.5
+  - REACT=16.6
 sudo: false
 script:
   - npm run lint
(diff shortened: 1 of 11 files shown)


JoviDeCroock:
Wanted to include a react 16.6 test but i see that does not work :/

timdorr:
#914 is related here.

I'll check this out locally in a bit and see what's up. I believe what you've got is correct FWIW.

JoviDeCroock:
> #914 is related here.
> 
> I'll check this out locally in a bit and see what's up. I believe what you've got is correct FWIW.

It works when i test it locally, but yes i'd like to run that test only in React 16.6 which as far as i see is not possible with current config.

JoviDeCroock:
Will add the if when I get home :) Thanks for reviewing already, anything else needs changing?

timdorr:
Took care of that on the web for you. Looks like the only thing is you used an outdated version of npm, but I'll fix that up after the fact when I get to my laptop.

JoviDeCroock:
Oh, thanks for telling me. I mostly use Yarn and Yarn Pnp atm so didn't bring that in mind

timdorr:
Thanks for putting this together!

PR #1065, opened 2018-10-26
Use React Hooks for `connect`

markerikson:
This PR builds on the previous React-Redux version 6 preview work from #898, #995, aand #1000 .  As with the previous couple PRs, the major changes are:

- Switching from legacy context to `createContext()`
- Switching from `withRef` to `forwardRef`
- Removing the ability to pass a Redux store as a prop to a connected component

Our existing v6 PRs both passed our unit test suite, but the internal implementations were ugly.  Using `createContext` required that we split the wrapper component into two pieces: an outer wrapper to render the context consumer, and an inner wrapper to receive the store state, process it, and render the real wrapped component.

With the announcement of [React Hooks](https://reactjs.org/docs/hooks-intro.html), there's now another way to access the store state from context: via the `useContext()` hook.

I've taken my existing branch from #995, and rewritten the wrapper component to be a function component that uses hooks.  The implementation is _dramatically_ simpler.  v5, #995, and #1000 all have around 140 lines of complicated class-based component logic.  In this PR, the wrapper component is simply:

```js
    function createChildSelector(store) {
      return selectorFactory(store.dispatch, selectorFactoryOptions)
    }

    function ConnectFunction(props) {
      const {context, forwardRef, ...wrapperProps} = props
      const contextToRead = context || defaultContext

      const {store, storeState} = useContext(contextToRead)
      const childPropsSelector = useMemo(() =>  createChildSelector(store), [store])

      const childProps = childPropsSelector(storeState, wrapperProps)

      const renderedChild = useMemo(() => {
        return <WrappedComponent {...childProps} ref={forwardRef} />
      }, [childProps, forwardRef])

      return renderedChild
    }
```

I ran a test build through our benchmark suite just to see how it compares.  While this build was still slower than v5, particularly in the `stockticker` artificial stress test, it's slightly faster than either of the existing v6 PRs.  

I think this PR completely supersedes both #995 and #1000 , and is our best candidate approach for React-Redux v6.

It does rely on an alpha build of React (currently `16.7.0-alpha.0` ), and so we would have to leave it as an alpha in turn until the Hooks API is finalized and React 16.7 has gone final.  

I'd like to publish this as a new alpha tag on NPM, and get people to start playing with it so that we can find out how it performs in real applications.


markerikson:
@cellog : I did actually have the opportunity to talk to most of the React team today (Andrew, Sebastian, Dan, and Sophie).  They were nice enough to listen to my questions and respond.  I'm not going to say I entirely got answers for everything, but we did talk.

Unfortunately, the day has also been a blur, and I don't remember most of the details at this point.  Sorry :(  Maybe @gaearon , @sophiebits, or @acdlite can pop in and remind me what the rough plans are there.

There's really two related issues around performance here: how fast React can traverse the tree and find consumers (per https://github.com/facebook/react/issues/13739 ), and then React potentially skipping running updates for consumers based on `observedBits` (see #1018 for notional ideas).

I think we ought to focus on just getting v6 ready as-is for now, and let the `observedBits` stuff kinda sit in the background for later.  

I'll also note that I'm not entirely sure just how useful the `stockticker` benchmark actually is, on the grounds that A) it's a completely flat component tree, and B) the speed of the updates.  It's certainly _an_ objective number we can look at, I just don't know how meaningful the results are.

alexreardon:
This change looks super impressive!!!

This diff is incredible
https://github.com/reduxjs/react-redux/pull/1065/files#diff-4a561361865a8e6d0758088bf7cded8d

markerikson:
Hey, @timdorr : we need to figure out our branching strategy for v5 / v6.  Thoughts?

src/components/connectAdvanced.js
```diff
@@ -103,199 +85,55 @@ export default function connectAdvanced(
...
+      const {context, forwardRef, ...wrapperProps} = props
+      const contextToRead = context || defaultContext
+
+      const {store, storeState} = useContext(contextToRead)
+      const childPropsSelector = useMemo(() =>  createChildSelector(store), [store])
+
+      const childProps = childPropsSelector(storeState, wrapperProps)
+
+      const renderedChild = useMemo(() => {
+        return <WrappedComponent {...childProps} ref={forwardRef} />
```
Andarist: u could avoid copying those props with
```js
const renderedChild = useMemo(() => {
  const finalProps = forwardRef ? {...childProps, ref: forwardRef} : childProps
  return <WrappedComponent {...finalProps} />
}, [childProps, forwardRef])
```

timdorr:
Based on discussion behind closed doors in the super-secret invite-only gatekeeper zone, i.e. Twitter DMs, we're going to close this one out and work on the hooks-based approach instead. I'll nuke the branch, but it will stick around in this PR if we want to pull from it.

alexreardon:
Gasp!

markerikson:
Still sad that the v6 approach didn't work out how we thought, because the code in this branch was _beautiful_.

My v7 alpha branch?  Yeah, that's... uh... "functional".  That's a good word.  It functions.

It ain't pretty, though.

PR #1069, opened 2018-10-31
remove react-is/isValidElementType from production builds

alexreardon:
This is a proposal to only run the code for checking whether an element is valid in non-production builds. 

By implementing this change, bundlers can tree shake out `react-is` altogether for production builds 👍 

I think this is fine to be a development only check


markerikson:
Seems potentially reasonable.  @timdorr , thoughts?

alexreardon:
Given this is a setup issue, I think a development check is appropriate

alexreardon:
🎉

PR #1071, opened 2018-11-04
Use Prettier

NMinhNguyen:
I saw @timdorr's suggestion to use Prettier in https://github.com/reduxjs/react-redux/pull/1065#issuecomment-433670908 and decided to take a stab at integrating it. Given this project is already using ESLint, I thought running Prettier via ESLint was the simplest option (it does however mean that only linted files are run through Prettier, i.e. `src`, `test/utils` and `test/components`), but I'm happy to change the approach.


package-lock.json
```diff
@@ -1336,7 +1336,7 @@
           "integrity": "sha512-VT0ZI6kZRdTh8YyJw3SMbYm/u+NqfsAxEpWO0Pf9sq8/e94WxxOpPKx9FR1FlyCtOVDNOQ+8ntlqFxiRc+r5qA==",
           "dev": true,
           "requires": {
-            "color-convert": "1.9.1"
+            "color-convert": "^1.9.0"
```
NMinhNguyen: I'm not too sure why running `npm install` even on `master` leads to a diff in `package-lock.json` 😕

timdorr: Either someone before you wasn't using the latest version, or you need to update. 6.4.1 is the latest, I think.

timdorr:
OK, resolved the conflicts and updated the code. Thanks!

PR #1073, opened 2018-11-07
Add reference to why-did-you-update-redux

theKashey:
- Adding hint about using __why-did-you-update-redux__ as a mapStateToProps/selector quality accession.
- Adding a reference to __beautiful-react-redux__ as a way of automated selector memoization.

As a reference -  beautiful-react-redux is based on `memoize-state` which is based on automated state key usage tracking, and able to provide a proper memoization almost for any selector. If it did memoization job better than user - then it could guide user to a better memoization. This feature comes with a cost, but it's much lower that a cost of react component update.


markerikson:
I think I'm going to hold off on adding these links to this file.  However, we can probably mention them when we get to the upcoming pages on performance.

theKashey:
Up to you. Tools like this are a little missing piece of redux dev tooling right now.

Issue #1083, opened 2018-11-15
React-Redux v6 beta feedback thread


markerikson:
Opening this as a place to get user feedback on our React-Redux v6 beta releases.  We're specifically interested in bug reports and info on performance behavior, but would also appreciate feedback if it's working correctly as well.

To simplify the comment thread a bit: **if you've tried it and it works okay, please add a 👍 reaction to this comment.  If something broke, leave a 👎 reaction, plus a comment with details, including a link to your project if publicly available**.  

We'd also be interested in hearing details on:
- Particularly complex projects that work fine with v6
- Other libraries in the ecosystem that break due to changes in v6

The current beta is:  **[React-Redux v6.0.0-beta.3](https://github.com/reduxjs/react-redux/releases/tag/v6.0.0-beta.3)**

markerikson:
Known libraries that break in v6, and tracking issues if available:

- [`redux-form`: #4216](https://github.com/erikras/redux-form/pull/4216)
- [`connected-react-router`: #176](https://github.com/supasate/connected-react-router/issues/176)

yjimk:
I feel bad leaving a thumbs down, but with a very basic test and not delving too deep, upgrading to 6.0.0-beta.2 didn't work for me. I tried it on two large corporate eCommerce applications.

I'm not altogether surprised though, they are very large and complex applications touched by a wide range of developers and both have significant redux reliance.

Unfortunately I can't share the source with you as it is private code :(

react: 16.4.2
redux: 3.7.2
react-redux: 6.0.0-beta.2 (previously using 5.1.1 without issue)
react-router-redux: 5.0.0-alpha.9

Application uses SSR including server side redux store hydration

**Application 1**
Here's the stacktrace at runtime, server started and webpack was able to build without error. Looks like the error is occurring on the server before any client execution happens. Browser window is a blank screen.

```
Invariant Violation: withRef is removed. To access the wrapped instance, use a ref on the connected component
    at invariant (/Users/x/app/src/node_modules/invariant/invariant.js:40:15)
    at connectAdvanced (/Users/x/app/src/node_modules/react-redux/lib/components/connectAdvanced.js:70:26)
    at connect (/Users/x/app/src/node_modules/react-redux/lib/connect/connect.js:91:12)
    at Module.<anonymous> (/Users/x/app/src/build/webpack:/web/components/TraitWrapper/index.js:77:23)
[...]

PR #1085, opened 2018-11-19
Add a LegacyProvider component

timdorr:
This avoids polluting Provider with StrictMode-incompatible stuff. Hopefully, this makes working with outdated libraries easier.

This is per [Dan's comment](https://github.com/reduxjs/react-redux/issues/1083#issuecomment-439923981)

### DIFF
--- a/src/components/LegacyProvider.js
+++ b/src/components/LegacyProvider.js
@@ -1,0 +1,29 @@
+import React, { Component } from 'react'
+import PropTypes from 'prop-types'
+
+import Provider from './Provider'
+
+class LegacyProvider extends Component {
+  getChildContext() {
+    return { store: this.props.store }
+  }
+
+  render() {
+    const { store, ...props } = this.props
+    return <Provider store={store} {...props} />
+  }
+}
+
+LegacyProvider.propTypes = {
+  store: PropTypes.shape({
+    subscribe: PropTypes.func.isRequired,
+    dispatch: PropTypes.func.isRequired,
+    getState: PropTypes.func.isRequired
+  })
+}
+
+LegacyProvider.childContextTypes = {
+  store: PropTypes.object.isRequired
+}
+
+export default LegacyProvider
--- a/src/index.js
+++ b/src/index.js
@@ -1,6 +1,7 @@
 import Provider from './components/Provider'
+import LegacyProvider from './components/LegacyProvider'
 import connectAdvanced from './components/connectAdvanced'
 import { ReactReduxContext } from './components/Context'
 import connect from './connect/connect'
 
-export { Provider, connectAdvanced, ReactReduxContext, connect }
+export { Provider, LegacyProvider, connectAdvanced, ReactReduxContext, connect }
--- a/test/components/LegacyProvider.spec.js
+++ b/test/components/LegacyProvider.spec.js
@@ -1,0 +1,43 @@
+import React, { Component } from 'react'
+import PropTypes from 'prop-types'
+import { LegacyProvider } from '../../src/index.js'
+import { createStore } from 'redux'
+import * as rtl from 'react-testing-library'
+import 'jest-dom/extend-expect'
+
+const createExampleTextReducer = () => (state = 'example text') => state
+
+describe('LegacyProvider', () => {
+  it('provides a legacy context', () => {
+    class LegacyChild extends Component {
+      render() {
+        const store = this.context.store
+
+        let text = ''
+
+        if (store) {
+          text = store.getState().toString()
+        }
+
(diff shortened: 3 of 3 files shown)


gaearon:
It should probably warn that this is deprecated, exists strictly as a migration step, and will break in the future. And that this makes your app slower.

markerikson:
I don't think I like this idea, for several reasons:

- Legacy context will go away eventually
- React-Redux v5 still works fine, except potentially in concurrent mode
- Accessing the store out of context has never been part of our public API
- There's been several React issues filed where using legacy context and new context together caused problems
- People would need to slap another wrapper component around the entire tree

The community has done some neat stuff based on unofficially accessing the store out of context.  I don't want to _officially_ support it, but I don't want to prevent people from doing that either.

This is one of the reasons why I chose to put the entire store into new context, rather than just the `dispatch` method.  The primary goal is to use new context to make the latest store _state_ accessible, but if someone _really_ wants to grab the store out of the context consumer, they can.  That's also why we're exporting the default context object.

At most, we could give a semi-recommended pattern for "If you _really_ want to, here's how you _could_ get access to the store, but be aware this could break at any time" (which basically amounts to reusing the exported context object).

mpeyper:
`redux-subspace` and `redux-dynostore` author/maintainer here.  

Whilst this may work as a temporary measure to keep compatibility when moving to new context, I do not think it should be the end goal for supporting out use cases.  What we do in both my libraries is not possible with just a `getState` and `dispatch` function We store additional fields and utility functions on the store itself.  This has been a useful pattern as:

a) It's already bing passed to all the places we want to use it
b) It's relatively hidden to the rest of the ecosystem as anything outside out library only knows about `getState` and `dispatch`
c) Store enhancers provide a really convenient API for adding them

That said, I could envision libraries like mine providing a replacement provider to inject the store into an alternative context to get access to the store instead, e.g.

```js
// Context.js
import React from 'react'

export const Context = React.createContext()

// Provider.js
import React from 'react'
import { Provider as ReduxProvider } from 'react-redux'
import { Context } from './Context'

const ReduxSubspaceContext = React.createContext()

export const Provider = ({ store, children, ...props }) => {
  return (
    <ReduxProvider {...props} store={store}>
      <Context.Provider value={store}
        {children}
      </Context.Provider>
    </ReduxProvider>
  )
}

// SubspaceProvider.js - the interception component
import React from 'react'
import { subspace } from 'redux-subspace' // does the _real_ work
import { Provider } from './Provider'
import { Context } from './Context'

const SubspaceProvider = ({ mapState, namespace, children }) => {
  return (
    <Context.Consumer>
      {(store) => (
        <Provider store={subspace(mapState, namespace)(store)}>
          {children}
        </Provider>
      )}
    </Context.Consumer>
}
```

This would work fine as long as there was only a single library using this pattern.  As soon as both `redux-subspace` and `redux-dynostore` (which are designed to be used together) are required, we'd need to provide some mechanism for passing what the wrapped provider was so that one could wrap the redux provider and the other wrap that provider.

mpeyper:
For the record, I'm willing to accept that my use case is fairly niche, so I do understand if not supporting it to save bloating the library for everyone else might actually be the preferred approach here.

markerikson:
@mpeyper : I've seen other libraries that do the same kind of thing, so you're not the only one :)  Just the first example that comes to mind.

Off the top of your head, does accessing the store via the `ReactReduxContext` instance work for your situation?

mpeyper:
Yes, I believe it would, assuming it is just the same store that is passed to the `Provider` (which it looks like it is from a quick look at the code).

Although, we would need access to `ReactReduxContext.Consumer` to use it if I'm not mistaken.

markerikson:
Right, we export the entire context object - see [Context.js](https://github.com/reduxjs/react-redux/blob/7496b91d307b37ffb976beddf3a2a226be28679c/src/components/Context.js) and [index.js](https://github.com/reduxjs/react-redux/blob/7496b91d307b37ffb976beddf3a2a226be28679c/src/index.js), and that's specifically to (unofficially) support this use case.

mpeyper:
Your tweet made it sound like it was going away in favour of this PR. If that's not the case, great!

markerikson:
Ah, sorry if I gave that impression.  

No, the discussion in #1083 was how to handle libs that currently expect to access the store out of legacy context.  Dan proposed adding this, Tim filed a PR, and I'm saying I don't think it's a good idea.  But, I wanted to get other eyes on the proposal, especially for lib authors like you.

timdorr:
OK, I'm agreed on this not being the right approach. Figured it was worth a quick PR. Thanks for everyone's input!

PR #1092, opened 2018-11-26
simplified example in docs

stavalfi:


timdorr:
Thanks, but this changes the results from a single `null` value to an empty array. `null` actually has some value here, indicating that we haven't produced any todos at all, not just that they are none left. It may not matter in this example, but we want to establish that `null` is a good pattern to use.

PR #1098, opened 2018-12-05
React required version

joelgallant:
1. I removed the unsupported react version tests - if they should be kept for some kind of historical reasons, I can revert.
2. I added babel-jest, which is a dependency required for the test suite which wasn't depended on directly
3. I updated documentation where I found it (just grepped) that said 0.14 - there might be more I missed


README.md
```diff
@@ -11,7 +11,7 @@ Performant and flexible.
 
 ## Installation
 
-React Redux requires **React 0.14 or later.**
+React Redux requires **React 16.4 or later.**
```
timdorr: Nice catch!

timdorr:
Thanks!

Issue #1001, opened 2018-08-17
Rewrite React-Redux docs completely


markerikson:
The current React-Redux docs aren't very helpful, for a lot of reasons:

- They're just a Markdown file in the repo, instead of being published in HTML form
- It feels very "dense" and jargon-heavy
- It doesn't clearly lay out the various options for `connect` and ways that you can use it
- It's primarily in API reference form, rather than a "how to"

We do also have the "Using React with Redux" page over in the main Redux docs, which is a bit more of a "getting started" page, but it's also not great.

So, I would really like to completely rewrite our React-Redux docs from scratch.

As a rough outline, what I'd like to see is:

- Quick Start (copy-pasteable examples that show how to add `<Provider>` and do a basic call to `connect()`
- How It Works (an explanation of how `<Provider>` makes the store accessible, and roughly what `connect` does internally to subscribe and extract data)
- Advanced Techniques / Recipes (examples of things like using the "factory function" syntax for per-component selector memoization)
- API Reference (something similar to what we have now)

I do already have a Gitbook setup configured for this, same as the main Redux docs.  I should be able to get the `react-redux.js.org` domain name for this.

markerikson:
Awesome, thanks - really appreciate it!  

I'd previously discussed this with @jsonnull .  One note from a discussion:

> Since I was envisioning a breakdown into "High level concepts" ... "Recipes" ... "API Docs", I thought about making a PR which simply pulls apart the current docs into those categories.

Might be a place to start.

timdorr:
Since Redux + React is such a common combo, I'd rather keep it all under the redux.js.org site. There's already a ton of React code in those docs anyways. And it also has the side effect of being more discoverable that way (one single search bar for both libraries). 

We can point everything here to that page (or pages!). We can also consolidate the API reference, so there's one place for all the API docs.

markerikson:
@wgao19 : Hiya! Thanks for looking into this.  A couple quick thoughts:

- I did set up the initial Gitbook integration a while back.  It's viewable at https://redux.gitbook.io/react-redux/ .  I can see us going either way on whether it eventually becomes part of the main Redux docs page at https://redux.js.org , or whether it stays as a separate page (in which case I would want to get the `https://react-redux.js.org` domain name).  In any case, we can focus on the content part for now, and worry about the publishing aspect later.  
[...]

PR #1106, opened 2018-12-08
Enable versioned docs

wgao19:
### What does this PR do?

New doc piece regarding #1001:
Enable versioned docs

### Summary of the Changes

- Add a "versioned docs" page to doc site
- Add 5.1.1 docs
- Start 6.0.0 docs

### How to work with versioned docs

Docusaurus assumes that master is the docs for `next`. Once `next` is released, run `yarn version` within `website/`. This copies the diff between current master docs with latest doc site's version to the doc site's created new version.

### Notes

- I've created the 6.0 docs since 6.0 is already released. So for this particular version, we'll need to edit the docs under `website/versioned_docs/version-6.0.0/api/api.md`, etc. In the future, we should edit docs on master, and simply run `yarn version` with new version number after release.

- Docusaurus's versioned docs feature does not assume versioned code. We don't need to keep the code branches of older versions.


markerikson:
Sweet!  I'll try to take a look at this over the weekend.

timdorr:
> Docusaurus's versioned docs feature does not assume versioned code. We don't need to keep the code branches of older versions.

Can it? We have a 5.x branch that we'll keep around for a bit while folks migrate to 6.0. If we can isolate doc versions by branches, that would actually be ideal.

timdorr:
Is there a chance to run some script pre-build that essentially does `git --work-tree=website/versioned_docs/version-5.1.1/ checkout 5.x -- docs/`?

wgao19:
I think Docusaurus' idea is that it assumes master branch is always `next`. And whenever we run `docusaurus-version vx.x`, it creates a directory for that version, and generates docs for that version based off the difference between that version and the latest version prior to that.

> Only files in the docs directory and sidebar files that differ from those of the latest version will get copied each time a new version is specified. If there is no change across versions, Docusaurus will use the file from the latest version with that file.

In our case latest version was 5.1.1. master was 6.0, which was also `@next`. Once I ran `docusaurus-version 6.0.0`, it copied the docs from master that was different from 5.1.1. to `versioned_docs/version-6.0.0`. I changed `api.md` regarding `forwardRef` and passing store. Therefore only `api.md` was copied.

> But all that has to be on the same branch together

Yeah I believe so. The generated versioned docs are under the doc site's directory.

> Technically? Probably. But that's really not how DS is meant to work as far as I know, and I don't think it gains us anything useful.

I guess nothing would stop us from writing such script by ourselves. We just need to copy over the docs/ and the versioned_sidebars. The header and footer can be a bit off if they're not consistent across versions. And yeah, probably not how Docusaurus is designed to work.

timdorr:
Thanks for the explanation. In that case, since versioned docs _do_ matter for this project (not so much with Redux, though), I think this looks good.

Mark, did you have any further tweaks to make to the base docs before we split out the versions?

markerikson:
Yeah, gimme another day or two to finish polishing things.

timdorr:
Any chance we can still get to this? With 7.0 imminent, it's going to be more important.

wgao19:
I believe so. I'll spend some time this weekend to catch up a bit, update this branch and let's see.

wgao19:
Hi @timdorr, @markerikson,

I've updated this branch and have put v5 to previous version, v6 to stable version, and put up a link to v7 as pre-release version. It should now work under Netlify's preview.

Changes
---
- Make version docs work: There's a doc number next to the logo, clicking on it will go to the versioned docs page
- Older versions are now v5.x, v6.x
  - v5.x: docs before v6 released
  - v6.x: docs of current master
- Put up a link to v7.x in the versioned docs page, this points to version "next" of the docs, which is the current master

Versioned docs structure
---
```bash
.
├──  # ... other files
├── docs                      # pre-release version (v7)
│   ├── README.md
│   ├── api
│   ├── introduction
│   ├── troubleshooting.md
│   └── using-react-redux
└── website
    ├── # ... other files
    ├── siteConfig.js
    ├── versioned_docs        # contains stable + previous versions
    │   ├── version-5.x       # copies docs/ when running docusaurus' version command
    │   └── version-6.x
    ├── versioned_sidebars 
    │   ├── version-5.x-sidebars.json
    │   └── version-6.x-sidebars.json
    ├── versions.json         # list of versioned files
    └── yarn.lock
```

Working with versioned docs
---
- To update docs for pre-release versions (v7), update the docs under root `/docs`.
- To update docs for any stable or previous versions, update the docs under `/website/versioned_docs/`
- Once pre-release version releases, run Docusaurus command `yarn run version v7.x` under `/website`, and remove `nextVersion: 'v7.x'` from `website/siteConfig.js`.

timdorr:
Oh, this is so good! Thank you!

@markerikson Any objections? I'd say just merge this in right now.

markerikson:
Sure, let's do it!  @wgao19 , this is great - thanks!

PR #1122, opened 2018-12-12
Fix issue with isValidElementType invariant check crashing on valid t…

Andarist:
…ype with circular structure

Repro - https://codesandbox.io/s/jnmoxr5mk3

Emotion is attaching created component to the returned styled component (creating a circular structure) to easily "flatten" subsequent calls wrapping that component, as in:
```js
const Div = styled.div`color: white;`
const Div2 = styled(Div)`background-color: pink;`
```

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -42,12 +42,12 @@ npm run test
 
 To run in explicit React versions (the number is the version, so `test:16.3` will run in React version `16.3`):
 ```
-REACT=16.4 npm run test:ci
+REACT=16.4 npm run test
 ```
 
 To run tests in all supported React versions, `16.4`, 16.5`,
 ```
-REACT=all npm run test:ci
+REACT=all npm run test
 ```
 
 To continuously watch and run tests, run the following:
--- a/test/run-tests.js
+++ b/test/run-tests.js
@@ -27,4 +27,7 @@ if (version.toLowerCase() === 'all') {
   }
 }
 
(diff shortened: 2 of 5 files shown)


timdorr:
Can you run this through Prettier? And can you add a test?

test/run-tests.js
```diff
@@ -27,4 +27,7 @@ if (version.toLowerCase() === 'all') {
   }
 }
 
-npmRun.execSync(`jest -c '${JSON.stringify(jestConfig)}'`, { stdio: 'inherit' })
+npmRun.execSync(
+  `jest -c '${JSON.stringify(jestConfig)}' ${process.argv.slice(2).join(' ')}`,
```
Andarist: this has actually allowed me to use `--watch` (which is still not ideal because we have to edit copy of tests rather than real test files to benefit from that, but better this now than nothing)

CONTRIBUTING.md
```diff
@@ -42,12 +42,12 @@ npm run test
 
 To run in explicit React versions (the number is the version, so `test:16.3` will run in React version `16.3`):
 ```
-REACT=16.4 npm run test:ci
+REACT=16.4 npm run test
```
Andarist: there is no `test:ci` script

Andarist:
> Can you run this through Prettier? And can you add a test?

Done & done.

timdorr:
Thanks for this and the other updates!

PR #1131, opened 2018-12-14
Minor optimization: remove conditional

ericanderson:
The conditional in selectDerivedProps is unnecessary. The code behaves
identically now but saves about 10% of performance for that portion of
the function.

See: https://jsperf.com/conditional-vs-not

Code left in place but commented in case someone ever wants to add
something like it back for "clarity".


timdorr:
Thanks!

PR #1134, opened 2018-12-17
Ensure that component prop 'context' really contains a React context …

casdevs:
…before using it

after switching to react-redux 6.0.0, we've had a lot of errors stating

`Warning: React.createElement: type is invalid -- expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports.`

all over in our app. After digging deeper, we've discovered that we use a lot of (connected) components already using a property named `context` which conflicts with your new connectAdvanced code. So maybe you can improve the check on `this.props.context` a little bit to ensure it is used only if it really contains a valid React context.

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -2,7 +2,7 @@ import nodeResolve from 'rollup-plugin-node-resolve'
 import babel from 'rollup-plugin-babel'
 import replace from 'rollup-plugin-replace'
 import commonjs from 'rollup-plugin-commonjs'
-import {uglify} from 'rollup-plugin-uglify'
+import { uglify } from 'rollup-plugin-uglify'
 import pkg from './package.json'
 
 const env = process.env.NODE_ENV
@@ -22,14 +22,17 @@ const config = {
     nodeResolve(),
     babel({
       exclude: '**/node_modules/**',
-      runtimeHelpers: true,
+      runtimeHelpers: true
     }),
     replace({
       'process.env.NODE_ENV': JSON.stringify(env)
     }),
     commonjs({
       namedExports: {
-        'node_modules/react-is/index.js': ['isValidElementType'],
+        'node_modules/react-is/index.js': [
+          'isValidElementType',
+          'isContextConsumer'
+        ]
       }
     })
   ]
--- a/src/components/connectAdvanced.js
+++ b/src/components/connectAdvanced.js
@@ -213,7 +213,7 @@ export default function connectAdvanced(
       }
 
       render() {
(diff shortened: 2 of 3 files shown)


timdorr:
Can you add a test?

casdevs:
ok, I'm going to add tests within the next few days

casdevs:
improved checks & added test

casdevs:
maybe it would even be better to check for this.props.context.$$typeof === REACT_CONTEXT_TYPE, but react doesn't seem to export its type symbols...

markerikson:
We do have `react-is` being imported already - might be a function in there you can use?

casdevs:
we could check for `this.props.context.$$typeof === ContextConsumer`
with `export const ContextConsumer = REACT_CONTEXT_TYPE;` being exported by 'react-is'
this works and all tests would succeed.

But a dedicated method like isContext doesn't seem to be available from 'react-is'.

casdevs:
great, thanks!
When we decide to have type checks on members of `this.props.context`, we could even omit the check on `this.props.context.provider` as only `.consumer` is used after the checks.

timdorr:
Yeah, I removed that. It was also complaining about isContextProvider not being exported from react-is. Weird.

casdevs:
hm, it doesn't seem to know `isContextConsumer` either...!?

timdorr:
And now isContextConsumer. wut?

casdevs:
weird, when running the tests locally, `isContextConsumer` is imported correctly.

however, the check must be done this way, otherwise other test cases fail:
`
const ContextToUse =
          this.props.context &&
          this.props.context.Consumer &&
          isContextConsumer(<this.props.context.Consumer />)
            ? this.props.context
            : Context
`

seems to be a litte overhead, so maybe it would be better to simply check for
`this.props.context.$$typeof === ContextConsumer` as I've stated above.
I'm beginning to think that react-is' `ContextConsumer` export should rather be named `ContextToBeConsumedByContextConsumer` though....

casdevs:
Happy new year to all of you!
I've updated the code again and the pull request should now be ok to get merged in.

timdorr:
Ah, I missed the Rollup config! I knew it was something obvious.

timdorr:
OK, this looks good to me. Thanks!

casdevs:
great, thanks for merging that in!

maxkostow:
When do you plan on cutting a new release with this change?

markerikson:
@timdorr ?

Issue #1132, opened 2018-12-14
Update docs for using a different store key


ethanroday:
- Mark `createProvider` as deprecated/removed.
- Mark `storeKey` in `connectAdvanced()` as deprecated/removed.
- It would also be nice to add a new section detailing how to accomplish it with the new API.

markerikson:
Yeah, we're still working on updating the docs for v6 (and hopefully setting up versioned docs for v5 vs v6).

markerikson:
Out of curiosity, were you actually using that?  If so, what was your use case?

ethanroday:
It's for a Chrome extension, and it grew out of some architectural particulars there.

We have a main store in the background page of the extension that's the final source of truth for all user data. Our UI is a React app which can be inserted directly into pages as an iFrame, and there might be several instances of the UI inserted into different tabs at once. We use the excellent [`react-chrome-redux`](https://github.com/tshaddix/react-chrome-redux) library to proxy the main store to all of the different UI instances, but each UI instance also has a bunch of local UI state to keep track of. Rather than make a new slice of state in the background that stores UI state separately for each tab, we just made a second store that's local to each UI instance. (Going with the former solution would also mean paying a several-millisecond cost for messaging between the UI instance and the background.)

It can be confusing at times, but it has actually worked out rather well.

markerikson:
Ah, interesting.

Does the ability to supply a custom context instance in v6 still help for your use case?

ethanroday:
I haven't tried just yet, but I imagine it should be workable. I can create the context and pass it to the `Provider` at the app level, and then just call connect with the `context` option. (I'm basing this on the `You may also pass a {context : MyContext} option to connect` part of the error message.)

Does that sound right?

Issue #1123, opened 2018-12-13
Access store from context?


badtant:
Hi, I just updated to 6.0.0 and have a question. I can't really figure out from the docs how to access the store from context anymore.

My app is wrapped with
`<Provider store={store}>`

Then in any component I could access it with
```
MyComponent.contextTypes = {
    store: PropTypes.object.isRequired
};

->

this.context.store
```

How do i achieve the same thing with 6.0.0?

timdorr:
We use the new context API, which is [exported as `ReactReduxContext`](https://github.com/reduxjs/react-redux/blob/master/src/index.js#L6).

badtant:
Got it working but this also seem to subscribe to updates from the store.
What I want is just access to the store that I created without subscribing to updates.

I can of course export it from where I created it. Is that the best way?

markerikson:
What are you actually trying to accomplish?

badtant:
I want to read a value from the store in componentWillUnmount.
So I want access to the store without subscribing to any updates.

Previously I got access with this.context.store.getState()

markerikson:
Per the [React docs on using context](https://reactjs.org/docs/context.html), you would either need to add a `contextType` field to your component so that you can access the context value as `this.context`, or have a wrapper component that renders a `<Context.Consumer>`, extracts the store, and passes it as a prop to your own component.

What kind of value are you needing to read from the store on unmount?  Why not just connect the component, and extract that value in `mapState` as normal?

badtant:
- contextType is not supported with react-redux 6.0.0 if I understood correctly
- mapState works but will that subscribes me to updated I don't need
- Context.Consumer also works but did also subscribes me to updates

I'm just trying to think as performant as possible. Hence not wanting to subscribe if I don't need all the updates.

markerikson:
`contextType` is a React thing, not a React-Redux thing, and should work fine.

As I asked, what kind of value are you trying to use?  Is it something that you expect will be changing frequently?  Do you have specific reasons why you think connecting to get this value will be a performance issue?  Why do you only need this in `componentWillUnmount`?

badtant:
Ok. I used contextType before upgrading to 6.0.0. When I upgraded it stopped working... The release notes said this:
"Any library that attempts to access the store instance out of legacy context will break, because we now put the store state into a <Context.Provider> instead."

The value I want to read is a string that does change quite a lot and in this component I only need it when it unmounts. It's used for statistics tracking in this case.
[...]

Issue #1104, opened 2018-12-05
Any migration guide?


wolfy1339:
Is there any documentation on how to migrate from v5 to v6?
I'm not exactly sure what to change or how to change it

markerikson:
Migration should just be a matter of updating your `package.json` from `"react-redux" : "^5.0"` to `"react-redux" : "^6.0"`, and reinstalling with your package manager of choice.

If you are using one of the couple features affected by breaking changes (which is unlikely):

- The `withRef` option: use `forwardRef` instead, and change any uses of `theWrapperComponentInstance.getWrappedInstance()` to just be `theActualComponentInstance`
- If you were passing `store` as a prop to a connected component (likely in a test), you'll need to render a `<Provider store={store}>` around it. 

That should be it.

geminiyellow:
hi @markerikson , v5 > v6, 

```
react-dom.production.min.js:3833 Error: Could not find "store" in either the context or props of "Connect(FirestoreConnect(t))". Either wrap the root component in a <Provider>, or explicitly pass "store" as a prop to "Connect(FirestoreConnect(t))".
```

i did pass `store` use this:

```
<ReduxProvider store={store}>
```

markerikson:
Did you happen to just ask that question on Reddit, as well?  [I answered it there](https://www.reddit.com/r/reactjs/comments/a3j497/reactredux_v600_with_reactreduxfirebase/), but I'll paste the answer here too.

The `redux-firestore-connect` library appears to be [grabbing the Redux store out of old context](https://github.com/prescottprue/react-redux-firebase/blob/34b8f08e4f221ba9d04fdeeb071a26f393cfd9ff/src/firestoreConnect.js#L32).  Per the release notes, [that was never supported, and will break in v6](https://github.com/reduxjs/react-redux/releases/tag/v6.0.0).

That library will need to be modified to access the store out of new context instead.  See https://github.com/supasate/connected-react-router/pull/191 for an example of how `connected-react-router` is changing that way.

arthurgeron-work:
> Migration should just be a matter of updating your `package.json` from `"react-redux" : "^5.0"` to `"react-redux" : "^6.0"`, and reinstalling with your package manager of choice.
> 
> If you are using one of the couple features affected by breaking changes (which is unlikely):
> 
> * The `withRef` option: use `forwardRef` instead, and change any uses of `theWrapperComponentInstance.getWrappedInstance()` to just be `theActualComponentInstance`
[...]

PR #1148, opened 2018-12-27
Add / change docs about v6

wgao19:
## What does this PR do?

Regarding #1001, addressing #1104, #1123, #1132:

Add docs regarding v6 changes.

## Summary of the Changes

- Quick Start: Add a section regarding passing `context` to `<Provider />` and connected components
- Provider API: Remove the note about direct passing `store` as props to connected component
- Provider API: Add `context` props 
- New page: Add a doc page about accessing the store with v6

## Are there points in the code the reviewer needs to double check

Please feel free to make changes if my understanding is off from the intended use case.


markerikson:
Thanks!

Did some rework on the main page content.  Removed the additions to "Quick Start", because I don't think that's something that needs to be emphasized.  Also used this branch to copy over all the CSS updates from the main site.

PR #1152, opened 2019-01-02
Fix[Website]: Top section layout issue

shahzaibkhalid:
**What does this PR contains?**

Fix for a small UI issue for top features section.

**Does it contains any breaking change?**

No.

**Before Fix**

image

**After Fix**

image


timdorr:
Thanks, but there are actually some design changes in #1148 that we need to revert instead. That will cover mobile as well.

Issue #1150, opened 2018-12-28
RFC - API for dynamically adding reducers


Ephem:
See discussion in issue #1126 for context.

**Problem:** Because `ReactReduxContext` is not part of the public API, there is currently no way using only public APIs to safely support dynamic loading of reducers. Even using this, implementation is tricky and error prone and there is also room to create a better story around code splitting reducers.

**Proposed solution:** User provides a `getRootReducer(newReducers)`-function to `<Provider>`. New component `<WithReducers reducers={newReducers}>` can be used to wrap a subtree that needs these reducers.

---

# Summary of problem (recap of #1126)

`react-redux` version 6 now uses the new context instead of the legacy one, and instead of every connected component being subscribed directly to the store, only the root `<Provider>` is. That is:

* Before: Every connected component called `getState()`
* After: Only `<Provider>` calls `getState()`, state gets passed along via context *and does not change during a single render*

This is done to avoid “tearing” in concurrent rendering mode (avoid different parts of application rendering with different state).

In a world with concurrent rendering, `replaceReducer` is trickier to use from within the render tree. This is because a call to this function modifies the state in the store and subsequent components expect this state to be available, but now no longer fetches the latest state automatically, but instead from context.

On top of fetching `store` from the context and calling `replaceReducer` on it, the solution from #1126 is to *patch the current storeState from context with the state added by the new reducers* by wrapping that subtree in a new `<ReactReduxContext.Provider>`. However, `ReactReduxContext` is not part of the public API.

In a sense this is not a problem new to v6, since several of the old solutions for code splitting reducers in React relied on fetching `store` directly from context, which was never part of the public API.

On top of this, I feel like code splitting is something to be encouraged and there is room to improve the story around this and that this aligns with the goals of `react-redux`.

# Proposed solution

Please note that this is just a rough sketch of what a solution could look like. I’d love feedback and to get a discussion going around both large issues and small.

[Rough implementation and example usage in CodeSandbox](https://codesandbox.io/s/n410k63qk4) - This is forked from @rgrove’s excellent repro-case from #1126

My hope is that this solution strikes a balance between being immediately useful and preserving the flexibility for the user to construct a root-reducer and injection techniques however they feel like.

**Goals:**

* Add API for adding reducers from the component tree
* Do not require direct use of `store` or `ReactReduxContext` in userland
* Flexible API that supports both simple and complex usecases well and can be built on top of
* Small surface area
[...]

Issue #1126, opened 2018-12-13
`<Provider>` misses state changes that occur between when its constructor runs and when it mounts


rgrove:
**Do you want to request a _feature_ or report a _bug_?**

Bug

**What is the current behavior?**

If state changes occur during the time after `<Provider>`'s constructor runs and before `<Provider>` is mounted, `<Provider>` will miss those changes and won't make them available to connected components.

This is probably a rare use case, but it can occur in a very specific scenario that's useful in an app that uses both server-side and client rendering and needs to allow dynamically loaded components (such as components loaded when the route changes) to attach new reducers to an existing store.

Here's a reduced test case that fails in Redux 6.0: https://codesandbox.io/s/612k3pv1yz

**What is the expected behavior?**

Connected components should always see the most recent Redux state, even if that state changed between when `<Provider>` was constructed and when it was mounted. This was the behavior in Redux 5.x.

Here's an exact copy of the reduced test case above, but using Redux 5.1.1 to demonstrate that this works fine there: https://codesandbox.io/s/yvw1vmnkrv

**Which versions of Redux, and which browser and OS are affected by this issue? Did this work in previous versions of Redux?**

This worked in Redux 5.x, but broke in 6.0. It's unrelated to any specific browser or OS.

**More background on this use case**

I realize this use case may be a little hard to understand at first glance, so I'll try to explain in more detail why it's valuable.

The repro cases above simulates a technique that's useful when using Redux and React Redux in an SSR-compatible app that needs to be able to load components and attach new reducers on demand (such as when the route changes) without knowing up front what components or reducers might eventually be loaded.

This technique is used extensively by [Cake](https://www.cake.co). I believe [New Twitter](https://mobile.twitter.com/) uses a similar approach, but [they're still on React Redux 4.x](https://mobile.twitter.com/paularmstrong/status/1073037586438987776) so aren't yet affected by this bug.

When rendering on the server, we can't load components during the render phase (because the SSR render phase is synchronous). So instead we need to load all components for a given render pass up front, attach their reducers to the existing Redux store, then render the app.

Dynamically loaded components may themselves import other components, and components at any level of the import tree could be Redux-connected. This means each component must be able to attach its own reducers. The `withRedux` decorator in the repro case simplifies this by wrapping a dynamically loaded component (such as the `Route` component in the example) in a higher order component that attaches reducers on demand in its constructor.

[...]

PR #1154, opened 2019-01-02
RFC - StoreModifier, dynamically modifying store from component tree

Ephem:
> Warning: Do not merge!
> For now, this PR is an RFC submitted for discussion only.

This PR is a continuation and updated version of #1150, also see #1126 for more background.

I am posting this as a PR to help more nuanced discussions and for better change tracking. The code changes proposed are an unfinished demo-implementation only, it has not been tested exhaustively and is currently lacking any unit/integration-tests.

**Problem:** Because `ReactReduxContext` is not part of the public API, there is currently no way using only public APIs to safely support dynamic loading of reducers (or modifying the store from inside the component tree). Even using `ReactReduxContext`, implementation is tricky, error prone and hard to get right in userland (especially with SSR) and there is also room to create a better story around code splitting reducers.

**Proposed solution:** User provides a `modifyStore(modification, storeProxy, state)`-function to `<Provider>`. New component `<StoreModifier modification={modification}>` can be used to wrap a subtree that needs dynamically loaded reducers (or other store modifications). HOC `withModifiedStore` can be used as an alternative to `<StoreModifier>`.

[Full RFC as rendered text](https://github.com/Ephem/react-redux/blob/rfc-store-modifier/RFC_STORE_MODIFIER.md)

[Simple CodeSandbox example](https://codesandbox.io/s/3y1689prx6) (Originally forked from @rgrove)

I welcome any and all comments, suggestions and feedback!

### DIFF
--- a/RFC_STORE_MODIFIER.md
+++ b/RFC_STORE_MODIFIER.md
@@ -1,0 +1,185 @@
+# RFC - Dynamically modify the store from within the component tree
+
+The original version of this RFC can be found in [#1150](https://github.com/reduxjs/react-redux/issues/1150)
+
+The goal of this RFC is to provide an official way to modify the store from within the component tree (without getting it directly from `ReactReduxContext`). Modifications to the store might result in subtle bugs, especially with concurrent rendering, and should be considered unsafe.
+
+The only guarantee React-Redux makes, is that if you use `replaceReducer` as a modification, the `storeState` on context will be patched with the new state for all children, even for the current render.
+
+The secondary goal of this RFC is to make it easier out of the box to code split reducers.
+
+(This file itself is never meant to be merged)
+
+## Summary of problem (recap of #1126)
+
+Omitted for brevity, see summary of the problem space in [#1150](https://github.com/reduxjs/react-redux/issues/1150), read early discussions in [#1126](https://github.com/reduxjs/react-redux/issues/1126)
+
+## Proposed solution
+
+Please note that this is just a rough sketch of what a solution could look like. I’d love feedback and to get a discussion going around both large issues and small. I've listed alternative solutions towards the end.
+
+My hope is that this solution strikes a balance between being immediately useful and preserving the flexibility for users/libraries to construct a root-reducer and injection techniques however they feel like.
+
+I am NOT proposing to hide the `ReactReduxContext` or disallow getting the store directly from it, merely to keep it as a non-public API for usecases not covered by this RFC.
+
+**Goals:**
+
+- Add API for adding reducers from the component tree
+- Optionally: Make it easier/safer to do other dynamic store modifications (register sagas, epics etc)
+- Do not require direct use of `store` or `ReactReduxContext` in userland
+- Flexible API that supports both simple and complex usecases well
+- Both existing and new libraries should be able to build more complex solutions on top of this API, but not have to worry about the intricacies of patching `storeState` on context when replacing reducers
+- Small surface area
+- Aligned with `react-redux` v6 approach for preparing for concurrent-safeness
+- SSR-support
+
+### Example usage
+
+(Some imports etc omitted for brevity)
+
+```jsx
+import { withModifiedStore } from 'react-redux'
+import reducer from './namesReducer'
+
+function Greeter({ currentName }) {
+  return `Hi ${currentName}!`
+}
+
+const mapStateToProps = state => ({
+  currentName: state.names.currentName
+})
+
+// Structure of this object is not prescriptive,
+// it depends entirely on the implementation of
+// modifyStore
+// In a sense this is quite like an action, in that it
+// is an object that should describe the store modification
+const modification = {
+  newReducers: {
+    names: reducer
+  }
+}
+
+export default withModifiedStore(modification)(
+  connect(mapStateToProps)(Greeter)
+)
+```
+
+```jsx
+import staticReducers from './staticReducers'
+
+// Example simple userland implementation
+let reducers = { ...staticReducers }
+function modifyStore({ newReducers }, storeProxy, state) {
+  reducers = { ...reducers, ...newReducers }
+  storeProxy.replaceReducer(combineReducers(reducers))
+}
+
+const DynamicGreeter = React.lazy(() => import('./Greeter'))
+
+ReactDOM.render(
+  <Provider store={store} modifyStore={modifyStore}>
+    <Suspense fallback={<div>Loading...</div>}>
+      <DynamicGreeter />
+    </Suspense>
+  </Provider>
+)
+```
+
+### API
+
+#### `modifyStore(modification, storeProxy, state)`
+
+This is a callback that is supposed to be implemented by the user/by libraries and passed into `Provider` or directly into `StoreModifier` or `withModifiedStore`.
+
+Most implementations of `modifyStore` need to be stateful, and the only way to support this for SSR is to create one stateful instance per request. Passing it into `<Provider>` puts the function on the context behind the scenes and is the recommended approach even if SSR is not needed.
+
+**Arguments (automatically passed in to the function)**
+
+1. `modification` is an object that can contain any data that needs to be passed from the calling site (`<StoreModifier>`), for example reducers to be added to the store. There is no prescribed shape for this object, it is supposed to differ between implementations
+
+2. `storeProxy` here is a wrapped version of `store` that spies on `replaceReducer` to be able to safely patch the context with the new state if it is called. Maybe this should be a stripped down version of `store`, leaving out `subscribe` and possibly other methods unsafe to use here? Could also keep them in but provide warnings.
+
+3. `state` is passed in, in case it is needed for conditional logic, because `getState` is kind of unsafe here. This `state`-object contains the current state _stored on context for this render_ and is not the latest version of the state in the store.
+
+**Returns**
+
+This function can optionally return a cleanup-function, this will be called when the calling component unmounts. This way users can optionally implement functionality for removing reducers/unregistering sagas/epics when they are no longer needed.
+
+#### `<StoreModifier modification={} modifyStore={}>{children}</StoreModifier>`
+
+This component calls the `modifyStore` with the `modification` provided, a `storeProxy` and the current `storeState` from context. If `modifyStore` is not passed in directly, it is read from context. This only happens once. If a callback is returned from `modifyStore`, this is called on unmount.
+
+If `replaceReducer` is called in `modifyStore`, this component wraps the children with a new context-provider with an updated `storeState`. The provider is only rendered on the first render pass since `storeState` will be correct after this.
+
+**Arguments**
+
+1. `modification` is an object that will be passed into `modifyStore`
+
+2. `modifyStore` is an optional prop that allows for passing in a `modifyStore`-function directly to `<StoreModifier>`. This makes it possible to avoid having to pass it down via context and simplifying the API when SSR-support is not needed (libraries could support both versions).
+
+3. `children` are the children that needs a modified store
+
+#### `withModifiedStore(modification, options)(WrappedComponent)`
+
+This is a Higher-Order-Component that wraps the `WrappedComponent` in a `<StoreModifier>`. It additionally takes care of hoisting statics on `WrappedComponent` and can optionally `forwardRef` to the wrapped component. As opposed to using `<StoreModifier>` directly, this HOC also supports using a custom `context` (if you are using something other than the default `ReactReduxContext`).
+
+**Arguments**
+
+1. `modification` is an object that will be passed into `modifyStore`
+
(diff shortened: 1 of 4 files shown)


Ephem:
Ping @mpeyper and @abettadapur, I would love your feedback on the updated RFC!

RFC_STORE_MODIFIER.md
```diff
@@ -0,0 +1,185 @@
...
+  connect(mapStateToProps)(Greeter)
+)
+```
+
+```jsx
+import staticReducers from './staticReducers'
+
+// Example simple userland implementation
+let reducers = { ...staticReducers }
+function modifyStore({ newReducers }, storeProxy, state) {
```
abettadapur: I am not a huge fan of having users write this function like this, perhaps we can go with something like the chaining method? 

```jsx
function storeModification1(modification, storeProxy, state) {
    ///whatever
}

function storeModification2(modification, storeProxy, state) {
    ///whatever 2
}

return (
    <Provider store={store} storeModifier={composeModifiers(storeModifier1, storeModifier2)}>...</Provider>
);
```

What do you think?

abettadapur:
I think the rest of this looks quite good, thanks @Ephem!

RFC_STORE_MODIFIER.md
Ephem: I think I agree that chaining should be supported out of the box, though I'm not yet sure about the best API for that (a compose-function or supporting passing an array in)? Might be nice to support single functions as well for simple usecases.

The only unsolved issue I see with chaining is how to avoid clashes and collisions in the `modification`-object when multiple libs use it, do you have any ideas around that? :)

src/components/StoreModifier.js
```diff
@@ -0,0 +1,68 @@
...
+      const { store, storeState } = this.context
+      const modifyStore = this.props.modifyStore || this.context.modifyStore
+
+      // Create storeProxy
+      let newRootReducer
+      const replaceReducerSpy = (reducer, ...args) => {
+        newRootReducer = reducer
+        return store.replaceReducer(reducer, ...args)
+      }
+      const storeProxy = { ...store, replaceReducer: replaceReducerSpy }
```
mpeyper: Should this also include a `modifyStore` proxy to allow modifier from calling other modifiers?

```js
let storeProxy

const modifyStoreProxy = (modification) => modifyStore(modification, storeProxy, storeState)

storeProxy = { ...store, replaceReducer: replaceReducerSpy: modifyStore: modifyStoreProxy }
```

```js
function dynamicReducers(modification, storeProxy, state) {
  // whatever
}

function addAdminReducer(modification, storeProxy, state) {
  if (state.user.isAdmin) {
    storeProxy.modifyStore({ reducers: [adminReducer] })
  }
}

return (
    <Provider store={store} storeModifier={combineModifiers({ reducers: dynamicReducers, addAdminReducer })}>...</Provider>
);
```

Looking at that above example, I can see why we may not want to use a `combineReducers` like API to solve the namespacing issue.  Not all modifiers will necessarily require a slice of the `modifications` object.

RFC_STORE_MODIFIER.md
```diff
@@ -0,0 +1,185 @@
...
+
+**Arguments (automatically passed in to the function)**
+
+1. `modification` is an object that can contain any data that needs to be passed from the calling site (`<StoreModifier>`), for example reducers to be added to the store. There is no prescribed shape for this object, it is supposed to differ between implementations
+
+2. `storeProxy` here is a wrapped version of `store` that spies on `replaceReducer` to be able to safely patch the context with the new state if it is called. Maybe this should be a stripped down version of `store`, leaving out `subscribe` and possibly other methods unsafe to use here? Could also keep them in but provide warnings.
+
+3. `state` is passed in, in case it is needed for conditional logic, because `getState` is kind of unsafe here. This `state`-object contains the current state _stored on context for this render_ and is not the latest version of the state in the store.
```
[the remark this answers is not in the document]
Ephem: Well, it *can be*, but is not necessarily. I can imagine there being cases where you DO want to get the latest store state and not the one on this render. I hesitate changing the semantics of `store.getState()`, I think it would be confusing if it did not always get the latest state from store? (Unless we made it really clear that `storeProxy` is a proxy that works differently, but if we start changing the behaviour of functions to be "context-safe", the expectation would probably be that *all* functions would be safe, something I don't think we can guarantee)

I do agree it would be nice if you could get the resulting state from `replaceReducer` though. `store.getState()` after calling `replaceReducer` returns updated state, but that might contain other changes as well.

I think there is a tradeoff we need to consider carefully here, between "just use store and we'll try to patch things up magically behind the scenes" and being more explicit about what is happening. I'm starting to lean towards providing `replaceReducer` as a separate parameter instead so it is clear it is special, but also provide the unmodified `store` as an extra parameter for those that need it, but make it clear it is potentially unsafe to use. If it is clear to the user that `replaceReducer` is special, it could also return the new "safe" state.

RFC_STORE_MODIFIER.md
```diff
@@ -0,0 +1,185 @@
...
+
+**Returns**
+
+A `StoreModifier`-component that uses the `context` passed in
+
+### Unsolved questions
+
+- Can naming be improved?
+- It would be possible to support a middleware-style API for `modifyStore`, [as per this comment](https://github.com/reduxjs/react-redux/issues/1150#issuecomment-450596584). This could be implemented in userland, but it might make sense to add this as the official API to encourage modular `storeModifiers`?
+  - In that case: How do we deal with namespacing-issues in the `modifier`-object?
```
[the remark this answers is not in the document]
Ephem: I'm leaning towards **Option 1**, but with the caveat of providing advice along these lines in the documentation (better formulated..):

> When using multiple modifiers, or if you are a library author, we recommend creating utility-functions for creating the `modification`-object, for example: `addReducersModification(newReducers)`. This should work as an action-creator and could look something like this:
>
> ```jsx
> const addReducersModification = newReducers => ({
>     type: '@your-library-name/type-of-modification',
>     payload: newReducers
> });
> ```
> 
> Then make sure your modifier-functions only respond to the modifications with the correct type.

The `modification`-object really is very similar to an action, just describes a modification supposed to happen to the store instead. For actions we handle namespacing via `type`, so why not reuse that concept here?

It would of course be possible to rename `modification` to `action` instead, but I think that might be more confusing than it clarifies things?

src/components/StoreModifier.js
Ephem: Hmm, that's interesting. If we want modifiers to be truly modular and composable I think we'd need something like this. If we do add the modifiers-concept to Redux directly, that function will probably already exist on the store.

RFC_STORE_MODIFIER.md
mpeyper: My feeling is that `redux` already has a fairly solid concept for these types of things in the enhancer/middleware functionality.  I'm struggling to think of how modifiers would be different enough, or offer anything that the others don't already provide.

The only thing that is potentially missing is perhaps a more formal way of injecting values into middleware (i.e. running a new saga/observable) as each middleware currently has to implement this themselves and often they have their own, unique way of doing it.  If there was a standard API for pushing values into middleware than I think the `storeModifiers` concept goes away and then there is just a component required to call that API during rendering.

Just spitballing, but the middleware could _subscribe to modifications_ as part of their setup step:

```js

function createSagaMiddleware (options = {}) {
  function runSaga (saga) {
    // ...
  }

  return (middlewareAPI) => { // usually just called `store

    middlewareAPI.acceptModification('@@redux-saga/RUN_SAGA', (modifier) => runSaga(modifier.saga);

    return (next) => (action) => {
      // ...
    }
  }
}

const store = createStore(reducer, applyMiddleware(createSagaMiddleware()))

store.modify({ type: '@@redux-saga/RUN_SAGA', saga }) 
```

There a very strong similarity here between `actions`/`dispatch`/`subscribe` and `modifier`/`modify`/`acceptModification`, but it's important to note that they are different, both semantically and in the restrictions applied to them (e.g. actions should be serialisable, modifications would not need to be).

Ephem:
Thanks for some great comments. 👏 I wanted to add a few general thoughts around composing/chaining modifiers.

First of all, I like it! While it has some unsolved questions and can probably be improved upon, I think there might be a pretty strong and flexible API and concept here. If each library (like `redux-saga`, `redux-observable` etc) could easily provide their own modifiers, it would simplify things across the board. It's a great fit with any component model really, since often specific reducers, sagas, epics etc are only needed in parts of an application, but this is pretty hard to achieve today.

My one worry is that it might grow the API, implementation and complexity by quite a bit. It definitely feels like scope creep from the original problem. My initial feeling is that if we do want to add this, it should be added directly to Redux and that to be worth adding such functionality directly into core, the case needs to be pretty strong.

If we could further explore making a "modification" a first class citizen in the sense that they can be used to simplify the API for *creating* the store, that would definitely increase the value imo.

I'm starting to think that there are three ways forward:

1. Pause this RFC and explore the concept of adding an API around modifiers first
2. Grow this RFC and deal with both things at the same time
3. Shrink this RFC back down to only dealing with replacing reducers

This is the point where I'd really love some directional feedback from the core maintainers. 😄

RFC_STORE_MODIFIER.md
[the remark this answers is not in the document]
mpeyper: IMO, the [store has 4 functions](https://redux.js.org/api/api-reference#store-api) by default: `getState`, `dispatch`, `subscribe` and `replaceReducer`.

- `getState`: not "context-safe"
- `dispatch`: "context-safe" so long as you don't try to use the updated state
- `subscribe`: not "context-safe" but I agree we should strip it out of the `storeProxy` anyway
- `replaceReducer`: "context-safe" if `getState` returns the patched state after the reducer is added

I feel that adding the unmodified `store` but noting it's _unsafeness_ no better than accessing it from the "private" context and it will be easy to abuse and get wrong and become a footgun for users.

RFC_STORE_MODIFIER.md
Ephem: This feels like a really interesting idea! 💡 I haven't thought about it deeply, but intuitively it makes a lot of sense.

I'm starting to think that treating "modifications to middleware" and "replacing the root reducer" as the same problem and cramming them into a single API might be a mistake after all.. There is definitely worth in doing them both from the same utility-component/HOC (`<StoreModifier>` for example), but should we really think about them as the same kind of problem implementation-wise? Also see [this comment](https://github.com/reduxjs/react-redux/pull/1154#issuecomment-451144369).

mpeyper:
I agree that the core idea has grown, but I think it's necessary if we want to improve the code-splitting story in `redux`.  IMO, `redux` thrives on its ecosystem, and not at least planning for more general support will see us quickly build ourselves into a corner.

I do agree that without input and direction from the core maintainers, any more work into this is likely to be wasted.

I'm not sure you will get much buy in from @timdorr on adding a new concept to `redux` core.  [He was pretty firm](https://github.com/reduxjs/redux/issues/3321#issuecomment-450658239) about not wanting to include some fairly standard utilities and actually wanting to remove more from the core package.  I would like to hear his take on all of this, and of course @markerikson's as well.

Ephem:
I agree, but improving the code splitting story was actually a secondary objective of this RFC. 😄 My feeling is also that we *are* building ourselves into a corner with this RFC exactly because we are trying to cater too much to the entire ecosystem. 😉 (Though I like your idea of `acceptModification`..)

The original problem this RFC set out to solve was to add a minimal building block for using `replaceReducer` safely from the component tree, considering the level of complexity that was added for doing this in a futureproof way in userland in v6. While it is very debatable, I would personally argue that since `reducers` and `replaceReducer` are core concepts of Redux, the official React-bindings should provide a safe way to interact with these from React. This is also one thing that *everyone* needs for code splitting.

Another problem, which is related but separable, is how to ergonomically modify middleware, both with and without React. As you mentioned, there is currently no API in (React-)Redux for this (but several existing in userland), maybe there should be, and I'm very glad we're talking about it and exploring the space for precisely the reasons you mention.

That said, I do think this RFC is currently trying to tackle too much to get off the ground. Also, by narrowing it back down from what it is now to tackle a single problem, I think chances are better we can provide a *minimal but flexible API to build other/future solutions on top of*, this applies to both userland and any future changes to (React-)Redux. I'll give this some more thought.

Even a minimal API like that would likely take a long time to evaluate and implement, considering the uncertainties of concurrent rendering etc. Meanwhile, I will try to find some time to document the problem and solution in a more concise way to help out with userland implementations. 📜

Ephem:
I'm going to go ahead and close this PR and the related issue since there are more important things to tackle short-term and things will need time to stabilise before revisiting this.

For more details, see [the related issue](https://github.com/reduxjs/react-redux/issues/1150#issuecomment-460053367).

markerikson:
FWIW, I definitely appreciate the detailed thought that went into this idea :)  But yes, I think we've got other more important concerns atm, and there's not an immediate urgent need to try to get functionality like this directly into React-Redux itself atm.

PR #1137, opened 2018-12-20
fix: react-hot-loader compatibility

theKashey:
This PR restores _historically_ important ability to hot-reload redux.

The major change here is driven by `prettier` due to increased length of conditions, nothing more.

PR #1168, opened 2019-01-20
Updated: Support React-Hot-Loader compatibility

markerikson:
Rebased version of #1137 , by @thekashey.

Perf benchmarks indicate no noticeable difference in performance.

Tagging @timdorr , @shahzaibkhalid , and @glenjamin who were participating in the other thread.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@ test/**/lcov.info
 test/**/lcov-report
 test/react/*/test/**/*.spec.js
 test/react/**/src
+test/jest-config.json
 lcov.info
 
 lib/core/metadata.js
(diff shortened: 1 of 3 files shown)


markerikson:
And merging.

markerikson:
@timdorr : might as well go ahead and put out 6.0.1 with this and the "is it a real context?" change.

ericdfields:
Thank you!

hhaidar:
💯 to 6.0.1

GerryFudd:
Looks like react-redux 7.0.0 causes connected components to not load when used with react-hot-loader. Should that be filed as a separate ticket or should this one be re-opened?

theKashey:
Please update `react-hot-loader` to the latest version (4.9.0, released a week ago) which does support hooks hot reloading. 
Any older version would have issues with `react-redux` v7.

PR #1174, opened 2019-01-30
Upgrade to react-is v16.7.0

nobezawa:
I wish to upgrade react-is.

because `hoist-non-react-statics` is dependent react-is v16.7.0
https://github.com/mridgway/hoist-non-react-statics/blob/master/package.json#L27

error occurs this line.
https://github.com/facebook/react/blob/master/packages/react-is/src/ReactIs.js#L126


timdorr:
Sure, thanks!

PR #1180, opened 2019-02-09
Update react-router usage example

BTMPL:
The example is incorrect - you can't mix `children` and `component` prop. Looks like some left-over from v3 API?


timdorr:
Yep. Thanks!

PR #1189, opened 2019-02-14
Delay state read in Provider's subscriber

Andarist:
Thinking about how and in which order things are handled it should possibly result in more setState bailouts (so it should be slightly improve things)


markerikson:
Appreciate the tweak, but this is almost definitely going to be going away anyway per the discussion in #1177 .

Andarist:
Ok, sure - but the timing of the new release is unsure, should it block any other developments on the stable branch? Just discussing this out of curiosity about your point in view, I don't particularly care about this patch - it was just something I've noticed while reading the source.

PR #1190, opened 2019-02-15
Remove redundant checks for static options which cant change in the l…

Andarist:
…ifetime of a component

From what I see those options are completely static, they are never reassigned so it doesnt make sense to track their last values.


timdorr:
They would change during hot reloads, AFAIK. Let's leave them for now.

PR #1195, opened 2019-02-26
adding a react hooks test that fails with v7 alpha

saboya:


markerikson:
Thanks, really appreciate the effort!

markerikson:
One immediate observation:

With hooks, all interactions that would cause an update need to be wrapped in `act()`.  Specifically in this case, the `store.dispatch()` call needs to be wrapped in `rtl.act()`, like:

```js
rtl.act(() => {
    store.dispatch({ type: 'FOO' })
})
```

Does that change the test result here?

(Working on another task atm, so can't try it out myself right this second.)

saboya:
Thanks, that actually fixes the test to make it behave exactly like the test case (the spies weren't being called frequently enough).

But the render delta between v5/v6 compared to v7 still exists, and v7 still fails the test.

markerikson:
FWIW, counting number of `mapState` calls is always fragile, because that can (and has) changed between versions multiple times.  Yes, I know a bunch of our tests use that approach as a metric, and if you look back at the history of `connect.spec.js` you'll see that those keep getting tweaked as the implementation has changed.

But yeah, this should be a starting point for me to dig in further.  Thanks!

saboya:
I reduced the test down to 2 components. I was curious to see if it had anything to do with having a non-connected component in between 2 connected components had anything to do with it, and turns out it's (apparently) only the `useEffect` tha causes the issue.

markerikson:
I cherry-picked this into the v7-alpha and v7-beta branches, so the PR is obsolete but the commits are being used.  Thanks for filing this!

PR #1196, opened 2019-02-26
Fix perf problems with out-of-bounds array access

Andarist:
Based on the fact that this might be a perf problem according to v8 team ( https://github.com/babel/babel/pull/6582 )

### DIFF
--- a/src/components/connectAdvanced.js
+++ b/src/components/connectAdvanced.js
@@ -12,6 +12,8 @@ import Subscription from '../utils/Subscription'
 
 import { ReactReduxContext } from './Context'
 
+const EMPTY_ARRAY = []
+
 const stringifyComponent = Comp => {
   try {
     return JSON.stringify(Comp)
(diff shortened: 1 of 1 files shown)


src/components/connectAdvanced.js
```diff
@@ -20,6 +22,13 @@ const stringifyComponent = Comp => {
   }
 }
 
+function storeStateUpdatesReducer(state, action) {
```
Andarist: i've hoisted this as this didnt depend on any closure variables, so should be better to reuse a single function for this rather than to create a copy for each connect

timdorr:
Thanks, Mateusz. Unfortunately, we're going to be going full-hooks for a v7 bump and will be dropping this branch. So, merging it won't really do much.

markerikson:
@timdorr : note that this PR is actually targeted at my v7 hooks branch, not `master`.

That said, I already pulled the commits in via cherry-picking locally so I could experiment with them:

https://github.com/reduxjs/react-redux/commit/5f60435b71b5664861e907bd4c57899088cad3e4

Not sure if it made any actual difference, but doesn't seem like it should hurt anything.

Thanks, @Andarist !

PR #1197, opened 2019-03-02
Add first SSR integration-tests

Ephem:
As requested in #1177 this PR adds a few tests verifying behaviour with serverside rendering. It's in no way exhaustive, but rather a start.

It adds a new folder `/test/integration` where other future integration-tests could live as well (such as for dynamic injection of reducers).

All tests except `should render children with original state even if actions are dispatched in ancestor` are passing in `v5.1.1` as well, I have not tested with the v7-alpha, but I suspect the same is true there.

I am very much open to feedback on how to improve these and ideas for more tests to add. :)

### DIFF
--- a/test/install-test-deps.js
+++ b/test/install-test-deps.js
@@ -12,7 +12,11 @@ readdirSync(join(__dirname, 'react')).forEach(version => {
     console.log(`skipping ${version}, ${reactVersion} was specified`)
     return
   }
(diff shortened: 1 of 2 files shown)


markerikson:
Great!  See, this is exactly the kind of stuff I wouldn't have been able to come up with myself.

I'm hoping to dig in further on both the v7 and v6.1 branches this weekend, and will look at pulling these in.

timdorr:
LGTM. Thanks!

PR #1198, opened 2019-03-04
Change CommonJS Broken Docs Link in the Readme File

mytuny:
The Readme file currently referencing an old Webpack documentation link for the CommonJS module that is no longer available (http://webpack.github.io/docs/commonjs.html).
I suggest to change it by this => [https://webpack.js.org/api/module-methods/#commonjs](https://webpack.js.org/api/module-methods/#commonjs)


timdorr:
Thanks!

PR #1201, opened 2019-03-07
6.x - Refactor to fix performance

salvoravida:
### Abstract
This PR should fix performance issue with new 6.x, and restore it to 5.x speed 

- ### Remove getDerivatedStateFromProps

First of all we should have a look to React Internals: 

```javascript
// Invokes the update life-cycles and returns false if it shouldn't rerender.
function updateClassInstance(current, workInProgress, ctor, newProps, renderExpirationTime) {
  var instance = workInProgress.stateNode;

 //...................

  if (typeof getDerivedStateFromProps === 'function') {
    applyDerivedStateFromProps(workInProgress, ctor, getDerivedStateFromProps, newProps);
    newState = workInProgress.memoizedState;
  }

  var shouldUpdate = checkHasForceUpdateAfterProcessing() || checkShouldComponentUpdate(workInProgress, ctor, oldProps, newProps, oldState, newState, nextContext);

  /// ......

  return shouldUpdate;
}
```

so we can replace
```javascript
static getDerivedStateFromProps (nextProps){
   doSomeCalculationAndMemoizeToFakeState(nextProps)
}
 ```
with 
```javascript
shouldComponentUpdate(nextProps) {
     let update false;
        if (nextProps !== this.props)  {
            update = doSomeCalculation(nextProps)
         }
        return update
      }
```
that for react-redux use case (calculate new mapStateToProps when ownProps may have changed),
is safe.

- ### use static contestType propagation => no double wrapper 
With this pr i have removed double wrapper for context propagation and i use only static class contextType.

**Constraints :** 
**_Custom context can be assigned only on connect options_** 
```javascript
connect(mapState,mapDispat, { context:CustomContext })(Component)
```
tbh i think that context on props should be removed as it  is a duplication of connect options.

Anyway there are some alternative 💯 
1)Just use a different ConnectedComponet for every context you need, like:
```javascript
const MyConnectedComponentWithContext1= connect(mapState,mapDispat, { context:CustomContext1 })(Component)
const MyConnectedComponentWithContext2= connect(mapState,mapDispat, { context:CustomContext2 })(Component)
```
2) use read_context from ReactInternals to allow use context as props on 
```javascript
<ConnectedComponent context={CustomContext} />
```
you can activate it with unstable_enableReadContextFromProps = true  on connect options

```javascript
const ConnectedComponent = connect(mapState,mapDisp, { unstable_enableReadContextFromProps:true })(Component)
```
I already use it successfully  on https://github.com/salvoravida/react-universal-hooks
for `useClassContext` (useContext for classes) here https://github.com/salvoravida/react-class-hooks/blob/master/src/core/useClassContext.js

```
┌───────────────┬─────────┬──────────────┬───────────┬───────────┬──────────┬────────────────┐
│ Version       │ Avg FPS │ Render       │ Scripting │ Rendering │ Painting │ FPS Values     │
│               │         │ (Mount, Avg) │           │           │          │                │
├───────────────┼─────────┼──────────────┼───────────┼───────────┼──────────┼────────────────┤
│ 5.1.1         │ 29.72   │ 69.3, 0.0    │ 2926.76   │ 1231.81   │ 516.42   │ 29,30,31,31    │
├───────────────┼─────────┼──────────────┼───────────┼───────────┼──────────┼────────────────┤
│ 5.2.1         │ 59.36   │ 68.9, 1.0    │ 1487.73   │ 1902.74   │ 970.93   │ 60,59,60,60    │
├───────────────┼─────────┼──────────────┼───────────┼───────────┼──────────┼────────────────┤
│ 6.0.0         │ 28.04   │ 66.8, 2.1    │ 3065.45   │ 1200.56   │ 436.10   │ 27,29,28,29,29 │
├───────────────┼─────────┼──────────────┼───────────┼───────────┼──────────┼────────────────┤
│ 6.0.1-classes │ 26.00   │ 71.2, 1.5    │ 2989.46   │ 1135.38   │ 522.30   │ 26,26          │
├───────────────┼─────────┼──────────────┼───────────┼───────────┼──────────┼────────────────┤
│ 6.2.1         │ 59.36   │ 68.6, 1.0    │ 1429.56   │ 1906.84   │ 989.53   │ 60,59,60,60    │
├───────────────┼─────────┼──────────────┼───────────┼───────────┼──────────┼────────────────┤
│ 7.0.0.alpha-3 │ 59.68   │ 73.1, 1.1    │ 1606.08   │ 1863.47   │ 945.96   │ 60,59,60,60    │
└───────────────┴─────────┴──────────────┴───────────┴───────────┴──────────┴────────────────┘
```
Test Suites: 5 passed, 5 total

- ### React peer version >16.6.0 (static contextType) 

if you want to test this PR right now you can use Yarn alias npm
```
   "react-redux": "npm:@salvoravida/react-redux@^6.2.1"
```
Your feedback is highly appreciated .... :D

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,4 @@ website/build/
 website/yarn.lock
 website/node_modules
 website/i18n/*
+.idea/
--- a/package.json
+++ b/package.json
@@ -36,7 +36,8 @@
     "coverage": "codecov"
   },
   "peerDependencies": {
-    "react": "^16.4.0-0",
+    "react": "^16.6.0-0",
+    "react-dom": "^16.6.0-0",
     "redux": "^2.0.0 || ^3.0.0 || ^4.0.0-0"
   },
   "dependencies": {
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -9,13 +9,14 @@ const env = process.env.NODE_ENV
 
 const config = {
   input: 'src/index.js',
-  external: Object.keys(pkg.peerDependencies || {}),
+  external: ['react', 'redux', 'react-dom'], //Object.keys(pkg.peerDependencies || {}),
   output: {
     format: 'umd',
     name: 'ReactRedux',
     globals: {
       react: 'React',
-      redux: 'Redux'
+      redux: 'Redux',
+      'react-dom': 'ReactDOM'
     }
   },
   plugins: [
@@ -32,7 +33,8 @@ const config = {
         'node_modules/react-is/index.js': [
           'isValidElementType',
           'isContextConsumer'
-        ]
+        ],
+        'node_modules/react-dom/index.js': ['unstable_batchedUpdates']
       }
     })
   ]
--- a/src/components/Provider.js
+++ b/src/components/Provider.js
@@ -1,6 +1,7 @@
 import React, { Component } from 'react'
 import PropTypes from 'prop-types'
 import { ReactReduxContext } from './Context'
+import Subscription from '../utils/Subscription'
 
 class Provider extends Component {
   constructor(props) {
@@ -8,56 +9,47 @@ class Provider extends Component {
 
     const { store } = props
 
+    this.notifySubscribers = this.notifySubscribers.bind(this)
+    const subscription = new Subscription(store)
+    subscription.onStateChange = this.notifySubscribers
+
     this.state = {
-      storeState: store.getState(),
-      store
+      store,
+      subscription
     }
+
+    this.previousState = store.getState()
   }
 
   componentDidMount() {
     this._isMounted = true
-    this.subscribe()
+
+    this.state.subscription.trySubscribe()
+
+    if (this.previousState !== this.props.store.getState()) {
+      this.state.subscription.notifyNestedSubs()
+    }
   }
 
   componentWillUnmount() {
     if (this.unsubscribe) this.unsubscribe()
+
+    this.state.subscription.tryUnsubscribe()
 
     this._isMounted = false
   }
 
   componentDidUpdate(prevProps) {
     if (this.props.store !== prevProps.store) {
-      if (this.unsubscribe) this.unsubscribe()
-
-      this.subscribe()
+      this.state.subscription.tryUnsubscribe()
+      const subscription = new Subscription(this.props.store)
+      subscription.onStateChange = this.notifySubscribers
+      this.setState({ store: this.props.store, subscription })
     }
   }
 
-  subscribe() {
-    const { store } = this.props
-
-    this.unsubscribe = store.subscribe(() => {
-      const newStoreState = store.getState()
-
-      if (!this._isMounted) {
-        return
-      }
-
-      this.setState(providerState => {
-        // If the value is the same, skip the unnecessary state update.
-        if (providerState.storeState === newStoreState) {
-          return null
-        }
-
-        return { storeState: newStoreState }
-      })
-    })
-
-    // Actions might have been dispatched between render and mount - handle those
-    const postMountStoreState = store.getState()
-    if (postMountStoreState !== this.state.storeState) {
-      this.setState({ storeState: postMountStoreState })
-    }
+  notifySubscribers() {
+    this.state.subscription.notifyNestedSubs()
   }
 
   render() {
--- a/src/components/connectAdvanced.js
+++ b/src/components/connectAdvanced.js
@@ -25,6 +25,7 @@ function makeSelectorStateful(sourceSelector, store) {
   const selector = {
     run: function runComponentSelector(props) {
       try {
+        selector.lastProcessedProps = props
         const nextProps = sourceSelector(store.getState(), props)
         if (nextProps !== selector.props || selector.error) {
           selector.shouldComponentUpdate = true
@@ -286,6 +287,8 @@ export default function connectAdvanced(
 
       render() {
         const selector = this.selector
+        //forceUpdate from external
+        if (selector.lastProcessedProps !== this.props) selector.run(this.props)
         selector.shouldComponentUpdate = false
 
         //clean wrapperProps and set ref if forwardRef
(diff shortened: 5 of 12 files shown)


salvoravida:
Some points  about customContext and React-Redux

### 99% You do NOT need customContext on React-Redux
In 99% of Apps you have One Store @ App Root.

If it is your case just use 6.x the SAME EXACT WAY you have already used 5.x
 ```javascript
<Provider store={store}/>
  <App>
   <ConnectedComp />
 </App>
</Provider >
//where
const ConnectedComp = connect(mapState,mapDisp)(Component);
```
In this case  you can think to 6.x as Exact 5.x with deprecated Warning removed (deprecated lyfecicles and legacyContext) on React 16 so it is React 17 Ready.

### 1% You May need customContext on React-Redux
In 1% of Apps you have to use Two or More Store on the same app.

If it is your case, in both 6.0.0 and 6.2.1 you can have this code
 ```javascript
<Provider store={store} context={CustomContext1}/>
  <App>
       <Provider store={store2} context={CustomContext2}/>
           <ConnectedComponentWithContext1/>
           <ConnectedComponentWithContext2/>
        </Provider > 
  </App>
</Provider >
//where
const ConnectedComponentWithContext1= connect(mapState,mapDispat, { context:CustomContext1 })(Component)
const ConnectedComponentWithContext2= connect(mapState,mapDispat, { context:CustomContext2 })(Component)
```
Also in 6.0.0 you can write an EQUIVALENT way that is no more supported (for performance issue) on 6.2.1

 ```javascript
<Provider store={store} context={CustomContext1}/>
  <App>
       <Provider store={store} context={CustomContext2}/>
           <ConnectedComponent context={CustomContext1} />
           <ConnectedComponent context={CustomContext2}/>
        </Provider > 
  </App>
</Provider >

//where
const ConnectedComponent= connect(mapState,mapDispat)(Component)
```

So tbh, i think that 2x performance is better than have 2 identical unuseful way to do the same (rare-case) thing.

markerikson:
Just now having a chance to look at this.

First: @salvoravida , thanks for filing the PR.  I do appreciate that you've taken time to work on this.

However, having said that: I'm _not_ happy that you've apparently gone around and advertised this in multiple places (including Reddit!), without us having had a chance to evaluate and discuss it first.  

As one of the largest libraries in the React ecosystem, we have to carefully consider the needs of our users.  That includes both what they need to do with the library, how we maintain it, and how we "advertise" changes.  In particular, we don't want people getting confused about what is and isn't available, and what's actually "official".

We haven't even seriously publicized the v7 alpha at this point, and I've deliberately published it as a personal scoped package ( `@acemarke/react-redux` ) rather than an actual `react-redux` version just to keep it clear that this isn't ready yet.  So, running around advertising a PR that you _just_ filed, before we even have any idea whether it _might_ be something we would consider, is not helpful.

markerikson:
I'll need time to sit down and actually dig into the specific changes you made here.  However, glancing over things, I immediately see at least three problems that make this PR unlikely to be accepted as-is:

1. Doing work that has side effects in `shouldComponentUpdate` is not correct.
2. Bumping the React peer dependency from 16.4 to 16.6 would require a corresponding **major** version bump for React-Redux.  That would mean this could not be a v6.2.0, it would have to be a v7.
3. Similarly, it looks like you've removed the ability to pass `context` as a prop to a component.  Whether you like that or not, it's part of our public API for v6.  Removing that _is_ a breaking change, and would also force this to be a major version bump.

salvoravida:
> I'll need time to sit down and actually dig into the specific changes you made here. However, glancing over things, I immediately see at least three problems that make this PR unlikely to be accepted as-is:
> 
> 1. Doing work that has side effects in `shouldComponentUpdate` is not correct.
> 2. Bumping the React peer dependency from 16.4 to 16.6 would require a corresponding **major** version bump for React-Redux.  That would mean this could not be a v6.2.0, it would have to be a v7.
> 3. Similarly, it looks like you've removed the ability to pass `context` as a prop to a component.  Whether you like that or not, it's part of our public API for v6.  Removing that _is_ a breaking change, and would also force this to be a major version bump.
> 
> So, overall, this PR doesn't appear to fit within the constraints we have to deal with.
> 
> As an additional thought: given what we know about how state propagation works via context (v6) vs how direct subscriptions work, I could plausibly argue that that behavior change itself would justify a major version bump (and thus any use of direct subscriptions could only happen in a v7 rather than a v6.x).

1) I'm sorry but i do not agree at all:
The theorical problem is: **shouldUpdate ? selectorAndMerge(s,a)!=selectorAndMerge(s,b)**
thats are pure function executed sinc ,( with partial value memoization), and without request/enqueque any callback/future/events, so there are NO side effects. And it is a fact, not a my opinion. 
Moreover the natural lifecycle where evaluate that problem is of course the lifecycle that has the same name!
2) It is a political cost/benefit decision. 
3) Same as 2), but it could resolved via readContext Internal api, if it is needed 

That said, i do some advs for getting real word tests on big apps to test real performance improvment,
as this problem affects many users, and this is the normal behaviour on open-source projects.

Regards.

markerikson:
Bluntly, I'm a maintainer, and it's my decision that matters here.

I previously asked [whether bumping a peer dependency (and specifically React) would require a major version change](https://twitter.com/acemarke/status/1087062667007848449), and was told in no uncertain terms that **yes, it requires a semver major bump**.

I agree that `readContext()` is a potential alternative solution in some ways.  Unfortunately, that was removed from React's public API.  In that PR, the React team said it was theoretically possible for folks to reach directly into the internals of the dispatcher to do that themselves, which is what it appears your PR does.  However, I'm definitely uncomfortable shipping code that solely depends on undocumented internals, however "well-known" that might be.  It's bad enough that `batchedUpdates()` is still considered "unstable", but at least there we have public comments from the React team that it's a widely used API, not actually likely to be changed or removed any time soon, and they specifically encouraged us to use it.

timdorr:
Yeah, `__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED`? I don't want to get fired :(

timdorr:
Thank you for the PR, but this isn't the direction we're going to go.

salvoravida:
> Thank you for the PR, but this isn't the direction we're going to go.

it's a pity, now all tests passes and is still 2.5x faster.

Regards

PR #1205, opened 2019-03-17
Remove batch arg from createListenerCollection

migueloller:
This prevents a bug with Terser (webpack's default minifier) where the
returned batch function isn't defined due to function inlining.


markerikson:
Sweet, thanks for helping debug this and figure out what was going on!

I'll push out `alpha.5` shortly.

PR #1211, opened 2019-03-22
[v7-beta] Add test for dynamically injecting reducers

Ephem:
As asked for in #1177, this PR contains a new integration test-file `dynamic-reducers.spec.js` with an implementation for dynamically injecting reducers during the render phase with one test for the client and one for the server.

Right now it only tests the basic expectation that the wrapped child renders with the initial state from the dynamically injected reducer, this seemed most valuable to me to start with, but can be easily expanded upon. With more tests and tweaks, I guess this implementation could grow/morph to become some form of semi-documented best practice for dynamic reducer injection as long as no public APIs exists.

The reason we do this as a side effect in the render phase is to support SSR, see the comment in the test file for slightly more details. `redux-dynamic-modules` does something similar to this, but executes the side effect in `construct` instead, I tried to simplify by placing all logic in render. Instead of relying on the constructor only running once, this has guards to determine when injection and possible patching of `storeState` needs to happen or not during rendering.

By only patching `storeState` if it exists and otherwise relying on connected child components to fetch an updated state themselves, this test works both with `v6` and the current `v7-beta`.

Do note that this is a pretty fragile test that relies on internal and undocumented implementation details of Redux (and probably React), but as @markerikson pointed out, it's hopefully better than nothing. 🤷‍♂️

I am very open to feedback on how to improve this further, besides from maintainers, I would love feedback from @abettadapur and @mpeyper 😄

### DIFF
--- a/test/integration/dynamic-reducers.spec.js
+++ b/test/integration/dynamic-reducers.spec.js
@@ -1,0 +1,165 @@
+/*eslint-disable react/prop-types*/
+
+import React from 'react'
+import ReactDOMServer from 'react-dom/server'
+import { createStore, combineReducers } from 'redux'
+import { connect, Provider, ReactReduxContext } from '../../src/index.js'
+import * as rtl from 'react-testing-library'
+
+describe('React', () => {
+  /*
+    For SSR to work, there are three options for injecting
+    dynamic reducers:
+
+    1. Make sure all dynamic reducers are known before rendering
+       (requires keeping knowledge about this outside of the
+       React component-tree)
+    2. Double rendering (first render injects required reducers)
+    3. Inject reducers as a side effect during the render phase
+       (in construct or render), and try to control for any
+       issues with that. This requires grabbing the store from
+       context and possibly patching any storeState that exists
+       on there, these are undocumented APIs that might change
+       at any time.
+
+    Because the tradeoffs in 1 and 2 are quite hefty and also
+    because it's the popular approach, this test targets nr 3.
+  */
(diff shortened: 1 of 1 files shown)


test/run-tests.js
```diff
@@ -1,7 +1,7 @@
 const npmRun = require('npm-run')
 const fs = require('fs')
 const path = require('path')
-const LATEST_VERSION = '16.6'
+const LATEST_VERSION = '16.8'
```
Ephem: I included this in the PR since this version mismatches with the one in the `install-test-deps.js`-script on master, creating problems when running tests.

test/integration/dynamic-reducers.spec.js
```diff
@@ -0,0 +1,165 @@
...
+              <DynamicGreeting />
+            </ExtraReducersProvider>
+          </InjectReducersContext.Provider>
+        </Provider>
+      )
+
+      getByText('Hello world')
+      getByText('Hello dynamic world')
+    })
+    it('should render child with initial state on the server', () => {
```
Ephem: Sadly the test-environment can only differ between files, not single tests. If you think it makes more sense, I'll refactor the implementation into a util-file and move this test over to the SSR-suite where the true node test-environment is running, or a new `dynamic-reducers.server.spec.js`.

timdorr:
Can we base this on the v7-beta branch instead?

Ephem:
Done! I suspected you might want that so I had it prepared. 😄

timdorr:
Perfect. Thanks!

PR #1212, opened 2019-03-22
Add a test to verify batched dispatches result in a single render

lecstor:
Test file may need renaming/relocating.

Test does two separate rtl.act wrapped dispatches to verify that they result in 2 renders then repeats this wrapped in `batch`.


lecstor:
@markerikson I thought I'd have a crack at this. `batch` looks good to me. Let me know what changes I need to make, or if I've missed the mark completely.

test/utils/batch.spec.js
```diff
@@ -0,0 +1,94 @@
...
+import * as rtl from 'react-testing-library'
+import 'jest-dom/extend-expect'
+
+describe('React', () => {
+  describe('batch', () => {
+    afterEach(() => rtl.cleanup())
+
+    it('should render only once for batched dispatches', () => {
+      const store = createStore((state, action) => {
+        let newState =
```
markerikson: Please extract the reducer as a separate variable, and use ES6 default arguments to initialize the state, like:

```js
const initialState = {a : 1};
const reducer = (state = initialState, action) => {}
```

markerikson:
Seems like a good first check at behavior.  

Left one edit request for the code structure.

Also, could you add a second test for the `alternate-renderers` entry point? something like:

```js
import {batch as defaultBatch} from "../../../something/alternate-renderers";

// later
defaultBatch(() => {
    dispatch(increment())
    dispatch(increment())
})

expect(numberOfRenders).toBe(2)
```

lecstor:
I see what you did there.. that `something` is more elusive than expected.

```
import { Provider as ProviderMock, batch, connect } from '../../src/index.js'
import { batch as defaultNoopBatch } from '../../src/alternate-renderers.js'
```
`Cannot find module '../../src/alternate-renderers.js' from 'batch.spec.js'`

lecstor:
tests indicate that `defaultNoopBatch` is just as effective as the real thing.. I'm guessing the tests are flawed.

lecstor:
copying the `defaultNoopBatch` function into the test and removing the import showed it worked as expected so with the combined tests/imports it seems the `defaultNoopBatch` was actually the real thing. Splitting the tests into separate files has resolved this.

lecstor:
Should we be testing the render props are as expected, or is that out of scope?
eg
```
      expect(componentSpy).toHaveBeenLastCalledWith(
        {
          bar: 2,
          dispatch: store.dispatch,
          foo: 2
        },
        {}
      )
```

timdorr:
Is this dead?

lecstor:
I'm happy to make changes if needed, I was just waiting for feedback.

markerikson:
I'll try to look at this over the weekend

timdorr:
Hey @lecstor, I'm going to close this out since it's gotten a bit messy. Feel free to reopen it with a fresh branch and I'll take a look ASAP.

PR #1213, opened 2019-03-22
[v7-beta] Update outdated SSR-test - dispatch in ancestors

Ephem:
The test `should render children with original state even if actions are dispatched in ancestor` was only meant to verify current behaviour in `v6`, which has changed in the `v7-beta`. This PR reactivates and renames it `should render children with updated state if actions are dispatched in ancestor` and updates it to match current behaviour (which is the same as in v5 and earlier).

**Note:** No one should ever rely on this behaviour, dispatching actions during the render phase is almost guaranteed to be a bug. As before, this test is only meant to alert us to if internal implementation has changed this specific behaviour.

### DIFF
--- a/test/integration/server-rendering.spec.js
+++ b/test/integration/server-rendering.spec.js
@@ -28,6 +28,28 @@ describe('React', () => {
       </div>
     )
 
+    class Dispatcher extends React.Component {
+      constructor(props) {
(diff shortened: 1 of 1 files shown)


markerikson:
Thanks!

PR #1215, opened 2019-03-23
add test to verify that mapStateToProps is always called with latest store state

MrWolfZ:
As discussed in [my comment](https://github.com/reduxjs/react-redux/issues/1177#issuecomment-475418993), my wrong proposed change (to use `previousStateUpdateResult.latestStoreState` when computing `actualChildProps`) passes all existing tests. This PR adds a test which fails with that suggested change applied, but succeeds otherwise.

The test makes two changes to the store, but only updates the resulting props for the first change. When `mapStateToProps` is then called due to a wrapper props change, the test ensures `mapStateToProps` sees the latest store state (which is not the case with my proposed change).


timdorr:
I don't see any reason not to include this. Thanks!

PR #1229, opened 2019-04-11
downgrading peer dependency to 16.8.0

alexreardon:
Looking at the [changelog for `React`](https://github.com/facebook/react/blob/master/CHANGELOG.md), I do not see why `react-redux` cannot have a more lenient peer dependency version.

This can be released in a `patch` release as it is a more expansive peer dependency range.

Without this change, `react-beautiful-dnd` would need to have a react peer dependency of `^16.8.4` to avoid potential warnings. Otherwise, it could use `^16.8.0`

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ Performant and flexible.
 
 ## Installation
 
-React Redux requires **React 16.8.4 or later.**
+React Redux requires **React 16.8.0 or later.**
 
 ```
 npm install --save react-redux
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "coverage": "codecov"
   },
   "peerDependencies": {
-    "react": "^16.8.4",
+    "react": "^16.8.0",
     "redux": "^2.0.0 || ^3.0.0 || ^4.0.0-0"
   },
   "dependencies": {
--- a/test/react/16.8/package.json
+++ b/test/react/16.8/package.json
@@ -2,8 +2,8 @@
   "private": true,
   "devDependencies": {
     "create-react-class": "^15.6.3",
-    "react": "^16.8.4",
-    "react-dom": "^16.8.4"
+    "react": "^16.8.0",
+    "react-dom": "^16.8.0"
   },
   "dependencies": {
     "jest-dom": "^3.1.2",


markerikson:
Given that 16.8.6 is already out, and 16.9 is on the way, is this really that much of an issue?

If someone's not on 16.8 at all, they shouldn't be using v7.  If they're on 16.8.1 or something, should be easy to bump to 16.8.6.

alexreardon:
This is about keeping libraries and builds happy. Given that nothing is specifically being relied on in `16.8.4` then a more generous `16.8.0` makes sense as a peer dependency. As it stands, if a library states that they require `^16.8.0` then technically introducing a dependency on `react-redux` which requires `16.8.4` would be a version mismatch. In extreme cases, this might lead to a double bundling of react

alexreardon:
I think as a library declaring a peer dependency it makes sense to take the lower `16.8` to avoid potential clashes, warnings and issues

alexreardon:
If an existing library had a peer dep of ^16.8.0 then they would need to make a breaking change to get to ^16.8.4 to safely use react-redux v7. Safely from the perspective of package managers and bundlers

timdorr:
A peer deep will never cause a duplicate version issue. That's not how they work. 

This is no different from any other part of the version string mismatching between libraries. It would happen the same if we started requiring 16.9.0.

We haven't ever tested the code on versions in this range, so saying we support it would be incorrect. Sorry, but this isn't something we should accept.

alexreardon:
> A peer deep will never cause a duplicate version issue

I cannot speak for all bundlers, but it would definitely cause a big old warning for sure.  

> We haven't ever tested the code on versions in this range

Isn't this what the update in the `test/react/16.8/package.json` would do - run the tests against `16.8.0`? And if not, the `devDependency` for the project is greater than `16.8.4` so it would not be using that particular version anyway 🤔

JCheungX:
react native is using strict react `16.8.3`, 
So I cant upgrade to this project without having warning and given 6.0 has performance issue. I only can use 5.x even when i am using latest version fo react native.

tonnenpinguin:
I'm running into the same issue as @JCheungX 

Expo (still) requires us to strictly use `16.8.3` and because the mobile app lives in the same umbrella project that the web apps it's very hard (as in I gave up after a week) to not run into duplicate react version issues. :(

PR #1231, opened 2019-04-11
Export subscription to improve extensibility

badeleux:
There are some cases where it's beneficial to define your own `Provider`, which defines custom logic for store subscription.

For example if your RN app uses native navigation (https://github.com/wix/react-native-navigation), there is no point maintaining store subscription for components which are no longer on screen. Therefore it's beneficial to unsubscribe when screen _disappears_ but it's still mounted.

In order to build your own Provider one needs to have access to `Subscription` class.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -3,9 +3,17 @@ import connectAdvanced from './components/connectAdvanced'
 import { ReactReduxContext } from './components/Context'
 import connect from './connect/connect'
 
+import Subscription from './utils/Subscription'
 import { setBatch } from './utils/batch'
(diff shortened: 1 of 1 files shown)


josepot:
Hi @badeleux !

TBH I would much rather not to do that as I think that would be exposing an implementation detail. Perhaps a better option could be to move that `Subscription` into a separate repo? And then react-redux can have it among its dependencies, and you could access it as-well? (at your own risk of course, because that wouldn't be part of react-redux API).

markerikson:
Yeah, given how our internals have changed over time, I'd really rather not expose this in case we have to change things again.

Issue #1226, opened 2019-04-10
mapStateToProps called when component not due to render


justincrow:
Hi,
I have a conditionally displayed component, based on a value in redux. All components below are `connect`ed.

eg in a wrapper component's render method: 
 ` { (showValueInRedux) ? <ComponentUsingShowValue/> : <EmptyComponent/> }`

`ComponentUsingShowValue` has a `mapStateToProps` that uses redux state that it expects to be there, hence we guard for that with `showValueInRedux` in the wrapper component.

In version 6, `mapStateToProps` in `ComponentUsingShowValue` is only called when that component is due to render.

In version 7, `mapStateToProps` is called even if the wrapping component has switched to choosing `EmptyComponent` due to `showValueInRedux` being false (after having once been true).

This causes an exception in our current code because the redux values that `mapStateToProps` uses ( in the now non-rendered component) are not available.

Please could you clarify this change in behaviour?

timdorr:
We still ensure top-down update ordering. It is one of our biggest challenges with this library. AFAIK, we didn't change this behavior in 7.0.

Can you provide an example of what you're seeing in a Codesandbox? We need a reproducible test case to work with in order to figure out what's going on here.

Andrekra:
I can confirm we're having the same issue (effectively breaking our application when upgrading from v6 to v7)

Here's an simple example (mind the typo's):
```
const shouldRender = WrappedComponent => props => props.shouldRender ? <WrappedComponent {...props} /> : null

const MyComponent = compose(
  connect(state => ({ shouldRender: Boolean(state.page) })),
  shouldRender,
  connect(state => ({ title: state.page.title }))
)(MyBaseComponent)

<MyComponent />

store.dispatch({ type: 'CLEAR_STATE' }) // => Sets state.page to null
```

markerikson:
@Andrekra : as Tim said, it would really help if you can provide a project that reproduces the issue, preferably as a CodeSandbox.

iamandrewluca:
I also encountered this problem after upgrading to `^7.0`.  
After some analysis with my colleague we understood why this happened.  
But did not understood why this was not happened before.  
In our case this is how JavaScript scopes works.

https://codesandbox.io/s/p9mp9rqq4m

Andrekra:
@markerikson Here's an example https://codesandbox.io/s/v629pkllj0

MrWolfZ:
[...]

PR #1235, opened 2019-04-12
fix timing issue with setting up store subscription inside a connected component

MrWolfZ:
This is a fix for #1226.

### DIFF
--- a/src/components/connectAdvanced.js
+++ b/src/components/connectAdvanced.js
@@ -291,7 +291,7 @@ export default function connectAdvanced(
       })
 
       // Our re-subscribe logic only runs when the store/subscription setup changes
-      useEffect(() => {
+      useIsomorphicLayoutEffect(() => {
         // If we're not subscribed to the store, nothing to do here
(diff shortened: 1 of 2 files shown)


timdorr:
Perfect. Thanks!

markerikson:
Thanks.  Given that there's a few other things that have popped up too, I'd like to investigate a bit further over the weekend before pushing out a new release.

Andrekra:
I'm still having the issues. Although not anymore in my codesandbox example (mentioned here https://github.com/reduxjs/react-redux/issues/1226#issuecomment-482477365), my real life app does. It uses the same pattern, but with way more nesting. For version 7 it was working, and after adding some console.logs, I see the same error pattern (The connect of the a child components triggers even when parent prevents rendering). I'll investigate further and see if I can create a codesandbox of the problem.

justincrow:
We are also experiencing the same issue with 7.0.2. I've tried to re-create with a minimal example in codesandbox, but I cannot get the same error. I'll keep trying.

PR #1251, opened 2019-04-22
add `deps` to `useSelector` to allow controlling stability of selector

MrWolfZ:
This PR adds a `deps` parameter to `useSelector` as discussed [here](https://github.com/reduxjs/react-redux/pull/1248#issuecomment-485335594).

It also fixes an issue with staleness of the selector.

### DIFF
--- a/src/hooks/useSelector.js
+++ b/src/hooks/useSelector.js
@@ -18,8 +18,14 @@ const useIsomorphicLayoutEffect =
 /**
  * A hook to access the redux store's state. This hook takes a selector function
  * as an argument. The selector is called with the store state.
+ *
+ * This hook takes a dependencies array as an optional second argument,
+ * which when passed ensures referential stability of the selector (this is primarily
+ * useful if you provide a selector that memoizes values).
  * 
  * @param {Function} selector the selector function
+ * @param {any[]} deps (optional) dependencies array to control referential stability
+ * of the selector
  * 
  * @returns {any} the selected state
  *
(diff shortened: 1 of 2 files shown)


test/hooks/useSelector.spec.js
```diff
@@ -165,15 +165,40 @@ describe('React', () => {
...
+          // does not cause a re-render
+          store.dispatch({ type: '' })
+
+          expect(renderedItems).toEqual([0, 0])
+        })
       })
 
       describe('edge cases', () => {
         it('ignores transient errors in selector (e.g. due to stale props)', () => {
-          // TODO Not sure this test is really testing what we want.
```
MrWolfZ: This test was correct in what it tested. In fact, it failing showed an error in the implementation.

src/hooks/useSelector.js
```diff
@@ -46,13 +52,15 @@ export function useSelector(selector) {
...
+
   const latestSubscriptionCallbackError = useRef()
-  const latestSelector = useRef(selector)
+  const latestSelector = useRef(memoizedSelector)
 
   let selectedState = undefined
 
   try {
-    selectedState = latestSelector.current(store.getState())
+    selectedState = memoizedSelector(store.getState())
```
MrWolfZ: using `latestSelector.current` was wrong, since it was an stale selector. This change fixes that issue, which makes the test for stale props succeed.

Dudeonyx:
If no dependency array is passed what happens?

src/hooks/useSelector.js
markerikson: Hah, nice catch :)

markerikson:
@Dudeonyx : the latest passed-in selector should get used, instead of a previous one.

markerikson:
LGTM.  I'll try to get alpha.1 published shortly.

PR #1259, opened 2019-04-24
Use react-hooks-testing-library to test hooks

mpeyper:
For your consideration, I've updated the hook unit tests to use [`react-hooks-testing-library`](https://github.com/mpeyper/react-hooks-testing-library), which is similar to `react-testing-library` but with a focus on testing reuable hooks.

The only hook tests I left using actual components were the `useSelector/lifeycle interactions` and `useSelector/performance optimizations and bail-outs` tests as they seem to rely on the interplay between the `Parent` and `Child` components to work.  I could probably hack something together to use the `wrapper` to simulate the relationship, but I felt that they are actually better in their current form as it more clearly describes the relationship between the interactions.

There are no hard feelings if you choose not to merge this and continue with the tests you have.

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,6 +74,8 @@
     "prettier": "^1.16.4",
     "react": "^16.8.6",
     "react-dom": "^16.8.6",
+    "react-hooks-testing-library": "^0.5.0",
+    "react-test-renderer": "^16.8.6",
     "react-testing-library": "^5.9.0",
     "redux": "^4.0.1",
     "rimraf": "^2.6.3",
--- a/test/hooks/useActions.spec.js
+++ b/test/hooks/useActions.spec.js
@@ -1,6 +1,8 @@
+/*eslint-disable react/display-name*/
+
 import React from 'react'
 import { createStore } from 'redux'
-import * as rtl from 'react-testing-library'
+import { renderHook } from 'react-hooks-testing-library'
 import { Provider as ProviderMock, useActions } from '../../src/index.js'
 
 describe('React', () => {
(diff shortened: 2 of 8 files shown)


test/hooks/useActions.spec.js
```diff
@@ -1,6 +1,8 @@
+/*eslint-disable react/display-name*/
```
timdorr: Since we have this in each file, can we just add this to a .eslintrc file instead?

package.json
```diff
@@ -74,6 +74,8 @@
     "prettier": "^1.16.4",
     "react": "^16.8.6",
     "react-dom": "^16.8.6",
+    "react-hooks-testing-library": "^0.5.0",
+    "react-test-renderer": "^16.8.6",
```
timdorr: Is this dep needed? I'm not seeing it used.

timdorr:
I like it! It definitely cleans things up a bit for a number of tests cases.

test/hooks/useActions.spec.js
mpeyper: You could, but then you won't get the warning (which is legitimate) for the production code, which is probably not ideal.

package.json
mpeyper: It's a peer dependency of `react-hooks-testing-library`.  We use it to render the test component without requiring a DOM to be in place, but it should be the corresponding version to the provided `react` version (also a peer dependency).

test/hooks/useActions.spec.js
timdorr: I mean a separate file in the test or test/hooks path. That scopes it to just these files.

mpeyper: TIL eslint supports nested configs...

My latest commit removes the per-file disable comment and [adds a `.eslintrc` file](https://github.com/reduxjs/react-redux/pull/1259/files#diff-c5976c613641831504f0ebe88740bbce) to the `test/hooks` directory.

timdorr:
Perfect. Thanks!

PR #1263, opened 2019-04-26
fix timing issue in component updates due to consecutive dispatches

MrWolfZ:
This PR fixes #1262 and should hopefully fix the problems described in #1226 as well.

PR #1264, opened 2019-04-28
Revert "fix timing issue in component updates due to consecutive dispatches"

markerikson:
Reverts reduxjs/react-redux#1263

This should have gone into `v7-hooks-alpha` instead.


markerikson:
Well, I'm a doofus and shouldn't have done that.  Fixed.

PR #1068, opened 2018-10-29
Docs: overview of selector functions

zsid:
__What does this PR do__
* Adds a section about selectors 

__Summary of the Changes__
* Adds a sub-section about selectors in `Using React-Redux` section

__Questions__
* I am trying to find how to run the docs and see how they look on the page but I was unable to find it. What is the best way to do it?

PR #1267, opened 2019-04-30
adjust the hooks docs with more details about useSelector and add some more examples

MrWolfZ:
I have adjusted the hooks docs with some more details regarding `useSelector`. I have also added some more examples.

I know some of these changes may be controversial, especially since it is not always a good idea to talk too much about implementation details in the docs. However, I feel in this case it is required due to the complexity of the stale props issue.

I don't expect all of these changes to be accepted, but I would like to discuss them anyways.

### DIFF
--- a/docs/api/hooks.md
+++ b/docs/api/hooks.md
@@ -37,7 +37,7 @@ From there, you may import any of the listed React Redux hooks APIs and use them
 ## `useSelector()`
 
 ```js
-const result : any = useSelector(selector : Function, deps : any[])
+const result : any = useSelector(selector : Function)
 ```
 
 Allows you to extract data from the Redux store state, using a selector function.
@@ -49,9 +49,9 @@ The selector is approximately equivalent to the [`mapStateToProps` argument to `
 However, there are some differences between the selectors passed to `useSelector()` and a `mapState` function:
 
 - The selector may return any value as a result, not just an object. The return value of the selector will be used as the return value of the `useSelector()` hook.
-- The selector function used will be based on the `deps` array. If no deps array is provided, the latest passed-in selector function will be used when the component renders, and also when any actions are dispatched before the next render. If a deps array is provided, the last saved selector will be used, and that selector will be overwritten whenever the deps array contents have changed. Therefore, the deps are only required if the selector needs to be referentially stable, e.g. if you are using memoizing selectors.
 - When an action is dispatched, `useSelector()` will do a shallow comparison of the previous selector result value and the current result value. If they are different, the component will be forced to re-render. If they are the same, they component will not re-render.
-- The selector function does _not_ receive an `ownProps` argument. However, props can be used through closure (see the examples below) or by using a curried selector. If you are providing `deps` to `useSelector()` make sure they contain all the props you are using.
+- The selector function does _not_ receive an `ownProps` argument. However, props can be used through closure (see the examples below) or by using a curried selector.
+- Extra care must be taken when using memoizing selectors (see examples below for more details).
 
 > **Note**: There are potential edge cases with using props in selectors that may cause errors. See the [Usage Warnings](#usage-warnings) section of this page for further details.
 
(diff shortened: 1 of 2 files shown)


MrWolfZ:
Reading through the docs another time just now I realized that it doesn't highlight one of the primary advantages of hooks: composability. Does it make sense to add an example for `useSelector` and `useActions` to show how they can be used to create custom hooks?

timdorr:
Yeah, I think that would be a good idea. Thanks, Jon!

docs/api/hooks.md
```diff
@@ -67,19 +69,37 @@ export const CounterComponent = () => {
...
+import { createSelector } from 'reselect'
+
+const allOtherTodos = createSelector(
+  state => state.todos,
+  (_, id) => id,
+  (todos, id) => todos.filter(todo => todo.id !== id)
+)
+
+export const OtherTodoListItems = ({ id }) => {
+  const otherTodos = useSelector(s => allOtherTodos(s, id), [id])
```
lukaszfiszer: I don't have an in-depth understating of `useSelector` so I might be wrong, but I think that in this example the same `allOtherTodos` selector will be used in each render irrespectively if `deps` are provided or not.  This is because the selector is created outside of `OtherTodoListItems`, so it's reference never changes. The arrow function wrapping it will/will not change, but it does not really matter.

Additionally, in this example the selector will be shared across all instances of this `OtherTodoListItems`. And because reselect selectors have a cache of 1, this mean it will always recompute (see last paragraph from https://github.com/reduxjs/reselect#accessing-react-props-in-selectors )

MrWolfZ: You are of course right. When writing this example I was thinking whether I should use a factory selector instead but thought that might be too much added complexity. However, it seems that is necessary for the selector to work correctly. I'll adjust the example.

MrWolfZ: I decided to just move the selector into the component after all to keep it simple

lukaszfiszer: I had more thoughts on the subject of `deps`. I have written them down in https://github.com/reduxjs/react-redux/issues/1252#issuecomment-488591342

MrWolfZ:
Depending on the outcome of the `useSelector` `deps` discussion and also the `useActions` discussion I'll make further changes to this PR (including those custom hook examples I mentioned).

timdorr:
Yep, looks like those both need to be updated 😄

MrWolfZ:
So, I finally got around to updating the docs. Please let me know what you think. I am especially interested in feedback regarding the memoizing selector examples. As far as I can see, there are in fact multiple ways how you can use them with component props. The `useMemo` approach I used in the examples felt most natural to me, since it only creates a single instance of the selector per component instance and you don't need to think about using the correct `deps`.

This is what I used in the docs for now:

```jsx
const makeNrOfTodosWithIsDoneSelector = () =>
  createSelector(
    state => state.todos,
    (_, isDone) => isDone,
    (todos, isDone) => todos.filter(todo => todo.isDone === isDone).length
  )

export const TodoCounterForIsDoneValue = ({ isDone }) => {
  const selectNrOfTodosWithIsDone = useMemo(makeNrOfTodosWithIsDoneSelector, [])

  const nrOfTodosWithIsDoneValue = useSelector(state =>
    selectNrOfTodosWithIsDoneValue(state, isDone)
  )

  return <div>{nrOfTodosWithIsDoneValue}</div>
}
```

However, the other ways I see are these:

Creating a new instance of the selector for each new set of props (which makes the memoization part a bit more cumbersome but simplifies the `useSelector` call:

```jsx
export const TodoCounterForIsDoneValue = ({ isDone }) => {
  const selectNrOfTodosWithIsDone = useCallback(
    state => makeNrOfTodosWithIsDoneSelector(state, isDone),
    [isDone]
  )

  const nrOfTodosWithIsDoneValue = useSelector(selectNrOfTodosWithIsDoneValue)
  return <div>{nrOfTodosWithIsDoneValue}</div>
}
```

Using a factory selector:

```jsx
const makeNrOfTodosWithIsDoneSelector = isDone =>
  createSelector(
    state => state.todos,
    todos => todos.filter(todo => todo.isDone === isDone).length
  )

export const TodoCounterForIsDoneValue = ({ isDone }) => {
  const selectNrOfTodosWithIsDone = useCallback(
    makeNrOfTodosWithIsDoneSelector(isDone),
    [isDone]
  )

  const nrOfTodosWithIsDoneValue = useSelector(selectNrOfTodosWithIsDone)
  return <div>{nrOfTodosWithIsDoneValue}</div>
}
```

I am honestly not sure which should be the recommended way shown in the docs.

MrWolfZ:
@josepot I think you had some examples in your PR that removed the `deps`, but they aren't there anymore. I also remember you saying you want to comment on these changes since you disagree with some. Can you please have a look at the examples above and the changes in general and tell me your thoughts? Is there another way I missed?

docs/api/hooks.md
```diff
@@ -69,19 +71,116 @@ export const CounterComponent = () => {
...
+When using `useSelector` with an inline selector as shown above, a new instance of the selector is created whenever the component is rendered. This works as long as the selector does not maintain any state. However, memoizing selectors (e.g. created via `createSelector` from `reselect`) do have internal state, and therefore care must be taken when using them. Below you can find typical usage scenarios for memoizing selectors.
+
+When the selector does only depend on the state, simply ensure that it is declared outside of the component so that the same selector instance is used for each render:
+
+```jsx
+import React from 'react'
+import { useSelector } from 'react-redux'
+import { createSelector } from 'reselect'
+
+const selectNrOfDoneTodos = createSelector(
```
timdorr: I know this is super-pedantic, but instead of "Nr" can you use "Num"? That's a more common shortening of "number".

MrWolfZ: done

docs/api/hooks.md
```diff
@@ -69,19 +71,119 @@ export const CounterComponent = () => {
...
+Using props via closure to determine what to extract:
 
 ```jsx
 import React from 'react'
 import { useSelector } from 'react-redux'
 
 export const TodoListItem = props => {
-  const todo = useSelector(state => state.todos[props.id], [props.id])
-
+  const todo = useSelector(state => state.todos[props.id])
```
josepot: I *think* that perhaps the docs in this case should encourage the usage of `useCallback`?

```js
const todoSelector = useCallback(state => state.todos[props.id], [props.id])
const todo = useSelector(todoSelector)
```

FTR I actually think that this is a good practice regardless of what ends up happening with #1273 :slightly_smiling_face:

josepot:
Thanks @MrWolfZ for pinging me into the discussion.

I have been thinking a lot about this and I think that perhaps it would be best if our docs don't even show an example of how to use `useSelector` with Reselect. I think that we should leave that to the Reselect team.

The reason for this is that I think that when the docs of Reselect were written, they had in mind the API of `connect`, and perhaps we are now making some assumptions about what the best way to leverage Reselect with React-Redux hooks are and I really think that we should leave that to them.

So, that would be what I would my first preference.

If we think that it's still valuable to add that section into the docs, then I would prefer changing the example of the selector creator so that the selector creator receives the parameter that it needs, like this:

```js
import React, { useMemo } from 'react'
import { useSelector } from 'react-redux'
import { createSelector } from 'reselect'

const makeNumOfTodosWithIsDoneSelector = (isDone) =>
  createSelector(
    state => state.todos,
    (todos) => todos.filter(todo => todo.isDone === isDone).length
  )

export const TodoCounterForIsDoneValue = ({ isDone }) => {
  const selectNumOfTodosWithIsDone = useMemo(
    () => makeNumOfTodosWithIsDoneSelector(isDone),
    [isDone]
  )

  const NumOfTodosWithIsDoneValue = useSelector(selectNumOfTodosWithIsDone)

  return <div>{NumOfTodosWithIsDoneValue}</div>
}

export const App = () => {
  return (
    <>
      <span>Number of done todos:</span>
      <TodoCounterForIsDoneValue isDone={true} />
      <span>Number of unfinished todos:</span>
      <TodoCounterForIsDoneValue isDone={false} />
    </>
  )
}
```

docs/api/hooks.md
lukaszfiszer: But why should we encourage usage of `useCallback`? I see no real benefit (the cost of creating new lamba per render is negligible) and it only adds complexity.

lukaszfiszer:
Agreeing with @josepot that  react-redux docs are not a good place for detailed examples of integration with a particular 3rd party library like reselect. It's rather reselect docs that should be updated with those examples as soon as v7.1 is published. 

Currently `connect` docs contain exactly 0 examples with reselect - they only have [few external links at the bottom](https://react-redux.js.org/6.x/using-react-redux/connect-mapstate#links-and-references). On the other hand, [the whole documentation of reselect](https://github.com/reduxjs/reselect) focuses on integration with connect.

docs/api/hooks.md
josepot: Yep @lukeapage you are right. Actually, that would be likely to make perf even worse because the overhead of an extra hook. Yep, that was a brain-fart. Sorry!

lukeapage: @josepot you meant @lukaszfiszer ?

josepot: Yep, I did, indeed mean @lukaszfiszer ... Sorry about that! :sweat_smile:

MrWolfZ:
@timdorr @markerikson what do you think? should we include `reselect` examples in the docs?

markerikson:
@lukaszfiszer , @MrWolfZ , @josepot : the only reason the React-Redux docs _don't_ mention use of Reselect right now is because we just haven't gotten around to it.  I've explicitly wanted to add pages on "performance optimization" (including use of selectors) since the start of the docs revamp - see #1001 .  

#1068 was an attempt to adapt my post [Idiomatic Redux: Using Reselect Selectors for Encapsulation and Performance]() as a docs page.  I just closed that PR, as I wasn't entirely happy with how the content got tweaked.

However, all that said: I _do_ think we should have at least some examples of using Reselect with `useSelector()`, especially now that I just merged the "ref equality" changes.

PR #1272, opened 2019-05-02
Remove deps of useSelector

josepot:
As @lukaszfiszer pointed out on #1252 it's better if we get rid of the dependencies of `useSelector` and we let the users to memoize the selectors themselves.

### DIFF
--- a/test/hooks/useSelector.spec.js
+++ b/test/hooks/useSelector.spec.js
@@ -1,6 +1,6 @@
 /*eslint-disable react/prop-types*/
 
-import React, { useReducer } from 'react'
+import React from 'react'
 import { createStore } from 'redux'
 import { renderHook, act } from 'react-hooks-testing-library'
 import * as rtl from 'react-testing-library'
@@ -153,34 +153,6 @@ describe('React', () => {
           store.dispatch({ type: '' })
 
           expect(renderedItems.length).toBe(1)
-        })
-
-        it('re-uses the selector if deps do not change', () => {
-          let selectorId = 0
-          let forceRender
-
-          const Comp = () => {
-            const [, f] = useReducer(c => c + 1, 0)
-            forceRender = f
-            const renderedSelectorId = selectorId++
-            const value = useSelector(() => renderedSelectorId, [])
-            renderedItems.push(value)
-            return <div />
-          }
-
-          rtl.render(
-            <ProviderMock store={store}>
-              <Comp />
-            </ProviderMock>
-          )
-
-          rtl.act(forceRender)
-
-          // this line verifies the susbcription callback uses the same memoized selector and therefore
-          // does not cause a re-render
-          store.dispatch({ type: '' })
-
-          expect(renderedItems).toEqual([0, 0])
         })
       })
 
--- a/src/hooks/useSelector.js
+++ b/src/hooks/useSelector.js
@@ -24,8 +24,6 @@ const useIsomorphicLayoutEffect =
  * useful if you provide a selector that memoizes values).
  *
  * @param {Function} selector the selector function
- * @param {any[]} deps (optional) dependencies array to control referential stability
- * of the selector
  *
  * @returns {any} the selected state
  *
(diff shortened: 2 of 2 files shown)


test/hooks/useSelector.spec.js
```diff
@@ -155,15 +155,16 @@ describe('React', () => {
           expect(renderedItems.length).toBe(1)
         })
 
-        it('re-uses the selector if deps do not change', () => {
+        it('re-uses the selector if it does not change', () => {
```
josepot: Actually, perhaps we should get rid of this test now...

MrWolfZ: yeah, this test was specifically for `deps`, so I would remove it

docs/api/hooks.md
```diff
@@ -68,16 +67,71 @@ export const CounterComponent = () => {
...
 ```jsx
-import React from 'react'
+import React, {useCallback} from 'react'
 import { useSelector } from 'react-redux'
 
-export const TodoListItem = props => (
-  const todo = useSelector(state => state.todos[props.id], [props.id])
+export const TodoListItem = ({id}) => (
+  const selector = useCallback(state => state.todos[id], [id])
+  const todo = useSelector(selector)
```
MrWolfZ: for this example it makes no sense to `useCallback` and it makes the example more complicated than it needs to be. Just do `useSelector(state => state.todos[id])` and leave the whole memoization stuff for other examples

docs/api/hooks.md
```diff
@@ -68,16 +67,71 @@ export const CounterComponent = () => {
...
+```jsx
+// ducks/todos.js
+import { createSelector } from 'reselect'
+
+export const selectPendingTodos = createSelector(
+  (state) => state.todos,
+  (todos) => todos.filter(todo => todo.completed !== true)
+)
+
+export const selectNPendingTodos = createSelector(
```
MrWolfZ: these docs are not about showing how to compose reselect selectors. just use the simples possible example, i.e. get the length directly in the first selector

MrWolfZ: I am a big fan of proper naming, so I would make sure the examples do it to, i.e. call this `selectNrOfPendingTodos`

docs/api/hooks.md
```diff
@@ -68,16 +67,71 @@ export const CounterComponent = () => {
...
-  const todo = useSelector(state => state.todos[props.id], [props.id])
+export const TodoListItem = ({id}) => (
+  const selector = useCallback(state => state.todos[id], [id])
+  const todo = useSelector(selector)
 
   return <div>{todo.text}</div>
 }
 ```
 
+With Reselect:
```
MrWolfZ: I would give much more context about this and the following pattern, i.e. in which situations to use it

lukaszfiszer:
This is indeed my proposed solution, but I still would like to hear what other maintainers think about it. Let's not rush too much with this change :-)

MrWolfZ:
Regarding the docs changes, since I have another [PR](https://github.com/reduxjs/react-redux/pull/1267) open with docs changes, I would like to add those docs changes there, if that's okay with you.

josepot:
> Regarding the docs changes, since I have another PR open with docs changes, I would like to add those docs changes there, if that's okay with you.

Totally! I went to open that PR and then I noticed that there was another PR for docs. That's why I didn't open it. Although, I will probably add some comments into that PR because I don't fully agree with some of the comments that you made to my docs proposal. No worries, nothing important. Thanks @MrWolfZ!

lukaszfiszer:
What's your view on this PR @markerikson @timdorr ?

timdorr:
@lukaszfiszer I'm watching the discussion in #1273, because that encompasses this change plus some additional optimizations with caveats.

josepot:
> @lukaszfiszer I'm watching the discussion in #1273, because that encompasses this change plus some additional optimizations with caveats.

@timdorr ups! That is the opposite of what I intended, sorry! I initially thought that this PR was going to get merged quickly, but I wasn't sure about the other one, that's why the other one also has this commit. Sorry! 

If you want I can close #1273 and re-open it once this PR gets merged or declined... (Or just leave it closed if you think that's better). Please, let me know if that would make things easier. Thanks!

timdorr:
I am a few days behind on the main feedback issue, so I need to catch up there too. I'll be able to look at this after I've read that novels worth of posts :)

gaearon:
FWIW our conclusion internally has been that custom Hooks should avoid dependency argument, and people should just memoize if it's necessary. You have to learn that pattern anyway.

timdorr:
I say we get this merged in now and rebase #1273 to get that _just_ about the evaluations change.

timdorr:
No big objections? Cool, in it goes...

PR #1273, opened 2019-05-02
Avoid unnecessary selector evaluations

josepot:
Hi,

While I was working on #1272 I noticed that the selector was being evaluated more times than what's necessary. This PR makes sure that the selector is only evaluated when it's strictly needed.

cc: @MrWolfZ I would really appreciate you having a look at this, just to make sure that I didn't miss anything. All the previous tests are passing, and I have also added a new one to make sure that updating the selector works correctly. Please let me know what you think. Thanks!

## Edit

A couple of things that I forgot to mention that perhaps aren't that obvious:

1. The only commit that's relevant for this PR is [this one](https://github.com/reduxjs/react-redux/pull/1273/commits/88dce648b9c3e7eeffbf0902345c4e43b49a54b6). The other one is the commit of #1272... I'm just being a bit optimistic and hoping that #1272 will get merged before this one.

2. My rationale for this change is the following:

- If the update comes as a result of the `forceRender` of `checkForUpdates`, then there is no need to re-evaluate the selector.

- If the update comes from "the outside", then it could be that:

1. The selector has been updated, in which case we need to re-evaluate the selector
2. It's a parent update that has nothing to do with us, so we don't need to re-evaluate the selector... What's likely, though, is that if the parent update was a result of the state changing, then `checkForUpdates` will already pick that change and trigger a render... Chances are that the child-update will have happened before the parent one, but it doesn't really matter, because it will happen during the same loop, so it will be batched... :slightly_smiling_face:

### DIFF
--- a/src/hooks/useSelector.js
+++ b/src/hooks/useSelector.js
@@ -50,12 +50,17 @@ export function useSelector(selector) {
   ])
 
   const latestSubscriptionCallbackError = useRef()
-  const latestSelector = useRef(selector)
+  const latestSelector = useRef()
+  const latestSelectedState = useRef()
 
-  let selectedState = undefined
+  let selectedState
 
   try {
-    selectedState = selector(store.getState())
+    selectedState =
+      selector !== latestSelector.current ||
+      latestSubscriptionCallbackError.current
+        ? selector(store.getState())
+        : latestSelectedState.current
   } catch (err) {
     let errorMessage = `An error occured while selecting the store state: ${
       err.message
@@ -69,8 +74,6 @@ export function useSelector(selector) {
 
     throw new Error(errorMessage)
   }
-
-  const latestSelectedState = useRef(selectedState)
 
   useIsomorphicLayoutEffect(() => {
     latestSelector.current = selector
--- a/src/hooks/useSelector.js
+++ b/src/hooks/useSelector.js
@@ -56,11 +56,14 @@ export function useSelector(selector) {
   let selectedState
 
   try {
-    selectedState =
+    if (
       selector !== latestSelector.current ||
       latestSubscriptionCallbackError.current
-        ? selector(store.getState())
-        : latestSelectedState.current
+    ) {
+      selectedState = selector(store.getState())
+    } else {
+      selectedState = latestSelectedState.current
+    }
   } catch (err) {
     let errorMessage = `An error occured while selecting the store state: ${
       err.message
--- a/test/hooks/useSelector.spec.js
+++ b/test/hooks/useSelector.spec.js
@@ -1,6 +1,6 @@
 /*eslint-disable react/prop-types*/
 
-import React from 'react'
+import React, { useCallback, useReducer } from 'react'
 import { createStore } from 'redux'
 import { renderHook, act } from 'react-hooks-testing-library'
 import * as rtl from 'react-testing-library'
@@ -47,6 +47,29 @@ describe('React', () => {
       })
 
       describe('lifeycle interactions', () => {
+        it('always uses the latest state', () => {
+          store = createStore(c => c + 1, -1)
+
+          const Comp = () => {
+            const selector = useCallback(c => c + 1, [])
+            const value = useSelector(selector)
+            renderedItems.push(value)
+            return <div />
+          }
+
+          rtl.render(
+            <ProviderMock store={store}>
+              <Comp />
+            </ProviderMock>
+          )
+
+          expect(renderedItems).toEqual([1])
+
+          store.dispatch({ type: '' })
+
+          expect(renderedItems).toEqual([1, 2])
+        })
+
         it('subscribes to the store synchronously', () => {
           let rootSubscription
 
@@ -154,6 +177,39 @@ describe('React', () => {
 
           expect(renderedItems.length).toBe(1)
         })
+      })
+
+      it('uses the latest selector', () => {
+        let selectorId = 0
+        let forceRender
+
+        const Comp = () => {
+          const [, f] = useReducer(c => c + 1, 0)
+          forceRender = f
+          const renderedSelectorId = selectorId++
+          const value = useSelector(() => renderedSelectorId)
+          renderedItems.push(value)
+          return <div />
+        }
+
+        rtl.render(
+          <ProviderMock store={store}>
+            <Comp />
+          </ProviderMock>
+        )
+
+        expect(renderedItems).toEqual([0])
+
+        rtl.act(forceRender)
+        expect(renderedItems).toEqual([0, 1])
+
+        rtl.act(() => {
+          store.dispatch({ type: '' })
+        })
+        expect(renderedItems).toEqual([0, 1])
+
+        rtl.act(forceRender)
+        expect(renderedItems).toEqual([0, 1, 2])
       })
 
       describe('edge cases', () => {
--- a/src/alternate-renderers.js
+++ b/src/alternate-renderers.js
@@ -8,6 +8,7 @@ import { useSelector } from './hooks/useSelector'
 import { useStore } from './hooks/useStore'
 
 import { getBatch } from './utils/batch'
+import shallowEqual from './utils/shallowEqual'
 
 // For other renderers besides ReactDOM and React Native, use the default noop batch function
 const batch = getBatch()
@@ -20,5 +21,6 @@ export {
   batch,
   useDispatch,
   useSelector,
-  useStore
+  useStore,
+  shallowEqual
 }


src/hooks/useSelector.js
```diff
@@ -51,15 +49,15 @@ export function useSelector(selector, deps) {
...
   const latestSubscriptionCallbackError = useRef()
-  const latestSelector = useRef(memoizedSelector)
+  const latestSelector = useRef(selector)
 
-  let selectedState = undefined
+  let latestSelectedState = useRef()
 
   try {
-    selectedState = memoizedSelector(store.getState())
+    latestSelectedState.current = useMemo(() => selector(store.getState()), [
```
MrWolfZ: This is completely wrong, since if the selector does not change, it will always have the result based of the state at the point the selector changed.

josepot: Hi @MrWolfZ I have added an Edit that explains the rationale of this PR a bit better. If you really think that this is wrong, could you please provide a test that proves that?

MrWolfZ: Sure, here you go:

```jsx
it('always uses the latest state', () => {
  store = createStore(c => c + 1, -1)

  const Comp = () => {
    const selector = useCallback(c => c + 1, [])
    const value = useSelector(selector)
    renderedItems.push(value)
    return <div />
  }

  rtl.render(
    <ProviderMock store={store}>
      <Comp />
    </ProviderMock>
  )

  expect(renderedItems).toEqual([1])

  store.dispatch({ type: '' })

  expect(renderedItems).toBe([1, 2])
})
```

MrWolfZ: In general, I really wouldn't worry too much about how often the selector is called. Even if this change worked, it wouldn't be worth the added complexity

josepot: Ups! You were right! I just updated it the code, I think that I made it a lot simpler (a lot less refs) and more performant... Wit this latest code I think that I haven't added complexity, on the contrary and on top of that we don't have unnecessary evaluations of selector :smile:

test/hooks/useSelector.spec.js
```diff
@@ -47,6 +47,31 @@ describe('React', () => {
       })
 
       describe('lifeycle interactions', () => {
+        it.only('always uses the latest state', () => {
```
lukaszfiszer: `it.only` - only this particular test is executed

josepot: OMG! Thanks and sorry!

josepot:
I will close this for now, sorry :sweat_smile:

josepot:
I'm reopening this PR.

I _think_ that this should actually make a difference in terms of performance, because the selector will only be evaluated when it's strictly necessary. Sure, it does add some complexity, but the tests are all passing (now they are 🙈😅).

I guess that it would only make sense to merge this PR if the performance benchmarks are actually better, otherwise it wouldn't make sense to merge it, of course.

Please @MrWolfZ , could you please have another look at this 🙏? Thanks a lot!

src/hooks/useSelector.js
```diff
@@ -100,7 +78,10 @@ export function useSelector(selector, deps) {
         latestSubscriptionCallbackError.current = err
       }
 
-      forceRender({})
+      setState(prev => ({
```
MrWolfZ: have you actually run this in an application instead of just in a test? doesn't `setState` inside an effect that executes on each render cause an infinite render loop?

src/hooks/useSelector.js
```diff
@@ -111,5 +92,32 @@ export function useSelector(selector, deps) {
     return () => subscription.tryUnsubscribe()
   }, [store, subscription])
 
-  return selectedState
+  try {
+    latestSelectedState.current = useMemo(() => selector(store.getState()), [
```
MrWolfZ: I think instead of `useMemo` and the whole `id` stuff you could just capture the store state into a `latestState` ref and compare `store.getState()` to that value during render and if so, return `latestSelectedState.current`

src/hooks/useSelector.js
josepot: > have you actually run this in an application instead of just in a test

Yep. Now I have.

> doesn't setState inside an effect that executes on each render cause an infinite render loop

This is not inside the render. If this caused an infinite loop, the forceRender would also cause an infinite loop.

src/hooks/useSelector.js
MrWolfZ: Oh, and assigning a ref during render is actually a side-effect, and the render function should be pure.

josepot:
> I still think this change doesn't really add anything and makes the code just more complex. 

I honestly don't think that it makes the code a lot more complex than what it already was, but that is subjective, of course.

> If the user cares about the selector logic being called too often, let them use a memoizing selector.

But this is the thing, even if the user memoized the selector, the selector would be executed twice when an update is detected and when the render that happens after that takes place... Memoized selectors are cheap, but they are not free, in the case of reselect the whole tree of selectors has be be re-evaluated with their latest arguments, if we can avoid that I think that we should.

Also, without this possible improvement, any time that there is an update from an ancestor, if that update is not changing the selector, the selector would be evaluated again... Which is totally unnecessary. Honestly, I don't know why you are so against this.

> Also, since this change is supposed to bring better performance, please run this code against the benchmark repository.

I would love to do that. Could you please show me where that repo is?

josepot:
Also @MrWolfZ, please look at the bright side. Even if this change is actually bad or unnecessary, at the very least it will help us to come up with more and better tests for possible future refactors. So far we already got one new test 😉

src/hooks/useSelector.js
josepot: Sure...

I've moved the assignment to the `latestSelectedState.current` into the `finally` clause of the hook.

Please notice that's exactly the same as executing synchronously after the return of the function, which is what `useLayoutEffect` does (in our case `useIsomorphicLayoutEffect`).

I.e: If that makes this hook impure, it also does assigning to references synchronously once the render function has finished... It's the same.

josepot: > Oh, and assigning a ref during render is actually a side-effect, and the render function should be pure.

Ok, you are right. That was not nice. I've changed that, I've also got rid of the superfluous `useMemo`. Thanks!

josepot: > I think instead of useMemo and the whole id stuff you could just capture the store state into a latestState ref and compare store.getState() to that value during render (and of course compare the selector) and if so, return latestSelectedState.current

Actually, I came up with a cleaner and more performant way to do that, you should check it out! I really think that my latest code makes the implementation even more understandable than it was before. :blush:

josepot:
Hi all,

I have been benchmarking this and I think that the results are very promising.

Let me explain what I have tried and what I have found out:

I'm comparing 2 different builds, the build produced by #1272 and the build produced by the code of this PR. The reason for just comparing this 2 is that they both have the same API: no `useActions` and deps on `useSelector`, so I can test the benchmarks against the exact same code.

Next I have updated the code of react-redux-benchmarks to work with those APIs, I'm not using `connect`, I'm just using `useSelector` and `useDispatch`, also, I intentionally don't wrap the components with `React.memo`.

Another thing that I noticed is that some benchmarks gave very different results when I repeated the runs. For instance, tree-view was being very inconsistent between the different runs. After some investigation I concluded that the reason was [this fragment of code](https://github.com/reduxjs/react-redux-benchmarks/blob/master/sources/tree-view/src/actions/index.js#L35-L37). So, in my runs I have changed that [to this](https://github.com/josepot/react-redux-benchmarks/blob/feat/hooks/sources/tree-view/src/actions/index.js#L37-L40). After that change, I'm getting very consistent results between the different runs.

Now that you know how I tested these benchmarks, let me explain the results that I've got.

The results of the benchmarks are pretty much identical between the 2 different versions (the one that avoid unnecessary selector evaluations and the one that it doesn't). For instance, these are the results of tree-view:

```
┌─────────────────────────────────────┬─────────┬──────────────┬───────────┬───────────┬──────────┬────────────────────────────────────────────────────────────────┐
│ Version                             │ Avg FPS │ Render       │ Scripting │ Rendering │ Painting │ FPS Values                                                     │
│                                     │         │ (Mount, Avg) │           │           │          │                                                                │
├─────────────────────────────────────┼─────────┼──────────────┼───────────┼───────────┼──────────┼────────────────────────────────────────────────────────────────┤
│ 7.0.0.alpha-avoidSelectorEvaluation │ 59.16   │ 608.5, 3.0   │ 635.25    │ 276.85    │ 68.76    │ 50,60,59,60,59,60,59,60,59,60,59,60,59,60,59,58,59,60,59,60,60 │
├─────────────────────────────────────┼─────────┼──────────────┼───────────┼───────────┼──────────┼────────────────────────────────────────────────────────────────┤
│ 7.0.0.alpha-useSelectorNoDeps       │ 59.14   │ 572.6, 3.1   │ 615.77    │ 265.51    │ 63.36    │ 48,60,59,60,59,60,59,60,59,60,59,60,58,60,59,60,59,59          │
└─────────────────────────────────────┴─────────┴──────────────┴───────────┴───────────┴──────────┴────────────────────────────────────────────────────────────────┘
```
However, I did notice that all the selectors of the benchmarks are trivial (they just select a part of the state without computing any derivative data). So, I decided to to try adding a bit of pointless computations to some selectors and that did make an difference. For instance, I changed the selector of the tree-viiew benchmark from this:

```js
const nodeSelector = useCallback(state => state[id], [id]);
```

to this:

```js
const nodeSelector = useCallback(
  state => {
    const sillyLength = Object.keys(state).length;
    return state[id] || sillyLength;
  },
  [id]
);
```

With that little change, I got the following results:

```
┌─────────────────────────────────────┬─────────┬──────────────┬───────────┬───────────┬──────────┬─────────────────────────────────────────────┐
│ Version                             │ Avg FPS │ Render       │ Scripting │ Rendering │ Painting │ FPS Values                                  │
│                                     │         │ (Mount, Avg) │           │           │          │                                             │
├─────────────────────────────────────┼─────────┼──────────────┼───────────┼───────────┼──────────┼─────────────────────────────────────────────┤
│ 7.0.0.alpha-avoidSelectorEvaluation │ 29.36   │ 930.0, 2.8   │ 15610.97  │ 582.25    │ 45.91    │ 3,5,4,5,6,3,4,6,38,60,59,60,60              │
├─────────────────────────────────────┼─────────┼──────────────┼───────────┼───────────┼──────────┼─────────────────────────────────────────────┤
│ 7.0.0.alpha-useSelectorNoDeps       │ 23.87   │ 914.4, 7.7   │ 16657.18  │ 633.54    │ 44.70    │ 5,4,2,3,2,3,2,3,4,5,37,60,59,60,59,60,59,59 │
└─────────────────────────────────────┴─────────┴──────────────┴───────────┴───────────┴──────────┴─────────────────────────────────────────────┘
```

I will share more details about the results that I'm getting with these benchmarks. But so far things are looking good, basically because when the selectors are trivial the performance is the same, but when the selectors are not trivial the performance is significantly better with the code of this PR.

I will keep you posted.

Also, my apologies for the crappy code that I shared in the initial version of this PR. I honestly think that the current code of this PR is "clearer" than the current code of `useSelector`... but I guess that I'm a bit biased :sweat_smile: .

MrWolfZ:
The latest version of this code looks much better, and I think it is now correct. However, I am still doubtful whether it's worth it. Here's why:

- it breaks the "do the simplest thing that can work" principle
- it does objectively increase the complexity of the code (and how could it not if it does more than the original code?); you can see by running both versions through [jshint](https://jshint.com/) that the number of statements increases and that the cyclomatic complexity doubles. There are also two examples from the code I can give that correspond to this.
    1. the current version has two `if` statements, while the proposed code has five
    2. the current code can be read from top to bottom in a linear fashion, while in the proposed version for example the variable `result` is declared in line 57, potentially mutated multiple times in the `try` block starting at line 96, but accessed via closure in line `60` again, so it needs some mental jumps to follow the code

    The reason I go on so much about complexity is that I want to prevent this code from becoming too much like `connectAdvanced`. Have you looked at it recently? Who even understands that anymore except a handful of people?
- the proposed solution involves caching, and as we all know, [there are only two hard things in computer science](https://martinfowler.com/bliki/TwoHardThings.html). While the code looks correct to me, are we really sure we have thought about all the possible edge cases?
- it seems like a premature optimization that can easily be added later if we get reports about performance issues
- while the synthetic benchmark shows there are situations in which it definitely performs better, in reality you would probably never use that pattern, and if you would, you would use a memoizing selector (@josepot mentioned there is still an associated cost with evaluating those, which is of course true, but I would like to see at which point it actually becomes an issue; I think you'll need selectors nested incredibly deeply for this to ever become noticeable)

But again, the code seems correct, and as demonstrated it can provide performance improvements. I can absolutely see both sides of the argument. Since I am not the maintainer of this repo, I can't decide this in any case. It is up to the maintainers to decide whether they prefer simpler more maintainable code or more performant code.

josepot:
@MrWolfZ 

Let me see if I understood you correctly:

You think that this code is correct. You also can see that it can't hurt performance, while it can certainly improve it in some cases (deriving data is something that should be done in selectors). However, you seem to be certain that **any** implementation that tries to avoid unnecessary evaluations of the selector will add so much complexity that it just won't be worth it... Yep?

Well, as you can imagine: I strongly disagree with you :smile:. I think that it is possible to have an implementation that actually makes the current implementation even less complex and that at the same time it avoids unnecessary evaluations of the current selector. I just did another update of the code, maybe you want to check that one out. Although, at this point, I'm pretty sure that you won't like it :man_shrugging:

Also, let me answer some of your comments:

> it breaks the "do the simplest thing that can work" principle

Principles are not dogmas. They are principles. Also, this principle applies better to "App development" than it does to "library development". Sometimes, a library needs to be complex, so that the code in the App can be simpler. If I was working on an App, I wouldn't care about a selector being evaluated twice inside a component unless that proved to be a performance issue... But that is because I would know how the component is being used. When writing code for a library, since we don't know how it's going to end up being used, we should try to make the code as performant as possible. Even if that adds a bit of complexity. That should not be an excuse for creating spaghetti code, of course. But most non-trivial libraries are complex inside, they have to be. This library is living proof of that.

> the current version has two if statements, while the proposed code has five

Actually, my latest code has only 2 if statements. I did add 3 more in the previous implementation to make things more explicit. The 3 extra if statements were all together creating guards, and avoiding nested if statements as much as possible, without a single "else" statement... But since you think that those were creating *so* much complexity, I got rid of them. No problem. 

> the current code can be read from top to bottom in a linear fashion, while in the proposed version for example the variable result is declared in line 57, potentially mutated multiple times in the try block starting at line 96, but accessed via closure in line 60 again, so it needs some mental jumps to follow the code

Actually, I think that the code of this PR reads better from top to bottom than the current one. I will elaborate later on this. However, please notice that with my latest implementation the variable `result` is only assigned **once**. In the `return` statement that's at the bottom, that's it.

> but accessed via closure in line 60 again, so it needs some mental jumps to follow the code

Seriously? The variable is named `result`. It's only assigned when returning a result and it's only read in the same effect that keeps track of all the other previous references, by a ref named `previousResult`... How is that more complex than assigning the selector to the `latestSelector` ref? :smile: Or what the current code does with `selectedState`? Which is [defined in line 59](https://github.com/reduxjs/react-redux/blob/92baa0474914c52bdae5a146dad4deee3301ed8b/src/hooks/useSelector.js#L59), then consumed [in line 81](https://github.com/reduxjs/react-redux/blob/92baa0474914c52bdae5a146dad4deee3301ed8b/src/hooks/useSelector.js#L81) and returned (if nothing in the middle has thrown) by line 114... Also, with the current code `latestSelectedState.current` is being assigned more than once, making things more complex than it should be... I will elaborate on this later.

> The reason I go on so much about complexity is that I want to prevent this code from becoming too much like connectAdvanced. Have you looked at it recently? Who even understands that anymore except a handful of people?

:open_mouth: I mean, please, pretty please with sugar on top, open [this file](https://github.com/reduxjs/react-redux/blob/d129ef16e60f0dd7fa3a050df93400f59daa823a/src/hooks/useSelector.js) (without any diffs or anything) and have a look at it with a fresh set of eyes, and then open [this other one](https://github.com/reduxjs/react-redux/blob/92baa0474914c52bdae5a146dad4deee3301ed8b/src/hooks/useSelector.js) and put it beside the first one, also try looking at the second one with a fresh set of eyes. And after you do that, please tell me: do you really think that the first one makes things more complex?

> the current code can be read from top to bottom

I disagree, I think that the code of this PR is much better for reading the code from top to bottom, and I think that some of the changes introduced by this PR make the code less complex. Let me explain:

With the code of this PR all the hooks are at the top and at the bottom we have the return statement, which is inside a `try`/`catch` block. Meaning that the logic for returning/throwing is in one same place, at the bottom. Unlike with the current code, where the logic is scattered around.

Also, with the code of this PR, the references are all being assigned in the same spot. No one competes for the value of `latestSelectedState.current`. Actually, the code of [this line of the current code](https://github.com/reduxjs/react-redux/blob/92baa0474914c52bdae5a146dad4deee3301ed8b/src/hooks/useSelector.js#L94) it adds a lot of unnecessary complexity because someone could think that the value of that ref will be updated in the next update, but it won't be. That is a lot of complexity that we are getting rid of with the code of this PR.

Also, the code of this PR better describes and explains the different kind of updates that this hook will have.

> the proposed solution involves caching, and as we all know, there are only two hard things in computer science.

Not true at all. What it's hard is cache invalidation, which is not needed in our case. We just keep track of the latest result, [just like the current implementation does with a name that's more confusing](https://github.com/reduxjs/react-redux/blob/92baa0474914c52bdae5a146dad4deee3301ed8b/src/hooks/useSelector.js#L81).

> While the code looks correct to me, are we really sure we have thought about all the possible edge cases?

The same could be said for any piece of code, like the current one. Also, it can help at coming up with more and better tests. So far, we already have 2 new tests.

> while the synthetic benchmark shows there are situations in which it definitely performs better, in reality you would probably never use that pattern

That is just not true. Computing derived data is a very common thing to be done inside selectors.

josepot:
Hi @MrWolfZ!

Let me start by saying that I think that I truly regret my latest response. I think that I should have handled that a bit better. I don't think that I was disrespectful or anything like that, but I do think that I should have tried harder at not making this PR a battle about "who is right" and "who is wrong". At the end of the day, we both want to help at improving this library.

To be completely honest, when I opened this PR and I asked you to have a look at it. I was hoping to collaborate with you in finding an elegant way to avoid unnecessary evaluations of the selector. It made me feel really bad the way that you took my suggestion. I know that was not your intention, but that's how it made me feel. I was honestly looking for collaboration and I think that I ended up making a silly competition out of this. I want to apologize for that. I will edit my previous comment (one last time :slightly_smiling_face:). 

All that being said, could you please have a look at the latest code of this PR? I think that it addresses all the concerns that you pointed out before and the perf benchmarks are looking good so far.

Basically, I found a way to do what I wanted to do without completely refactoring your original code. I did move the try/catch block towards the end, because I think that reads better... But I'm zero percent attach to that, so I can bring it back to where it was before if you don't like that.

If you still think that this change is not worth it because it adds too much complexity. No worries. Let me know it, of course. That won't make me feel bad anymore and I'm not going to over-react again. I will understand that you are saying that because that's your honest opinion.

Thanks!

MrWolfZ:
Indeed, there is no right or wrong here, just two different viewpoints. The code is not "bad" in any way and as far as I can tell, it does what it is supposed to do.

I still believe that in the situations where this optimization comes into play, it is very likely that the user is already using memoizing selectors, at which point there is not much to be gained from this.

I thought a bit more about what it is with this code that rubs me the wrong way, and this is what I came up with:

- by removing the `deps` (assuming that change gets accepted), `useSelector` does not care about the selector identity anymore, which I really liked. With this change that is no longer true (however, contrary to `deps` this change doesn't bleed into the public API, which is good)
- it mixes concerns: react-redux's main concern is to let a component know when it needs to update itself due to a store change; if the user cares about selector performance there are other options out there that take care of this

As I said, it is a trade-off. If the maintainers feel the situations for which this code is written are common enough, they will happily accept this PR. If they think otherwise, they won't. In any case, it is not up to me.

PS: I came up with a very silly and unlikely scenario where this change could cause issues: if a user accidentally closes over some mutable state when memoizing the selector (i.e. making it impure), then the selector reference won't change but the result might (as you know, this is what `reselect` et al do, just that they ensure the selector still seems pure from the outside). In that case, with this change we will return a stale value. Again, this scenario is probably incredibly rare, and even if it happened it would be the user using the API wrong.

josepot:
Since this PR has gone through many different code modifications. I would like to highlight the latest diff, [which is this one](https://github.com/reduxjs/react-redux/pull/1273/commits/a9d95ba894a38d8a30f12cfb75e16cb35a67315e#diff-0a2d9ab9c3fe4ea90b28088602b294ba). Meaning that -leaving tests aside- the only change that this PR would introduce is this changing the assignment of `selectedState` from this:

```js
selectedState = selector(store.getState())
```

to this:

```js
   selectedState =
        ? selector(store.getState())
        : latestSelectedState.current
```

And moving the declaration of the `latestSelectedState` ref to the top (where the other refs are defined) without passing an initial argument to it. Nothing else.

IMO this change does no longer make the code *a lot* more complex:

- All the refs are now declared in the same place.
- The `selectedState` assignation is IMO pretty straight forward: if the selector has changed or if an error happened in the subscription, then lets evaluate the selector, otherwise lets return the latest selected state.

To me, what's important is:

- Can this change hurt performance for any one?

I don't think so.

- Can this change make performance better for those users that compute derived data in their selectors (even for the ones that use Reselect)?

I think so, and the perf benchmarks that I have been running so far also show that (I will share them tonight, I'm at work ATM).

- Will the hypothetical performance gain be worth it all the headaches that this PR is giving to the maintainers?

Most certainly not :sweat_smile:.

Finally, I would like to give my opinion on the latest comments of @MrWolfZ:

> by removing the deps (assuming that change gets accepted), useSelector does not care about the selector identity anymore, which I really liked. With this change that is no longer true (however, contrary to deps this change doesn't bleed into the public API, which is good)

I see it in a different way. We are removing the deps because we want to promote "pure" selectors. I still think that even without this change, the best way to create a selector that depends on some props is to use `useCallback`... But either way, even if the user doesn't do it like that this change won't make things worse for those users, either... that's why I find a bit contradictory that you say that this change doesn't bleed into the public API, at the same time that you say that this change makes the identity of selector relevant for the user... Because to me: it doesn't. It's an implementational detail, it's not documented, it's only for the cases that we can make things better. Users should avoid creating new callbacks in the render functions that will be updated frequently, regardless of this change... But even if they don't, things will work just the same with or without this change.

> it mixes concerns: react-redux's main concern is to let a component know when it needs to update itself due to a store change; if the user cares about selector performance there are other options out there that take care of this

As a library user, I want to think that the library will try to take advantage of any possible perf improvements, so that I don't have to. That's why [I also think that the main reason for using Reselect shouldn't be memoization](https://github.com/reduxjs/reselect/issues/401#issuecomment-485204450). I guess that we see things a bit differently here.

I won't answer the comment of your PS because in the case that you are describing in there, things won't work properly with or without this change :smile:.

timdorr:
Kind of an obvious question: Why would the selector ever change? I'm pretty Dan's hands will emerge from your computer monitor and slap you in the face if you ever did this:
```js
useEffect(user_id ? loadUserEffect : loadLoginEffect)
```
While not enforced by a eslint rule, that kind of thing is basically the same as putting your Hooks inside a conditional. It's objectively bad code and should feel bad.

I think that kind of makes this change a moot point, no?

josepot:
@timdorr 

> Kind of an obvious question: Why would the selector ever change?

For isntance:

```js
function User({ id }) {
  const getUser = useCallback(state => state.users[id], [id]);
  const user = useSelector(getUser);
  // return some elements to show the user-data here
}
```

It's quite likely that the id will also be the key of the Component, in which case the id would never change, of course... But check this out:

```js
function UserDiff({ idA, idB }) {
  const getUserA = useCallback(state => state.users[idA], [idA]);
  const getUserB = useCallback(state => state.users[idB], [idB]);

  const userA = useSelector(getUserA);
  const userB = useSelector(getUserB);

  // return some elements that show the differences or whatever
}
```

In the latest case, the IDs could change and the key of the component wouldn't be based on those IDs.

josepot:
Hi @MrWolfZ @timdorr and @markerikson,

Sorry that it took me a while to post the results. I got started a new job and I have been quite busy for the last 2 weeks.

These results are obtained based on the [code of this branch](https://github.com/josepot/react-redux-benchmarks/tree/feat/hooks). You may want to have a look at it just to make sure that I didn't screw anything. If I'm reading these results correctly, they are showing an improvement on performance and that is despite the fact that the selectors of the benchmarks don't actually perform any computations. These are the results:

```bash
Running benchmark deeptree                                                                                                                                                            
  react-redux version: 7.0.0.alpha-avoidSelectorEvaluation                                           
    Checking max FPS... (30 seconds)              
    Running trace...    (30 seconds)                                                                 
  react-redux version: 7.0.0.alpha-useSelectorNoDeps                                                 
    Checking max FPS... (30 seconds)                                                                 
    Running trace...    (30 seconds)                                                                 
                                                                                                     
┌─────────────────────────────────────┬─────────┬──────────────┬───────────┬───────────┬──────────┬─────────────────────────────────────────────────────────────────────────┐
│ Version                             │ Avg FPS │ Render       │ Scripting │ Rendering │ Painting │ FPS Values                                                              │
│                                     │         │ (Mount, Avg) │           │           │          │                                                                         │
├─────────────────────────────────────┼─────────┼──────────────┼───────────┼───────────┼──────────┼─────────────────────────────────────────────────────────────────────────┤
│ 7.0.0.alpha-avoidSelectorEvaluation │ 44.06   │ 86.9, 0.6    │ 9452.44   │ 10396.98  │ 6063.76  │ 56,57,58,55,57,58,57,58,41,35,37,34,35,34,33,24,27,24,24                │
├─────────────────────────────────────┼─────────┼──────────────┼───────────┼───────────┼──────────┼─────────────────────────────────────────────────────────────────────────┤
│ 7.0.0.alpha-useSelectorNoDeps       │ 42.27   │ 88.8, 0.6    │ 9521.25   │ 10229.33  │ 6096.60  │ 56,55,56,55,56,53,54,53,56,52,53,37,33,37,33,34,33,32,24,21,23,22,25,25 │
└─────────────────────────────────────┴─────────┴──────────────┴───────────┴───────────┴──────────┴─────────────────────────────────────────────────────────────────────────┘
Running benchmark deeptree-nested                                                                    
  react-redux version: 7.0.0.alpha-avoidSelectorEvaluation                                     
    Checking max FPS... (30 seconds)                                                                 
    Running trace...    (30 seconds)             
  react-redux version: 7.0.0.alpha-useSelectorNoDeps
    Checking max FPS... (30 seconds)         
    Running trace...    (30 seconds)         
                                                  
┌─────────────────────────────────────┬─────────┬──────────────┬───────────┬───────────┬──────────┬────────────────────────────────────────────────────────────────────────────┐
│ Version                             │ Avg FPS │ Render       │ Scripting │ Rendering │ Painting │ FPS Values                                                                 │
│                                     │         │ (Mount, Avg) │           │           │          │                                                                            │
├─────────────────────────────────────┼─────────┼──────────────┼───────────┼───────────┼──────────┼────────────────────────────────────────────────────────────────────────────┤
│ 7.0.0.alpha-avoidSelectorEvaluation │ 56.75   │ 128.6, 0.3   │ 3937.45   │ 2915.23   │ 1803.65  │ 58,57,60,59,60,59,56,58,59,58,60,59,58,57,60,57,55,54,53,55,52,51,49,53,53 │
├─────────────────────────────────────┼─────────┼──────────────┼───────────┼───────────┼──────────┼────────────────────────────────────────────────────────────────────────────┤
│ 7.0.0.alpha-useSelectorNoDeps       │ 56.87   │ 126.1, 0.4   │ 3993.32   │ 2951.71   │ 1872.79  │ 58,60,59,60,58,57,58,60,59,58,59,58,56,59,56,53,55,53,56,55,51,54,54       │
└─────────────────────────────────────┴─────────┴──────────────┴───────────┴───────────┴──────────┴────────────────────────────────────────────────────────────────────────────┘
Running benchmark forms                                                                              
  react-redux version: 7.0.0.alpha-avoidSelectorEvaluation
    Checking max FPS... (30 seconds)                                                                 
    Running trace...    (30 seconds)                                                                 
  react-redux version: 7.0.0.alpha-useSelectorNoDeps             
    Checking max FPS... (30 seconds)                                                                 
    Running trace...    (30 seconds)                                                                 
                                                  
Results for benchmark forms:         
┌─────────────────────────────────────┬─────────┬──────────────┬───────────┬───────────┬──────────┬─────────────────────────────────────────────────────────────────────────┐
│ Version                             │ Avg FPS │ Render       │ Scripting │ Rendering │ Painting │ FPS Values                                                              │
│                                     │         │ (Mount, Avg) │           │           │          │                                                                         │
├─────────────────────────────────────┼─────────┼──────────────┼───────────┼───────────┼──────────┼─────────────────────────────────────────────────────────────────────────┤
│ 7.0.0.alpha-avoidSelectorEvaluation │ 56.26   │ 943.0, 0.2   │ 6138.50   │ 972.85    │ 2727.64  │ 55,57,56,55,57,55,57,56,57,60,56,57,56,53,58,56,57,55,57,55,56,55,55    │
├─────────────────────────────────────┼─────────┼──────────────┼───────────┼───────────┼──────────┼─────────────────────────────────────────────────────────────────────────┤
│ 7.0.0.alpha-useSelectorNoDeps       │ 55.58   │ 898.3, 0.3   │ 6272.87   │ 1015.13   │ 2758.28  │ 56,57,56,55,56,55,57,56,57,56,57,56,57,52,58,57,53,56,54,56,55,50,55,55 │
└─────────────────────────────────────┴─────────┴──────────────┴───────────┴───────────┴──────────┴─────────────────────────────────────────────────────────────────────────┘
Running benchmark stockticker                                                                        
  react-redux version: 7.0.0.alpha-avoidSelectorEvaluation
    Checking max FPS... (30 seconds)
    Running trace...    (30 seconds)
  react-redux version: 7.0.0.alpha-useSelectorNoDeps
    Checking max FPS... (30 seconds)
    Running trace...    (30 seconds)

┌─────────────────────────────────────┬─────────┬──────────────┬───────────┬───────────┬──────────┬────────────────────────────────────────────────────────────────────────────┐
│ Version                             │ Avg FPS │ Render       │ Scripting │ Rendering │ Painting │ FPS Values                                                                 │
│                                     │         │ (Mount, Avg) │           │           │          │                                                                            │
├─────────────────────────────────────┼─────────┼──────────────┼───────────┼───────────┼──────────┼────────────────────────────────────────────────────────────────────────────┤
│ 7.0.0.alpha-avoidSelectorEvaluation │ 38.37   │ 170.5, 3.3   │ 18877.63  │ 7037.41   │ 2102.51  │ 40,39,41,38,39,40,38,39,41,38,39,38,39,40,38,37,36,38,37,36,36             │
├─────────────────────────────────────┼─────────┼──────────────┼───────────┼───────────┼──────────┼────────────────────────────────────────────────────────────────────────────┤
│ 7.0.0.alpha-useSelectorNoDeps       │ 37.07   │ 168.9, 3.3   │ 19129.98  │ 6931.06   │ 2035.21  │ 39,40,42,39,37,38,37,36,39,35,36,37,35,34,37,38,39,37,35,36,37,34,36,37,37 │
└─────────────────────────────────────┴─────────┴──────────────┴───────────┴───────────┴──────────┴────────────────────────────────────────────────────────────────────────────┘
Running benchmark tree-view
  react-redux version: 7.0.0.alpha-avoidSelectorEvaluation
    Checking max FPS... (30 seconds)
    Running trace...    (30 seconds)
  react-redux version: 7.0.0.alpha-useSelectorNoDeps
    Checking max FPS... (30 seconds)
    Running trace...    (30 seconds)

┌─────────────────────────────────────┬─────────┬──────────────┬───────────┬───────────┬──────────┬───────────────────────────────────────────────────────────────────────────────┐
│ Version                             │ Avg FPS │ Render       │ Scripting │ Rendering │ Painting │ FPS Values                                                                    │
│                                     │         │ (Mount, Avg) │           │           │          │                                                                               │
├─────────────────────────────────────┼─────────┼──────────────┼───────────┼───────────┼──────────┼───────────────────────────────────────────────────────────────────────────────┤
│ 7.0.0.alpha-avoidSelectorEvaluation │ 50.04   │ 426.4, 0.9   │ 5597.07   │ 12955.30  │ 463.26   │ 55,39,53,55,42,48,53,52,54,41,52,46,49,50,54,51,55,52,54,52,48,51,48,51,50,50 │
├─────────────────────────────────────┼─────────┼──────────────┼───────────┼───────────┼──────────┼───────────────────────────────────────────────────────────────────────────────┤
│ 7.0.0.alpha-useSelectorNoDeps       │ 48.20   │ 422.7, 1.1   │ 6539.58   │ 9667.43   │ 544.14   │ 42,49,53,44,49,51,55,52,43,50,47,44,46,49,52,47,53,44,49,43,55,50,51,44,53,53 │
└─────────────────────────────────────┴─────────┴──────────────┴───────────┴───────────┴──────────┴───────────────────────────────────────────────────────────────────────────────┘
Running benchmark twitter-lite
  react-redux version: 7.0.0.alpha-avoidSelectorEvaluation
    Checking max FPS... (30 seconds)
    Running trace...    (30 seconds)
  react-redux version: 7.0.0.alpha-useSelectorNoDeps
    Checking max FPS... (30 seconds)
    Running trace...    (30 seconds)

┌─────────────────────────────────────┬─────────┬──────────────┬───────────┬───────────┬──────────┬──────────────────────────────────────────────────────────────────────────────────┐
│ Version                             │ Avg FPS │ Render       │ Scripting │ Rendering │ Painting │ FPS Values                                                                       │
│                                     │         │ (Mount, Avg) │           │           │          │                                                                                  │
├─────────────────────────────────────┼─────────┼──────────────┼───────────┼───────────┼──────────┼──────────────────────────────────────────────────────────────────────────────────┤
│ 7.0.0.alpha-avoidSelectorEvaluation │ 47.54   │ 2.2, 0.5     │ 21001.58  │ 3599.72   │ 382.40   │ 59,60,59,60,58,59,60,58,56,50,47,44,40,39,34,29,28,27,27                         │
├─────────────────────────────────────┼─────────┼──────────────┼───────────┼───────────┼──────────┼──────────────────────────────────────────────────────────────────────────────────┤
│ 7.0.0.alpha-useSelectorNoDeps       │ 47.39   │ 2.1, 0.5     │ 21157.33  │ 3660.14   │ 403.49   │ 59,60,58,59,58,59,56,59,58,59,55,57,55,54,48,46,44,41,38,34,33,32,31,30,29,28,28 │
└─────────────────────────────────────┴─────────┴──────────────┴───────────┴───────────┴──────────┴──────────────────────────────────────────────────────────────────────────────────┘
Done in 862.88s.
```

I have also tried benchmarking the same, but making the selectors perform some silly work on purpose and as I already expected the results in those cases show a much more significant improvement in performance in favor of the code of this PR.

However, I think that just with these results we should consider merging this PR :slightly_smiling_face:

markerikson:
FWIW, FPS values will fluctuate from run to run.  Anything +-3-ish FPS should be basically considered equal.

josepot:
> FWIW, FPS values will fluctuate from run to run. Anything +-3-ish FPS should be basically considered equal.

@markerikson 

Sure, this optimization is only meant to be significant for those selectors that perform computations. All the selectors of our benchmarks are trivial selectors that only select things directly from the state... What I meant by showing these results is that this change can only make things better, never worse.

Do you want me to share the results and the code that test how this change would perform on non-trivial selectors? 

Do you want me to create a new benchmark that has some non-trivial selectors?

Do you prefer me to just close this PR?

Thanks!

markerikson:
I haven't made up my mind on this one overall. Still trying to get a better understanding of the tradeoffs and weighing options. So yes, more benchmarks, examples, and edge cases would be helpful.

markerikson:
@josepot : hmm.  apologies, I'm getting the PRs mixed up :)

Gimme some time to sit down and actually evaluate this one in the next few days.

markerikson:
Awright. I don't see any harm in this.  I did swap the ternary for an if/else, though. Not a ternary fan.

PR #1282, opened 2019-05-10
Update package.json, react-dom added in peerDependencies section.

sultan99:
Added missing react-dom package to peerDependencies.

It cause problem for some plugins like [dynamic-cdn-webpack-plugin](https://github.com/mastilver/dynamic-cdn-webpack-plugin).

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
   },
   "peerDependencies": {
     "react": "^16.8.3",
+    "react-dom": "^16.8.6",
     "redux": "^2.0.0 || ^3.0.0 || ^4.0.0-0"
   },
   "dependencies": {


timdorr:
We support React Native as well, which doesn't use react-dom as a renderer. So, we can't make this a peer or regular dependency.

sultan99:
If you make it peer dependancy will it break react-native?

sultan99:
@timmorr what about optional dependancy?

timdorr:
> If you make it peer dependency will it break react-native?

No, it will just warn upon install for every RN user. 

An optional dep doesn't apply, as that is for non-production use cases.

sultan99:
@timdorr I think optional dependency is right choice:
 - [npm-optional-dependency](http://npm.github.io/using-pkgs-docs/package-json/types/optionaldependencies.html)
 - [yarn-optional-dependency]https://yarnpkg.com/en/docs/dependency-types

and there is no deal with production or non-production use cases.

By refusing it you are depriving users of use [cdn](https://www.cloudflare.com/learning/cdn/what-is-a-cdn/) and [other use cases](https://github.com/npm/npm/issues/3066).

timdorr:
No, we're not. This library works just fine with a CDN. We publish a UMD bundle with every release and link to it from the package.json for Unpkg. 

It just doesn't work for that particular Webpack plugin. You can still add a script tag manually to use it from a CDN.

PR #1283, opened 2019-05-13
Support server rendering with window polyfill

tobhult:
React warns about using useLayoutEffect during server rendering if window has been polyfilled since only the existence of window is checked in connectAdvanced.js

I suggest using the same check that React has in [ExecutionEnvironment.js](https://github.com/facebook/react/blob/master/packages/shared/ExecutionEnvironment.js).


timdorr:
Thanks!

markerikson:
Ooooo, nice!  Thanks!

javache:
Is this logic correct when using react-redux in React Native? window.document and window.document.createElement do not exist there but we should still use useLayoutEffect.

PR #1285, opened 2019-05-15
fixed small typo in docs

jonyonson:
"Prefer `useSelector()` as your primary choices." => "...as your primary choice."


markerikson:
That's what happens when we delete an alpha API :)  Thanks!

PR #1290, opened 2019-05-20
Add useShallowEqualSelector

perrin4869:
The last PR exposed `shallowEqual`, with the intention of facilitating the implementation of a `useShallowEqualSelector`, using the new `equalFn` option of `useSelector`:

```js
const useShallowEqualSelector = selector => useSelector(selector, shallowEqual)
```

However, since this is already expanding the public API and there are no other current use-cases for `shallowEqual` to be used other than an argument to `useSelector`, I do not see why not go ahead and expose the new hook instead. The `equalFn` to `useSelector` remains there, if someone wants to provide their own custom one.

### DIFF
--- a/src/hooks/useSelector.js
+++ b/src/hooks/useSelector.js
@@ -20,10 +20,6 @@ const refEquality = (a, b) => a === b
  * A hook to access the redux store's state. This hook takes a selector function
  * as an argument. The selector is called with the store state.
  *
- * This hook takes a dependencies array as an optional second argument,
- * which when passed ensures referential stability of the selector (this is primarily
- * useful if you provide a selector that memoizes values).
- *
  * @param {Function} selector the selector function
  * @param {Function} equalityFn the function that will be used to determine equality
  *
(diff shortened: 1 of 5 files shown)


markerikson:
I'll leave this open for the moment, but I don't want to actually add another hook to the list right now.

timdorr:
I don't want to add this via a new function either. The correct way is to add options and arguments to existing functions, rather than packaging them up in different configurations.

perrin4869:
Oh well I tried lol
`shallowEqual` was exported with the explicit purpose of using it as an argument to `useSelector`, so I don't understand why you'd leave it up to your users to connect them together, especially when `shallowEqual` has, at least as of now, no other uses other than as an argument as of now, and has no real relevance to the library. I think if you're not going to export `useShallowEqualSelector` you might as well not export `shallowEqual` and leave it up to the users to bring their own. But that's just my two cents
Thanks for all your hard work and look forward to the final `7.1` release!

PR #1295, opened 2019-05-25
Update jsdocs for hooks

trappar:
useSelector: Not having the second argument be optional in the jsdoc is making my IDE complain "Invalid number of arguments, expected 2"

useDispatch: My IDE sees this {any} return value and doesn't assume it can be called. I get "Method expression is not of function type". Suggesting it might be a function seems safe enough.


markerikson:
What IDE are you using?

trappar:
Webstorm. This will be the case in all IntelliJ IDEs though.

markerikson:
Okay, seems reasonable.

PR #1296, opened 2019-05-26
mark equalityFn parameter of useSelector as optional in JSDoc comment

MrWolfZ:


josepot:
I *think* that #1295 does the exact same thing but with a slightly different syntax.

MrWolfZ:
Yeah, I didn't see that one. I'll close this.

PR #1297, opened 2019-05-29
fix doc: fix link on why-use-react-redux.md

toshihidetagami:
Current link on [React/Redux Links] takes you to the reddit which is already mentioned above. I changed it to presumably correct page.


timdorr:
Thanks!

PR #1301, opened 2019-06-05
This seems to be useless

alex89lj:
React is not being used


timdorr:
It's implicitly used by the JSX at the bottom. That gets transpiled to `React.createElement` calls, so the import is actually needed.

PR #1307, opened 2019-06-11
Fix link on docs

halitogunc:
Some of the links take you to the wrong url which does not exist. This PR fixes it hopefully.

### DIFF
--- a/docs/api/connect.md
+++ b/docs/api/connect.md
@@ -67,9 +67,9 @@ const mapStateToProps = (state, ownProps) => ({
 
 Your `mapStateToProps` functions are expected to return an object. This object, normally referred to as `stateProps`, will be merged as props to your connected component. If you define `mergeProps`, it will be supplied as the first parameter to `mergeProps`.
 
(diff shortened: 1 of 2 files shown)


timdorr:
Thanks!

markerikson:
This seems wrong and hacky to me.

The docs links are _supposed_ to be based on the Markdown file paths, not the final web URL links.

I did note that we had some issues with relative file links when I was fixing up the hooks docs page for final release last night.  @endiliey 	pointed out that Docusaurus really doesn't like relative links - they _should_ be absolute instead.

Can we get another PR that changes all these links to be absolute instead?  Like `using-react-redux/connect-extracting-data-with-mapStateToProps.md`, I think.

halitogunc:
I am a bit confused about that topic. I don't have much experience on Docusaurus but its structure looks so dependent to relative links. Also there is an open issue regarding docs #1001.  I might be missing something, need guidance :smile: . On the other hand we can keep the docs as it is and fix it on referenced issue above.

Issue #1304, opened 2019-06-10
[feature-request] Expose custom context to hooks


ryaninvents:
**What is the current behavior?**

The `Provider` component allows consumers to pass a custom `context`, but the current hooks API does not allow that custom context to be used.

**What is the expected behavior?**

I'd like to be able to access a custom context via React Redux hooks. I'd expect this would be implemented with a "hook factory" pattern for performance reasons, but that's an implementation detail.

**Which versions of React, ReactDOM/React Native, Redux, and React Redux are you using?  Which browser and OS are affected by this issue? Did this work in previous versions of React Redux?**

This issue is unique to the `v7-hooks` branch. I suppose you could say this has never worked in previous versions of React Redux, since hook support is new 🙂or, it might be argued that the HOC pattern was able to access alternate contexts, so hook support would be incomplete without it.

**Note:** I have a PR prepared; I didn't read the contributing guidelines until after I'd written the code, which is why I'm doubling back and creating the discussion issue now.

timdorr:
What API are you envisioning for this? I think the factory approach would work, but it would be helpful to see some code.

ryaninvents:
I've got [a branch](https://github.com/ryaninvents/react-redux/blob/feat%2Fhooks-with-custom-context/docs/api/hooks.md#createcustomhooks) with the following as a proposal:

> ## `createCustomHooks()`
> 
> This utility function returns a set of hooks bound to a different context. You might use this to help manage a very complex reusable component, allowing it to coexist within a Redux application using its own store instance without colliding.
> 
> ```js
> import React from 'react'
> import { createCustomHooks, Provider } from 'react-redux'
> import createComponentStore, { increment } from './component-store'
>
> const customContext = React.createContext()
> const customStore = createComponentStore()
>
> const customHooks = createCustomHooks(customContext)
>
> const Increment = () => {
>   const dispatch = customHooks.useDispatch()
>   return (
>     <button onClick={() => dispatch(increment())}>+1</button>
>   )
> }
>
> const InternalCount = () => {
>   const count = customHooks.useSelector(s => s.count)
>   return (
>     <span>{String(count)}</span>
>   )
> }
>
> export const ReusableCounterComponent = ({ children }) => {
>   return (
>     <Provider context={customContext} store={customStore}>
>       <div>
>         <Increment /> <InternalCount />
>         {children}
>       </div>
>     </Provider>
>   )
> }
> ```
>
[...]

PR #1309, opened 2019-06-11
feat: implement `create*Hook` APIs

ryaninvents:
As discussed in issue #1304.

For the `create*Hook` factory functions, I was debating between using `context` as the argument to all of them, vs asking users to call `createReduxContextHook` and pass their custom `useReduxContext` as the argument to the remaining three. I chose the latter since it seemed like less churn, but I'd be happy to explore other options or refactor as needed.


src/hooks/useDispatch.js
```diff
@@ -1,4 +1,20 @@
-import { useStore } from './useStore'
+import { useStore as useDefaultStore, createStoreHook } from './useStore'
+
+/**
+ * Hook factory, which creates a `useDispatch` hook bound to a given context.
+ *
+ * @param {Function} [useReduxContext] Hook which returns the Redux context.
+ * @returns {Function} A `useDispatch` hook bound to the specified context.
+ */
+export function createDispatchHook(useReduxContext = null) {
```
ryaninvents: `useDispatch` relies on `useStore` for its implementation, not `useReduxContext`. I was tempted to use the signature `createDispatchHook(useStore)` instead of the current `createDispatchHook(useReduxContext)`, but I thought it might lead to user error since the remaining hook factories accept `useReduxContext`.

src/hooks/useSelector.js
```diff
@@ -16,33 +16,12 @@ const useIsomorphicLayoutEffect =
...
- * export const CounterComponent = () => {
- *   const counter = useSelector(state => state.counter)
- *   return <div>{counter}</div>
- * }
- */
-export function useSelector(selector, equalityFn = refEquality) {
-  invariant(selector, `You must pass a selector to useSelectors`)
-
-  const { store, subscription: contextSub } = useReduxContext()
+function useSelectorWithStoreAndSubscription(
```
ryaninvents: I factored this out to make the diff easier to read, but can inline it again if you'd prefer.

timdorr:
So, I was kind of expecting this:
```js
export const myContext = React.createContext(null)

export const useStore = createStoreHook(myContext)
export const useDispatch = createDispatchHook(myContext)
export const useSelector = createSelectorHook(myContext)
```
I don't think creating a `useContext` factory adds a ton of value.  I think it's just as simple as `context === ReactReduxContext ? useReduxContext() : useContext(context)`.

markerikson:
As a vaguely related note - @drmcda made some kind of suggestion on Reddit about [somehow using closures to avoid needing a Provider at all](https://www.reddit.com/r/reactjs/comments/bz6vcu/reactredux_710_the_one_with_hooks_released/eqr02cq/).  I don't actually get what he's suggesting there, but maybe he can pop in and explain?

src/hooks/useReduxContext.js
```diff
@@ -18,8 +19,8 @@ import { ReactReduxContext } from '../components/Context'
  *   return <div>{store.getState()}</div>
  * }
  */
-export function useReduxContext() {
-  const contextValue = useContext(ReactReduxContext)
+export function useReduxContext(context = ReactReduxContext) {
```
timdorr: I'm not sure there's much value in making this change. It's essentially a useContext + invariant. I'd just leave this as-is.

ryaninvents:
@markerikson I think that Reddit thread describes what I was originally envisioning, but Tim described the potential issues in [this comment](https://github.com/reduxjs/react-redux/issues/1304#issuecomment-500852772).

timdorr:
OK, I think this looks good. Any further comments from anyone else?

markerikson:
Just out of curiosity, was there any discussion or thought put into just updating the hooks to accept a context object either as a direct parameter or in an options object?

dai-shi:
If it's still not too late, I'd like to open a discussion to create all hooks and a provider in one function.

```javascript
const { Provider, useStore, useDispatch, useSelector } = createCustomHooks(myContext);
```

This could make type inference easier for DefinitelyTyped and flow-typed.

> thought put into just updating the hooks to accept a context object

To me, if we go with`create*Hook`, `createProvider(myContext)` seems more consistent. Or, accept a context object for hooks and Provider, which I did [so](https://github.com/dai-shi/reactive-react-redux/blob/cde66a5b147aff767d8d3da341e36b63e6dd9588/src/useDispatch.js#L6-L8) without any thoughtful reason.

timdorr:
> Just out of curiosity, was there any discussion or thought put into just updating the hooks to accept a context object either as a direct parameter or in an options object?

I'm much more in favor of composition over configuration. We've already established a 2nd arg on useSelector, so now remembering argument order would come into play. Also, we'll have to handle defaults or do the polymorphic stuff that createStore does. A factory function feels like a better option, especially when it makes reuse easy.

ryaninvents:
@markerikson I thought about adding an extra parameter, but then all hooks using the default context (which would be the vast majority) would see a very small performance hit. I realize that it's unlikely that we'd even be able to measure the perf hit, but on principle I felt as though this enhancement for the edge case shouldn't impact the vast majority of users who would never run into this problem.

@dai-shi There's some related discussion in the [feature request issue](https://github.com/reduxjs/react-redux/issues/1304#issuecomment-500827537); I'd originally implemented it the way you describe.

joshjg:
@markerikson @timdorr @ryaninvents Anything blocking this from moving forward? Any help needed?

timdorr:
Just updated against master. Provided that passes, I'll merge this in.

mpeyper:
@timdorr just a gentle nudge to come back and merge this.

viniciusffj:
Is this already published on NPM anytime soon?
I can see it in your docs, but it seems to not be available on NPM yet.

timdorr:
Not yet. I need to separately version the master branch in the docs so that it's clear this isn't in the released version.

viniciusffj:
Yes, it would be great. Yesterday I spent a long time trying to use this feature by following the docs. Let me know if there is anything that I can do to contribute with that split between doc/master

timdorr:
This is out as [7.1.1](https://github.com/reduxjs/react-redux/releases/tag/v7.1.1) now!

PR #1311, opened 2019-06-11
Add missing import in hooks documentation example

morgsmccauley:


markerikson:
Heh, good catch :)  Thanks!

PR #1314, opened 2019-06-12
Fix releaseTags object

raybooysen:
Fixes the `versionToReleaseTags` object so links to release notes work for 7.1 and 7.0


timdorr:
Thanks! I'm a docusaurus newbie.

Issue #1322, opened 2019-06-14
code comment recommends using useActions hook in most cases where one would want to use the useDispatch hook


bmwadforth:
There is a comment in the source code within the useDispatch file located under /src/hooks/useDispatch  that recommends instead of using useDispatch, it is recommended to use the useActions hook in most cases. This comment contradicts the documentation found [here](https://react-redux.js.org/next/api/hooks#recipe-useactions). 

Specifically, the documentation mentions the useActions hook has been removed.

Thus, the comment within the useDispatch source code should be updated.

timdorr:
Can you submit a PR?

PR #1324, opened 2019-06-14
Removing redundant code comment. Comment referenced a hook that no lo…

bmwadforth:
Per issue #1322 
Removing a block of comments that reference a now redundant hook.


timdorr:
Thanks!

PR #1336, opened 2019-06-21
docs: annotating connected components with Flow

wgao19:
What does this PR do?
---
Addressing #1001, add docs for annotating connected components with Flow

Summary of Changes
---
- Original doc for annotating connected components with Flow
- Duplicate the doc for versioned docs starting from 5.x
- Edit the version sidebar files for the doc to show up on every versions

Anything to take note
---
@markerikson I think this doc is still very drafty to be used as an official doc. Could you please provide some suggestions on the writing? I'd like to work on it to make it more readable.

It's better if we can have someone on the Flow team to review the content as well.

### DIFF
--- a/docs/using-react-redux/annotating-connected-components-with-flow.md
+++ b/docs/using-react-redux/annotating-connected-components-with-flow.md
@@ -2,7 +2,7 @@
 id: annotating-connected-components-with-flow
 title: Annotating Connected Components with Flow
 hide_title: true
-sidebar_label: Annotating Connected Components with Flow
+sidebar_label: Flow
 ---
 
 # Annotating Connected Components with Flow
--- a/docs/using-react-redux/annotating-connected-components-with-flow.md
+++ b/docs/using-react-redux/annotating-connected-components-with-flow.md
@@ -119,10 +119,7 @@ type Props = {|
 const Com = (props: Props) => <div>{props.passthrough} {props.fromStateToProps}</div>
 
 type State = {a: number};
-type InputProps = {
-  forMapStateToProps: string
-};
-const mapStateToProps = (state: State, props: InputProps) => {
+const mapStateToProps = (state: State, props: OwnProps) => {
   return {
     fromStateToProps: 'str' + state.a
   }
--- a/docs/using-react-redux/annotating-connected-components-with-flow.md
+++ b/docs/using-react-redux/annotating-connected-components-with-flow.md
@@ -18,28 +18,104 @@ In general, to make Flow happy with connect after 0.85 is a two-phase fix. First
 
 ## Fixing the “implicitly instantiated” errors at calls of `connect`
 
-To begin, you need:
+> **Note:** We need `React.AbstractComponent` from Flow v0.89+
 
-- Flow upgraded to v0.89+
-- [Newest Flow Typed library definition for React Redux](https://github.com/flow-typed/flow-typed/blob/master/definitions/npm/react-redux_v5.x.x/flow_v0.89.x-/react-redux_v5.x.x.js)
+### Annotating at function return
 
-We include three major use cases for annotating `connect` with Flow:
+The easiest way to annotate connected components is to annotate at function call return. To do this, we need to know to types of props in our components:
+
+- `OwnProps`: likely contain or equal to what you need as the second parameter to `mapStateToProps`. If there are props that are not used by `mapStateToProps`, i.e., the props that "pass through", include them here in `OwnProps` as well
+- `Props`: `OwnProps` plus the props passed in by `mapStateToProps` and `mapDispatchToProps`
+
+> **Note:** Inexact objects don't spread nor `$Diff` very well. It is strongly recommended that you use exact objects for connected components all the time.
+
+```js
+type OwnProps = {|
+  passthrough: string,
+  forMapStateToProps: string
+|}
+
+type Props = {|
+  ...OwnProps,
+  fromMapStateToProps: string,
+  dispatch1: () => void
+|}
+```
+
+With `OwnProps` and `Props` in figured out, we are now ready to annotate the connected components.
+
+In _component definition_, annotate the props with `Props`. The component will have access to all the injected props from `connect`:
+
+```jsx
+import * as React from 'react'
+
+const MyComponent = (props: Props) => (
+  <div onClick={props.dispatch1}>
+    {props.passthrough}
+    {props.fromMapStateToProps}
+  </div>
+)
+```
+
+When we export, this is also when we normally call `connect`, annotate the exported component with _just_ `OwnProps`:
+
+```jsx
+import * as React from 'react'
+
+// const MyComponent = ...
+
+export default (connect(
+  mapStateToProps,
+  mapDispatchToProps
+)(MyComponent): React.AbstractComponent<OwnProps>)
+```
+
+### Annotating by providing explicit type parameters
+
+We may also annotate connected components by providing explicit type parameters at call of `connect` with the help of the [newest Flow Typed library definition for React Redux](https://github.com/flow-typed/flow-typed/blob/master/definitions/npm/react-redux_v5.x.x/flow_v0.89.x-/react-redux_v5.x.x.js). Note that this will also require Flow v0.89+.
+
+The Flow Typed library definition declares `connect` as follows:
+
+```js
+declare export function connect<-P, -OP, -SP, -DP, -S, -D>(
+  mapStateToProps?: null | void,
+  mapDispatchToProps?: null | void,
+  mergeProps?: null | void,
+  options?: ?Options<S, OP, {||}, MergeOP<OP, D>>
+): Connector<P, OP, MergeOP<OP, D>>
+```
+
+The libdef also contains a [glossary of the abbreviations](https://github.com/flow-typed/flow-typed/blob/master/definitions/npm/react-redux_v5.x.x/flow_v0.89.x-/react-redux_v5.x.x.js#L14) which decrypts the signature to:
+
+```jsx
+connect<Props, OwnProps, StateProps, DispatchProps, State, Dispatch>(…)
+```
+
+For the most common ways to connect components, we won't need all of the parameters. Normally, we need only `OwnProps` and `Props` at the call of `connect`, and `State` at the definition of `mapStateToProps`.
+
+We may use `_` ([what's this?](https://github.com/facebook/flow/commit/ec70da4510d3a092fa933081c083bd0e513d0518)) as placeholder at unused type parameter positions. A common `connect` call may look like this:
+
+```jsx
+connect<Props, OwnProps, _, _, _, _>(…)
+```
+
+We include examples for three major use cases of annotating `connect` with Flow:
 
 - Connecting stateless component with `mapStateToProps`
 - Connecting components with `mapDispatchToProps` of action creators
 - Connecting components with `mapStateToProps` and `mapDispatchToProps` of action creators
 
-### Connecting stateless component with `mapStateToProps`
+#### Connecting stateless component with `mapStateToProps`
 
 ```jsx
 type OwnProps = {|
   passthrough: number,
   forMapStateToProps: string,
 |};
-type Props = {
+type Props = {|
   ...OwnProps,
   fromStateToProps: string
-};
+|};
 const Com = (props: Props) => <div>{props.passthrough} {props.fromStateToProps}</div>
 
 type State = {a: number};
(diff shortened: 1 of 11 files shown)


website/versioned_docs/version-5.x/using-react-redux/annotating-connected-components-with-flow.md
```diff
@@ -0,0 +1,178 @@
...
+
+If you are at the unfortunate position where your component is wrapped with nested higher order component, it is probably more difficult to annotate by providing explicit type parameters, as doing so will probably require tediously deducing prop types at each layer:
+
+```jsx
+import { compose } from "redux";
+
+type AllProps = {/** ... */};
+
+compose(
+  connect(mapState, mapDispatch),
```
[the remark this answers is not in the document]
jbrown215: [Here's an example of what I mean](https://flow.org/try/#0PTACDMBsHsHcCh4gAQFEAeBDAtgB0gKbLTjICuAzgJYB2A5shbgE4GYAmytFALm5yWQASACJVwpHtC40AVgQDGPZC2i4KiBdBq9kAJTZLkAXmSsAjmSqsAFAHJWmJXYCUAbkQ8AnriIBJOUU+dgAFZjUKAEYTZABvAB9kcGhoAC5kGjJsACMCZmR4gF8PeHZFSExWJLIaJSptZAAJAHkAYUiAHlbtcCo6AD4beGRkbrxtAhoedIMnHgA6AEFs3mY5sdwJqY6E5Hn97ppeugAaPf2A+SUCUPD1aKL+k-gXGcMF5dX16HGaSZ4uj0+v1hsgkMBkABaaEw0Hwby+ZCXII3MIRABMMV22Uq6UyOTyBWKiDKCgqVXANTqDRarXRgKOwKGIw2W2m+neSxWPDWSlZf22u3280OxzOwuR11uGKJTxebzmXK+fJ+mwFANFwMQKBhsPBTTyBDsFGQsAAFobTURYFRIJBkHMyJg7V5TWtcMQyPktL9-vN9QAxay8M6wIgKTA0LjjZjKHgW5AIoiCeNEWhXYIqO4UM6pqNh-URqOUIip7CJ6RaHQ8shGaBe5Bo9SJnwEeGt5DNWA0Jsm0zYzAALzxWVy+SKHiTjezWPioOFXZ72eeIwlgSlvciK-O80lwV76OeE80FQoJv5-2QBHQfBo7BNsyUItVbI6vf6cUK2ohADk4MgLVYY0AP-MNkCLB0aBoaAeEwPhEwTa9NljG5wJfdVUngbBoHYMhCHmJDoFjPtkBsWlIjItp0RsC8phcV4OUVT4eW+X1tkXd93CAA)

docs/using-react-redux/annotating-connected-components-with-flow.md
```diff
@@ -0,0 +1,177 @@
...
+```
+
+### Connecting components with `mapDispatchToProps` of action creators
+
+```jsx
+type OwnProps = {|
+  passthrough: number,
+|};
+type Props = {
+  ...OwnProps,
```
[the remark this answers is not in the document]
wgao19: I see, thank you for sharing this @jbrown215! I think we don't have to use `$Diff` at all. Let me either remove it completely or say that it's not recommended.

docs/using-react-redux/annotating-connected-components-with-flow.md
```diff
@@ -0,0 +1,177 @@
...
+
+```jsx
+type OwnProps = {|
+  passthrough: number,
+  forMapStateToProps: string,
+|};
+type Props = {
+  ...OwnProps,
+  fromStateToProps: string
+};
```
villesau: Here you mix exact & inexact objects. I'd stick with either exact or inexact (but would definitely prefer exact objects as that's the future proof way)

villesau: Same with several other places.

website/versioned_docs/version-5.x/using-react-redux/annotating-connected-components-with-flow.md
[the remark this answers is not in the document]
villesau: In general $Diff should not be needed, but it could be OK to show it as a _special case_. e.g eif there are some complex cross-component requirements for props types, it might just be easier to add some instead, although in the long run I'd still claim that spreading would make things easier.

villesau:
Great initiative!

What caught my eye was the mix between exact & inexact props which you should not do as the result might be unexpected. In docs I would mostly use just exact objects & spreading. Then inexact & $Diff cases could be mentioned as a special cases.

wgao19:
Hi @villesau thanks for the tip! I'll change everything to exact in this doc. Do you have a pointer to where I can learn more about the inexact `$Diff` cases? I only know it's tricky but the whys and hows are still a bit blurry to me.

wgao19:
Hey @villesau @jbrown215 I've updated the content addressing the discussions above. Could you please review again? 

Summary of changes:
- Begin with an added section for annotation at function return, this shall be the simplest way of annotating connect, and does not require the libdef.
- Revised libdef intro to be more friendly
- Added notes on recommended use of exact objects for props
- Removed the chunk with `$Diff`

Take note that the two markdown files have the same content. This is to serve versioned docs. You don't have to read both. [Here](https://deploy-preview-1336--react-redux-docs.netlify.com/using-react-redux/annotating-connected-components-with-flow) is the link to a preview of what this doc will look like when it lands.

docs/using-react-redux/annotating-connected-components-with-flow.md
```diff
@@ -0,0 +1,255 @@
...
+
+```jsx
+import * as React from 'react'
+
+// const MyComponent = ...
+
+export default (connect(
+  mapStateToProps,
+  mapDispatchToProps
+)(MyComponent): React.AbstractComponent<OwnProps>)
```
villesau: Type casting should not be needed. In overall type casting should be avoided IMO. You can probably do something like this:

```
export default connect<Props, OwnProps, _, _, _>(
  mapStateToProps,
  mapDispatchToProps
)(MyComponent)
```

docs/using-react-redux/annotating-connected-components-with-flow.md
```diff
@@ -0,0 +1,255 @@
...
+  ...OwnProps,
+  fromStateToProps: string
+|};
+const Com = (props: Props) => <div>{props.passthrough} {props.fromStateToProps}</div>
+
+type State = {a: number};
+type InputProps = {
+  forMapStateToProps: string
+};
+const mapStateToProps = (state: State, props: InputProps) => {
```
villesau: InputProps also gets `passthrough`, so you should probably use `OwnProps` here and get rid of  `InputProps` altogether.

docs/using-react-redux/annotating-connected-components-with-flow.md
wgao19: > In what kind of cases would you prefer casting over type parameters? If there are valid use cases, there might be room for improvement in the libdefs as well.

In our code base we have a lot of nested higher order components alongside connect. So instead of doing this:

```jsx
import { compose } from "redux";
type AllProps = {/** ... */};
type withAProps = {/** ... */};
// and so forth
compose(
  connect(mapState, mapDispatch), // T_T
  withA<$Diff<$Diff<$Diff<AllProps, withDProps>, withCProps>, withBProps>>,
  withB<$Diff<$Diff<AllProps, withDProps>, withCProps>>,
  withC<$Diff<AllProps, withDProps>>,
  withD<AllProps>
)(ComponentWithEverything);
```

We figured we would do this instead:

```jsx
type OwnProps = {/** props essential to this component */};
type AllProps = { // spreaded props should be exact
  ...OwnProps, ...withAProps, ...withBProps, ...withCProps, ...withDProps,
}
const ComponentWithEverything = (props: AllProps) => {/** ... */};
export default (
  compose(withA, withB, withC, withD)
    (ComponentWithEverything): 
      React.AbstractComponent<OwnProps>
);
```

I previously included the description to this situation, but removed it since `$Diff` is tricky. And I put the casting up front because it requires less set up. I am not 100% sure about the best practices though, any suggestions?

docs/using-react-redux/annotating-connected-components-with-flow.md
wgao19: I find it more straightforward to just use `OwnProps` in both places as well but I wasn't sure since the `connect_test` uses `InputProps`. Thanks for clarifying!

docs/using-react-redux/annotating-connected-components-with-flow.md
villesau: @jbrown215 what do you think about above? Casting feels quite nasty to me, but the example above looks almost impossible to type. Redux `compose` is an alias to Flows internal compose

docs/using-react-redux/annotating-connected-components-with-flow.md
villesau: Yeah, in a way the `InputProps` _is_ correct as long as it's not exact: It has forMapStateToProps _and potentially anything else_ too. But on the other hand as you already have `OwnProps` having all the props this receives why not using it here too? :)

wgao19: Sweet, thank you for the explanation!

docs/using-react-redux/annotating-connected-components-with-flow.md
wgao19: What's the downside of casting?

I realize that within each of the hocs, providing explicit type parameters helps the hocs to deduce their own return component types, but we're only feeding them to the next layer. So it seems to me that it's not as necessary since all of those calls stay within one file. And given it's not practically possible as we have some big hundreds of components like that 😅, the workaround we figured out was to annotate at function call return.

villesau: >  the workaround we figured out was to annotate at function call return.

This is the point. Should workarounds be documented as a recommended way of working? Maybe if it's explicitly explained that it's a workaround.

There is a risk that you introduce some unintended discrepancies. see e.g these examples:
https://flow.org/try/#0BTDeEMC4AIEYCYDMBfGoBGB+GA7ArgLboCmATsgJRpTT5FmUDcAUM2DQimlroSeRRZtgFaAF4AfNFCpowGnX6jJ0AG4B7AJYATQUA

jbrown215: I actually am a big fan of casting. It has no runtime effect, so it really is just asking flow to ensure that the value matches the type. An equivalent way of doing this without casting is just:
```
// Before
module.exports = (SomeExpr: SomeType);

// After
const exports: SomeType = SomeExpr;
module.exports = exports;
```

IMO, you should just use whichever syntax is easier to express your intent. Sometimes that's explicit type args, other times that's a cast.

wgao19:
Hey @markerikson would you like to take a look at this piece now?

docs/using-react-redux/annotating-connected-components-with-flow.md
wgao19: @villesau I’m trying to understand this example you linked — is it casting the object twice and therefore producing the discrepancies?

villesau: It's just an example where you can cast the object to something else, but can't actually cast it back to original type. This is totally by design: Inexact objects allows that. But it also makes it possible to make mistakes like in the example. The point is that you might loose some type information with casting if you are not careful.

markerikson:
I'm mostly away atm, so I don't have time to go through this in detail.

I think my main observation atm is that the title and the content seem just a bit too specific for a top-level page.  Conceptually, I'd like to see either a single "Static Typing" docs page that covers both general TS and Flow usage, or perhaps a "Static Typing" sub-category within "Usage" that then has separate pages for TS and Flow, with the seemingly more-specific content here being part of the Flow page.

docs/using-react-redux/annotating-connected-components-with-flow.md
wgao19: So.. We have notes about inexact objects being tricky, I guess I shall stress more of that? 

And looks like we're not triggering the tricky serial casting thing since the two casting happen on different objects? The first one on annotating the unconnected component, the second time the exported connected component.

Meanwhile, I feel the benefit of annotating nested hocs + connect this way is larger than the potential caveat we're facing. Because we don't have hold of how the other hocs pass types information around. And so I feel annotating once at export is an expressive clean cut, after all simplicity is also helpful to reduce mistakes. We've been doing this with our code base and Flow is able to cover those connected components correctly.

wgao19:
Hey @markerikson looks like the sidebar does not support nested categories. What about I put the content about Flow into a "Static Typing" page, and leave out the TypeScript section and mention that we're looking for PR? I'm not practically experienced on the TS usage so I'll have to leave it out for now.

markerikson:
Hmm.  That's odd, because we've specifically got a nested category in the core Redux docs - the "Recipes > Structuring Reducers" section:

https://redux.js.org/recipes/structuring-reducers/structuring-reducers

Setup:

https://github.com/reduxjs/redux/blob/ff8fce055b79c63e085c4e1d394bf637523626b7/website/sidebars.json#L46-L63

Do we need to bump the Docusaurus version for React-Redux or something?

wgao19:
Hey @markerikson I created the subcategory, would you take a look at how it looks now?

timdorr:
We're not Flow experts here and don't have the expertise to know if this is correct, so I'm going to close this out. 

Until we adopt TS for this library, let's leave out static typing docs. We don't control the types, so we shouldn't have to document them (docs could easily go out of sync).

markerikson:
@timdorr : I'll disagree with that one.  We've already got a "Static Typing" docs page over in the Redux core that lists the basic approaches, and that's what I'm going for over in #1439.  I particularly want to officially document the `ConnectedProps<T>` approach publicly.

timdorr:
But we don't maintain those types. If the folks on DT want to change that interface, we have no control over that and then the docs are out of sync.

markerikson:
We haven't entirely maintained the Redux core types either.

Given the prevalence of TS usage, and that we recommend using TS, I want to officially document the basics of TS usage.

PR #1337, opened 2019-06-22
Updates package versions in Quick Start docs

rszalski:
Hello!

I just noticed these versions being outdated in [the docs](https://react-redux.js.org/introduction/quick-start#installation) for `react-redux` v7.1. Same thing happens in [docs for v7.0](https://react-redux.js.org/7.0/introduction/quick-start#installation), however I am not sure how you handle maintenance of the docs for older versions of `react-redux`.

### DIFF
--- a/docs/introduction/quick-start.md
+++ b/docs/introduction/quick-start.md
@@ -11,7 +11,7 @@ sidebar_label: Quick Start
 
 ## Installation
 
-React Redux 6.x requires **React 16.4 or later.**
+React Redux 7.1 requires **React 16.8.3 or later.**
 
 To use React Redux with your React app:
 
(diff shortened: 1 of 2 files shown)


markerikson:
Yeah, the versioning gets a bit weird.  Since we've specifically cut a 7.1 versioned set, you actually have to edit `versioned_docs/7.1/someFile.md` instead.  So, this change only winds up being useful when we have a `next`-type tag set up in the Docusaurus config.

Can you apply this change to https://github.com/reduxjs/react-redux/blob/master/website/versioned_docs/version-7.1/introduction/quick-start.md too?

rszalski:
> Yeah, the versioning gets a bit weird. Since we've specifically cut a 7.1 versioned set, you actually have to edit `versioned_docs/7.1/someFile.md` instead. 

This makes sense! Thanks for the explanation.

> Can you apply this change to https://github.com/reduxjs/react-redux/blob/master/website/versioned_docs/version-7.1/introduction/quick-start.md too?

Done :)

markerikson:
Thanks!

PR #1338, opened 2019-06-22
remove _isMounted

ee92:
it is not doing anything


timdorr:
Yep, looks like a hold-over from 6.0. Thanks!

PR #1344, opened 2019-06-30
Update react-hooks-testing-library to @testing-library package

mpeyper:
We recently moved `react-hooks-testing-library` to be part of the `testing-library` family of packages, which involved publishing under the `@testing-library` scope.  The API is the same, so it's just a matter of updating the imports.


timdorr:
Thanks!

PR #1364, opened 2019-07-26
fix(docs): remove `mapDispatchToProps` invalid use case

iamandrewluca:
Tried every form of `mapDispatchToProps`  
`dispatch` is never missing from `props`

### DIFF
--- a/docs/using-react-redux/connect-dispatching-actions-with-mapDispatchToProps.md
+++ b/docs/using-react-redux/connect-dispatching-actions-with-mapDispatchToProps.md
@@ -254,22 +254,6 @@ connect(
 )(Counter)
 ```
 
-### Manually Injecting `dispatch`
-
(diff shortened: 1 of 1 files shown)


timdorr:
Yep, that's out of date. Thanks!

markerikson:
Uh.... wait, what?  No, we don't pass `dispatch` in if you do `mapDispatch`.  We never have.

markerikson:
See the current implementation here:

https://github.com/reduxjs/react-redux/blob/ae6f033357ddfebbc44b1fae16b348b89d832f6c/src/connect/mapDispatchToProps.js#L4-L22

We return the result of `mapDispatch(dispatch)` if it's a function, the result of `bindActionCreators(mapDispatch, dispatch)` if it's an object, and `() => {dispatch}` if it's missing.

The existing docs were correct.  Reverting.

iamandrewluca:
Hmm, my bad, it seems another HOC is injecting dispatch.

timdorr:
Whoops, my bad for merging without thinking too hard!

PR #1209, opened 2019-03-22
Use Hooks internally (aka 7.0)

timdorr:
:fire::fire::fire:React-Redux v7 beta is out! :fire::fire::fire: We've just published our first beta of React-Redux version 7, and it should be the fastest version of React-Redux yet! Please try it out in your *production* apps, and let us know how well it works for you!

https://github.com/reduxjs/react-redux/releases/tag/v7.0.0-beta.0

We've rewritten connect to use hooks internally, switched back to using direct store subscriptions in components, and we're now using React's "batched updates" API to reduce re-renders. v7 is API-compatible with v6 - the major bump is primarily because we require React 16.8.

We've also added a new batch() API you can import and call in your own code to minimize re-renders with multiple dispatches. connect and batch should work correctly in ReactDOM and React Native, and there's a new alternate entry point to use with alternate React renderers.

We've tested and benchmarked, and are confident this beta is ready for you to try out in your own apps. We've got a lot of users who do things we don't know about, so if you do find an issue, please let us know in this issue.

([the other issue](https://github.com/reduxjs/react-redux/issues/1177) is getting all Big McLargeHuge.)

PR #1366, opened 2019-07-30
change react-dom from devDependency to dependency

legnaleurc:
Since #1209, we import `react-dom` explicitly, so maybe it should be put into `dependencies` instead of `devDependencies`.
Put it in `devDependencies` will confuse some plugins of build systems (e.g. Webpack), makes them think `react-dom` is not needed for `react-redux`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7155,7 +7155,6 @@
       "version": "16.8.6",
       "resolved": "https://registry.npmjs.org/react-dom/-/react-dom-16.8.6.tgz",
       "integrity": "sha512-1nL7PIq9LTL3fthPqwkvr2zY7phIPjYrT0jp4HjyEQrEROnw4dG41VVwi/wfoCneoleqrNX7iAD+pXebJZwrwA==",
-      "dev": true,
       "requires": {
         "loose-envify": "^1.1.0",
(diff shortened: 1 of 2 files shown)


markerikson:
Appreciate the PR, but no, this is not correct.

Because of how the renderers export `unstable_batchedUpdates`, we're in a weird situation where we have to import from `react-dom`, but allow the RN loader to fall back to the RN package instead when used with RN.

Adding `react-dom` as an explicit dependency would break that, and also force an extra download for all RN users.

legnaleurc:
We need `react-dom` for deploying `react-redux`, at least it appears when we run `npm run build:umd`.
Would it be acceptable to be `peerDependencies` or `optionalDependencies`?

markerikson:
Also no.  We don't want to be RN users to be getting warned when they install React-Redux but not ReactDOM.  And, since we build the UMDs ourselves as part of the dev process, that's already handled under `devDependencies`.

I know it's quirky, but what we have now works.

zkochan:
It can be an optional peer dependency. In that case, users will not be warned about it. Optional peer dependencies are supported by yarn, pnpm and will be supported in npm in the next minor version (thanks to @arcanis)

PR #1371, opened 2019-08-06
Fix add missing peer dependency

zkochan:
react-dom is used in production code, so it should be a peer dependency
https://github.com/reduxjs/react-redux/blob/fa5857281a37545c7c036fb2499159b865b1c57d/src/utils/reactBatchedUpdates.js#L2

It currently causes issues with pnpm which won't provide access to packages that
are not declared as dependencies in package.json

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
   },
   "peerDependencies": {
     "react": "^16.8.3",
+    "react-dom": "^16.8.6",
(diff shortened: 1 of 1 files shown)


markerikson:
No, it shouldn't be, because of how we're handling the imports.  Please see #1366 for prior discussion.

PR #1377, opened 2019-08-11
Convert Provider into function component with hooks

### LINKED ISSUE #1370: Provider uses previous store for one additional render when store changes

Bug

If a the `store` passed to a `Provider` changes the `store` in `ReactReduxContext` does not update until the next render.  This is causing me an issue the `children` are changing, causing the to crash when trying to `mapStateToProps` in that additional render.

[Sandbox.](https://codesandbox.io/s/thirsty-cherry-f49el)

Note: this is a contrived example to to show how this occurs, but I have an real world case using React Router V5 that has the same issue.

Also Note: my use case is not to have a distinct store per page, but actually using `redux-subspace` to create a virtual sub-store per page.  Again, I didn't bother with this detail for the example as it occurs with real stores too.

I would expect that if `Provider` permits me to change the store, that a `connect`ed component below it would not crash when rendering.  I believe the root of the issue is the fact that [the store is not updated in the `Providers`'s state until `componentDidUpdate`](https://github.com/reduxjs/react-redux/blob/master/src/components/Provider.js#L38-L45), but the render the would result in this update already has the new `this.props.children` value.

I do concede that this is a pretty niche use case, and may not be the highest priority to fix (assuming you agree that it is an issue at all).

```json
"react": "16.8.6",
"react-dom": "16.8.6",
"redux": "4.0.4"
"react-redux": "7.1.0",
```

mpeyper:
Fixes issue where the `store` and `children` values were updated in different render passes which could result in `mapState` errors if the new children were not compatible with the old store.

The change to a function component meant `Provider` could take advantage of `useEffect` to "watch" the `store` prop and unsubscribe to the old store and subscribe to the new store.

Fixes #1370

### DIFF
--- a/src/components/Provider.js
+++ b/src/components/Provider.js
@@ -24,6 +24,7 @@ function Provider({ store, context, children }) {
     }
     return () => {
       subscription.tryUnsubscribe()
+      subscription.onStateChange = null
     }
   }, [contextValue, previousState])
 
--- a/src/components/Provider.js
+++ b/src/components/Provider.js
@@ -1,62 +1,35 @@
-import React, { Component } from 'react'
+import React, { useMemo, useEffect } from 'react'
 import PropTypes from 'prop-types'
 import { ReactReduxContext } from './Context'
 import Subscription from '../utils/Subscription'
 
-class Provider extends Component {
-  constructor(props) {
-    super(props)
-
-    const { store } = props
-
-    this.notifySubscribers = this.notifySubscribers.bind(this)
+function Provider({ store, context, children }) {
+  const contextValue = useMemo(() => {
     const subscription = new Subscription(store)
-    subscription.onStateChange = this.notifySubscribers
-
-    this.state = {
+    subscription.onStateChange = subscription.notifyNestedSubs
+    return {
       store,
       subscription
     }
+  }, [store])
 
-    this.previousState = store.getState()
-  }
+  const previousState = useMemo(() => store.getState(), [store])
 
-  componentDidMount() {
-    this.state.subscription.trySubscribe()
+  useEffect(() => {
+    const { subscription } = contextValue
+    subscription.trySubscribe()
 
-    if (this.previousState !== this.props.store.getState()) {
-      this.state.subscription.notifyNestedSubs()
+    if (previousState !== store.getState()) {
+      subscription.notifyNestedSubs()
     }
-  }
-
-  componentWillUnmount() {
-    if (this.unsubscribe) this.unsubscribe()
-
-    this.state.subscription.tryUnsubscribe()
-  }
-
-  componentDidUpdate(prevProps) {
-    if (this.props.store !== prevProps.store) {
-      this.state.subscription.tryUnsubscribe()
-      const subscription = new Subscription(this.props.store)
-      subscription.onStateChange = this.notifySubscribers
-      this.setState({ store: this.props.store, subscription })
+    return () => {
+      subscription.tryUnsubscribe()
     }
-  }
+  }, [contextValue, previousState])
 
-  notifySubscribers() {
-    this.state.subscription.notifyNestedSubs()
-  }
+  const Context = context || ReactReduxContext
 
-  render() {
-    const Context = this.props.context || ReactReduxContext
-
-    return (
-      <Context.Provider value={this.state}>
-        {this.props.children}
-      </Context.Provider>
-    )
-  }
+  return <Context.Provider value={contextValue}>{children}</Context.Provider>
 }
 
 Provider.propTypes = {
--- a/test/components/Provider.spec.js
+++ b/test/components/Provider.spec.js
@@ -312,5 +312,43 @@ describe('React', () => {
       ReactDOM.unmountComponentAtNode(div)
       expect(spy).toHaveBeenCalledTimes(1)
     })
+
+    it('should handle store and children change in a the same render', () => {
+      const reducerA = (state = { nestedA: { value: 'expectedA' } }) => state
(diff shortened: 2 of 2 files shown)


src/components/Provider.js
```diff
@@ -1,62 +1,35 @@
...
-    const { store } = props
-
-    this.notifySubscribers = this.notifySubscribers.bind(this)
+function Provider({ store, context, children }) {
+  const contextValue = useMemo(() => {
     const subscription = new Subscription(store)
-    subscription.onStateChange = this.notifySubscribers
-
-    this.state = {
+    subscription.onStateChange = subscription.notifyNestedSubs
```
mpeyper: This is what I determined `subscription.onStateChange = this.notifySubscribers` to actually boiled down to.  I'm wondering if `subscription` couldn't just handle this internally?

test/components/Provider.spec.js
```diff
@@ -312,5 +312,43 @@ describe('React', () => {
       ReactDOM.unmountComponentAtNode(div)
       expect(spy).toHaveBeenCalledTimes(1)
     })
+
+    it('should handle store and children change in a the same render', () => {
```
mpeyper: I did verify that this did actually fail on the old code before making my change.

src/components/Provider.js
```diff
@@ -1,62 +1,35 @@
...
 
-    if (this.previousState !== this.props.store.getState()) {
-      this.state.subscription.notifyNestedSubs()
+    if (previousState !== store.getState()) {
+      subscription.notifyNestedSubs()
     }
-  }
-
-  componentWillUnmount() {
-    if (this.unsubscribe) this.unsubscribe()
```
mpeyper: I could not see where `this.unsubscribe` was actually being set.  I think this could have been safely removed, but the new implementation does not allow for this in any way, so if that is an issue I'll have to go back to the drawing board.

markerikson:
Hey, thanks for putting this together!  I'll have to come back and look at it a bit later, but noted.

src/components/Provider.js
markerikson: Yeah, I think this line is a leftover artifact from pre-7.0 that I missed removing.

src/components/Provider.js
timdorr: We use the same Subscription class for the `connect` components, where this is part of our top-down update mechanic. So, it needs to stay configurable. This is just a quirk of re-use.

mpeyper:
hey @markerikson/@timdorr, have either of you had a chance to take a look at this yet?

timdorr:
Looks good to me. @markerikson?

src/components/Provider.js
```diff
@@ -1,62 +1,35 @@
...
-  }
-
-  componentDidUpdate(prevProps) {
-    if (this.props.store !== prevProps.store) {
-      this.state.subscription.tryUnsubscribe()
-      const subscription = new Subscription(this.props.store)
-      subscription.onStateChange = this.notifySubscribers
-      this.setState({ store: this.props.store, subscription })
+    return () => {
+      subscription.tryUnsubscribe()
```
markerikson: Given what we just saw in #1380 , let's clear out the change callback here too:

```suggestion
     subscription.tryUnsubscribe()
     subscription.onStateChange = null
```

markerikson:
Okay, board is green.  LGTM!

mpeyper:
Hey, just curious what the release timeframe is on this.  Is there a specific milestone this will wait for or will a `7.1.1` version get pushed out soon?

markerikson:
Good question.  @timdorr , you got time to drop a new release with this and the memory clear tweak?

I'm at React Rally this week.  Busy with the conf the next two days.  Hoping to do some maintainer-ish stuff on the weekend.  If it's not out by then, I'll try to publish it.

mpeyper:
No major rush from me, just curious.

timdorr:
This is out as [7.1.1](https://github.com/reduxjs/react-redux/releases/tag/v7.1.1) now!

src/components/Provider.js
MichaelSnowden: I know this is a very old PR, but I'm wondering if you ever tested this in `StrictMode`? I'm noticing bugs where updates stop happening for us using `react-redux` 7.2.5, and I believe it's because we set `onStateChange` here in the `useMemo` callback, but then we clear it [here](https://github.com/reduxjs/react-redux/pull/1377/files#diff-ac6af22b2cd4dd0f3ae1c7a43037607d1d32f0d65fce060da62c92dcdb9eaab5R27) in the `useEffect` teardown. With strict mode, the `useMemo` callback here will run twice, but the result of the second run isn't actually used. So, the result is that children are rendered with the original subscription which is resubscribed but has no `onStateChange` callback, so it does nothing. I think the fix is to move this line to the `useEffect` call. I can patch that for us locally, but I'm wondering if you ever encountered this after making this change.

MichaelSnowden: Ahh, I see https://github.com/reduxjs/react-redux/commit/6acb58de1a2e93666bfec5e515d1aaea846e88d9 now

markerikson: @MichaelSnowden yeah, we actually fixed that bug in https://github.com/reduxjs/react-redux/releases/tag/v7.2.8 :)  latest 7.x release is 7.2.9.  (But also, we would recommend upgrading to v8.x if possible if you're still on React 17 or earlier, or v9 if you're on React 18.)

markerikson: Strict Mode would be the only real reason I can think of that this would be an issue off the top of my head.

PR #1380, opened 2019-08-15
Free resources (free detached nodes from memory)

MosheZemah:
Found an issue when using virtualized list (https://github.com/bvaughn/react-virtualized), where unmounted react components (connected to store using connect) were not collected by the garbage collector. 
I'm not sure what is the exact explanation here, but it solved my problem.

### DIFF
--- a/src/components/connectAdvanced.js
+++ b/src/components/connectAdvanced.js
@@ -370,7 +370,7 @@ export default function connectAdvanced(
           subscription.tryUnsubscribe()
           // free resources
           subscription.onStateChange = null
-          
+
           if (lastThrownError) {
             // It's possible that we caught an error due to a bad mapState function, but the
             // parent re-rendered without this component and we're about to unmount.
--- a/src/components/connectAdvanced.js
+++ b/src/components/connectAdvanced.js
@@ -369,8 +369,8 @@ export default function connectAdvanced(
           didUnsubscribe = true
           subscription.tryUnsubscribe()
           // free resources
-          subscription.onStateChange = null 
-
+          subscription.onStateChange = null
+          
           if (lastThrownError) {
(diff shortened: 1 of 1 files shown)


src/components/connectAdvanced.js
```diff
@@ -368,6 +368,8 @@ export default function connectAdvanced(
         const unsubscribeWrapper = () => {
           didUnsubscribe = true
           subscription.tryUnsubscribe()
+          // free resources
```
timdorr:
```suggestion
```

Let's leave out this comment

markerikson:
Can you put together a sandbox that shows the problem in action, so that we can check to see if this actually makes any difference?

MosheZemah:
> Can you put together a sandbox that shows the problem in action, so that we can check to see if this actually makes any difference?
> 
> I don't see any reason why this change would be a problem, I just want to see how much it helps.

Hi, thanks for replying. 
It seems like a specific case with virtualized list, where I scroll up and down and many components are mounting and unmounting (every time they enter/exit the DOM). I'm not sure I'll be able to reproduce it on a sandbox. Is there any possibility that "onStateChange" which holds reference to checkForUpdates somehow retain a reference to the component which keeps it from being garbage collected? I can add a screenshot where I have allot of Detached div elements, and I've traced them down to their holder - "Subscription" (forcing GC didn't free them). After applying my change, the problem seems to be solved.

markerikson:
Well, `connect` is now a function component, so there's no component reference that can be accessed by app logic.

In _theory_, what I would expect is:

- React unmounts the component and cleans up the hooks list
- No hooks point to the `Subscription` instance, so it's GC'd
- Anything the `Subscription` points to is GC'd

But, all that said, I see no reason why this change would be an issue, so let's get it in.

Issue #1374, opened 2019-08-09
React 16.9 Deprecation *Will* warning


BiosBoy:
**Do you want to request a _feature_ or report a _bug_?**
Bug

**What is the current behavior?**
React throw a warning about all deprecated WILL methods inside react-redux.
Example: https://i.ibb.co/mbWrWYC/Screen-Shot-2019-08-09-at-11-06-13-AM.png

**What is the expected behavior?**
As long as React works on cut out deprecation methods, this package (react-redux) should also be up to date with.

**Which versions of React, ReactDOM/React Native, Redux, and React Redux are you using?  Which browser and OS are affected by this issue? Did this work in previous versions of React Redux?**
React: 16.9.0
Redux: 16.9.0
React-Redux: 7.1.0

markerikson:
That can't be right, because we stopped using `componentWillUpdate` as of React-Redux v6, and v7 is implemented with hooks inside so it doesn't even use class components.

Please double-check the library versions you actually have installed in your app.

mikecousins:
I'm seeing this with the Provider component as well.

breadcrumbs.js:76 Warning: componentWillReceiveProps has been renamed, and is not recommended for use. See https://fb.me/react-async-component-lifecycle-hooks for details.

* Move data fetching code or side effects to componentDidUpdate.
* If you're updating state whenever props change, refactor your code to use memoization techniques or move it to static getDerivedStateFromProps. Learn more at: https://fb.me/react-derived-state
* Rename componentWillReceiveProps to UNSAFE_componentWillReceiveProps to suppress this warning in non-strict mode. In React 17.x, only the UNSAFE_ name will work. To rename all deprecated lifecycles to their new names, you can run `npx react-codemod rename-unsafe-lifecycles` in your project source folder.

Please update the following components: Provider

markerikson:
Well, as you can see, [`Provider` does _not_ use any of those lifecycles in v7](https://github.com/reduxjs/react-redux/blob/4b9cecefde7b3e4c6344d3277c46ad0471c4bbfe/src/components/Provider.js).

So, whatever you're seeing, it's not due to our current version.  We removed all the deprecated lifecycles in v6.

mikecousins:
Weird, it must be an error in something below it that's getting misreported or something...

BiosBoy:
Hm... seems like really the problem was inside misversion of imported `react-redux`. All good now, sorry.

mikecousins:
It's very strange.

[...]

PR #1383, opened 2019-08-21
Use UNSAFE_ lifecycle event names where supported.

elyobo:
This provides forward compatibility with React 17.x for those still
using the react-redux 5.x series by detecting support for the UNSAFE_
prefixed lifecycle event names (using `React.createContext` as a feature
check as it avoids semver checking but was introduced in React 16.3 at
the same time as the UNSAFE_ prefixes).

See discussion on https://github.com/reduxjs/react-redux/issues/1374 for context.


timdorr:
Any thoughts on this approach vs ReactTraining/react-router#6883?

elyobo:
@timdorr this has been updated to be consistent with the implementation in https://github.com/ReactTraining/react-router/pull/6883 (using `parseFloat` on `React.version` instead of checking for `React.createContext`.

elyobo:
The various failures don't appear to be related to the changes in this PR.

elyobo:
@timdorr any interest in merging this?  Essentially the same fix as the merged version in https://github.com/ReactTraining/react-router/pull/6883

elyobo:
No worries at all, thanks this this!

elyobo:
When you get time, a patch release for 5.x would be handy :)

gregnb:
Same would love a release for this as well

elyobo:
@timdorr `5.x` release for this still on the radar?

src/components/Provider.js
```diff
@@ -1,8 +1,10 @@
-import { Component, Children } from 'react'
+import React, { Component, Children } from 'react'
 import PropTypes from 'prop-types'
 import { storeShape, subscriptionShape } from '../utils/PropTypes'
 import warning from '../utils/warning'
 
+const prefixUnsafeLifecycleMethods = parseFloat(React.version) >= 16.3
```
samkline: @elyobo @timdorr Thanks for fixing this, but there's still one more issue here. `parseFloat("16.10.1")` parses as 16.1 which is of course less than 16.3, so this fails on the latest release of React.

timdorr: Ah, that's not good... Can you work up a fix PR?

timdorr: Actually, I got it: #1407

elyobo: Ha, we've gone back to a feature check like this PR originally shipped with before https://github.com/reduxjs/react-redux/pull/1383#issuecomment-524483129 :D  Thanks for the pick up Sam.

PR #1385, opened 2019-08-23
remove browserify config from package.json

alsotang:
Since we use rollup now, we don't need any browserify config.


timdorr:
This isn't for our use of browserify, it's for those that use browserify with this package.

PR #1387, opened 2019-08-26
Clarify two way dispatch binding in more detail

aminsoheyli:


aminsoheyli:
> Thanks!

Your welcome!

PR #1389, opened 2019-08-30
remove useless checkForUpdates()

alsotang:
## context

I start reading the react-redux source code tonight. I find here is a strange `checkForUpdates` call:

https://github.com/alsotang/react-redux/blob/5b268f3f07eaf054eca47bebfd0fcf1d8bcc96c4/src/hooks/useSelector.js#L95

I try to understand why it is here and figure out that perhaps the store would change before we subscribe it. And here is a test case to validate this situation:

https://github.com/alsotang/react-redux/blob/5b268f3f07eaf054eca47bebfd0fcf1d8bcc96c4/test/hooks/useSelector.spec.js#L135

## my exploration

I try to remove the `checkForUpdates` call and see what would happen. But all tests passed when I type `npm test`

And I put `console.log` in the first line of `checkForUpdates`. When I don't remove it, it logs twice. When I remove it, it logs once.

So I think manually call `checkForUpdates` when subscribe is a waste, but I don't know why.

## my thinking

1. Some mechanism changed, so manually call `checkForUpdates` is unnecessary.
2. We should add a specific test to ensure the call is not a waste.

### DIFF
--- a/src/hooks/useSelector.js
+++ b/src/hooks/useSelector.js
@@ -92,8 +92,6 @@ function useSelectorWithStoreAndSubscription(
     subscription.onStateChange = checkForUpdates
     subscription.trySubscribe()
 
-    checkForUpdates()
-
     return () => subscription.tryUnsubscribe()
   }, [store, subscription])
 


markerikson:
Appreciate the PR, but this is necessary.

This is specifically borrowed from [the approach shown in React's `useSubscription` hook](https://github.com/facebook/react/blob/2c1e6bf619e404fd1ac464eb35eff4c09b3861b3/packages/use-subscription/src/useSubscription.js#L106-L111).

Technically, the fact that we're doing this in `useLayoutEffect` vs `useEffect` _might_ mean the described problem isn't an issue here, but there's no actual reason to remove this.

alsotang:
Alright. Can we write some comments for this behavior as follow?

```
      // Because we're subscribing in a passive effect,
      // it's possible that an update has occurred between render and our effect handler.
      // Check for this and schedule an update if work has occurred.
```

Issue #130, opened 2015-09-29
Need peer dep on react-native?


erikras:
How can [this line](https://github.com/rackt/react-redux/blob/master/src/native.js#L1) work if there is no mention of `react-native` in `package.json`?
```js
import React from 'react-native';
```
I understand that there is an assumption that if anyone is `require`ing `native.js`, that they will already have `react-native` included in their project deps, but it seems very odd that a library could have an `import` of a package not mentioned in its `package.json`.

Am I wrong? I'm trying to understand this mechanism to duplicate it in my own project, and this is confusing me. Could you add a peer dep that didn't cause problems for people just using plain `react`, or is this just a strangeness of the npm dep system?

gaearon:
It wouldn't make sense to force installation of `react-native` onto people using `react`, and vice versa. This is why neither is in `peerDependencies` in `package.json`.

Yes, we assume the users are already using either of them. I agree it's kinda weird, but it's a better solution than forcing users to install both packages. (Especially considering they might get weird issues when `react-native` users get `react` installed side by side.)

This is going to be moot soon after 0.14 is out (and compatible RN release is out) anyway because `Component` and friends will be inside `react` package for both platforms, and `react-native` will become a yet another renderer, just like `react-dom`. When this happens we'll kill `react-redux/native` entry point, and everyone will be able to use `react-redux` directly regardless of the platform. This is when we can introduce a peer dependency.

erikras:
Got it. Thanks for the clarification.

PR #1390, opened 2019-09-05
fix: List optional peer dependencies

eps1lon:
`react-dom` is required from the main bundle (not really optional I suppose) while `react-native` is required when using it with `react-native` and their batched updates.

Additional context:
* [optional peer dependencies](https://github.com/yarnpkg/rfcs/blob/master/accepted/0000-optional-peer-dependencies.md) (implemented in `yarn@1.13.0`, `npm@6.11.0` but not [yet](https://github.com/nodejs/node/pull/29273) bundled with node and `pnpm@?`

I'm currently not using this feature myself until node bundles `npm@6.11`.  Just creating this PR now so that we can discuss any concerns you have.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,17 @@
   },
   "peerDependencies": {
     "react": "^16.8.3",
+    "react-dom": "^16.8.3",
+    "react-native": "^0.18.0",
     "redux": "^2.0.0 || ^3.0.0 || ^4.0.0-0"
+  },
+  "peerDependenciesMeta": {
+    "react-dom": {
+      "optional": true
+    },
+    "react-native": {
+      "optional": true
+    }
   },
   "dependencies": {
     "@babel/runtime": "^7.5.5",


package.json
```diff
@@ -37,8 +37,18 @@
   },
   "peerDependencies": {
     "react": "^16.8.3",
+    "react-dom": "^16.8.3",
+    "react-native": "^0.18.0",
```
eps1lon: Never used `react-native`. Just use the note from the readme:
> As of React Native 0.18, React Redux 5.x should work with React Native.

markerikson:
This has been brought up several times before.  The current before is quirky but correct, and there's no need to change.

timdorr:
Actually, this is correct, but only once npm >= 6.11 is in greater circulation. It won't show the error (the reason for this extension to the package.json format), but only for those versions.

timdorr:
BTW, npm releases are delayed 2 weeks before being bundled into node. nodejs/node#29273 is just about read to go and nodejs/node#29430 still has 12 days to go. But likely the next 12.x release will have it. I'm not sure what the LTS schedule is.

eps1lon:
>  The current behavior is quirky but correct, and there's no need to change.

I'm not sure how you define "correct" in this context but the dependency trees created from it are incorrect. All I could find about "react dom peer" in the issues is an old statement from dan (https://github.com/reduxjs/react-redux/issues/130#issuecomment-144125547) which 
1. no longer applies with peerDependenciesMeta
2. is in todays ecosystem a bit of an overstatement since a missing peer dependency warning is only a warning for a reason. Nobody is forcing you to install them and from my experience most packages seem to be ok with having them anyway.

markerikson:
@eps1lon : previous PRs:

#1261 
#1282 
#1371 
#1366

timdorr:
If they are marked optional with peerDependenciesMeta, they won't warn at all.

arcanis:
> And I don't care about "most packages". I care that this package would end up forcing everyone to see at least one warning - RN for RD users, and RD for RN users.

This is incorrect. With the optional peer dependencies, peer dep warnings will only be printed if:

- you use an outdated npm version or,
- you use a recentish npm version AND npm hasn't fixed their server AND only on initial installs (subsequent ones don't print the warning even now).

Anyway, [I've just suggested](https://github.com/reduxjs/react-redux/pull/1425#issuecomment-541749184) what I think is an reasonable compromise: if you don't list peer dependencies, please at least put the `peerDependenciesMeta` field. Yarn will understand that it implies a peer dependencies, even if unlisted, and everyone will be happy.

markerikson:
@arcanis : my statement about wanting to avoid unnecessary warnings still stands, but no, this is the first time I've heard of "optional peer dependencies", so I'm not familiar with the distinction here in how they behave.

timdorr:
I'll add the optional stuff to the package.json now. It doesn't hurt anything to add it. And if the package managers adopt it as working without the peerDeps entry, that's great.

arcanis:
@markerikson Fwiw documentation is here:
https://next.yarnpkg.com/configuration/manifest#peerDependenciesMeta

Note that it's already implemented in both Yarn 1 ([initial rfc](https://github.com/yarnpkg/rfcs/blob/master/accepted/0000-optional-peer-dependencies.md)) and pnpm. Npm has merged my PR too, but they have to do a fix on their server to return the field on the initial installs (they use a special endpoint which removes most fields from the manifest, including the `peerDependenciesMeta` one).

The distinction is that a package manager will simply discard the warning if the peer dependency is marked `optional`, but will be able to leverage the information to properly hoist packages (unlisted peer dependencies aren't having a visual effect - they have an actual technical impact regarding how the package managers will setup the tree, hence why not listing them is dangerous for your users).

The change I've just landed makes it possible to list `peerDependenciesMeta` without listing `peerDependencies`. That will at least fix Yarn, and npm will benefit from it as well once they'll have fixed their servers (since it'll help them layout the packages correctly).

@timdorr Awesome, thanks! 🙏

markerikson:
Looks like this was resolved by 388d9e4eb28da5fad6860057c5c5bcb2fa27e9f0 .

PR #1391, opened 2019-09-08
Update useCallback memoization example in hooks.md

kencoder97:
```
export const CounterComponent = ({ value }) => {
  const incrementCounter = useCallback(
    () => dispatch({ type: 'increment-counter' }),
    [dispatch]
  )
  return (
    <div>
      <span>{value}</span>
      <MyIncrementButton onIncrement={incrementCounter} />
    </div>
  )
}

export const MyIncrementButton = React.memo(({ onIncrement }) => (
  <button onClick={onIncrement}>Increment counter</button>
))
```
In the above code, `dispatch` shouldn't be used as a `useCallback`'s dependency because `dispatch` is different between each render, thus `incrementCounter` will be different too. We should only create and memoize `incrementCounter` once by using an empty array as second argument of `useCallback`


markerikson:
Appreciate the PR, but the current code is correct for two reasons:

- `dispatch` _will_ be the same between renders as long as the same store instance is being passed to `<Provider>`
- Regardless of that, the "rules of hooks" say you should pass in all values used in the hook as dependencies, and the lint plugin will yell at you if you don't pass in `dispatch` here.

Out of curiosity, why did you think that `dispatch` would keep changing?

PR #1394, opened 2019-09-10
Re-enable storeKey

### LINKED ISSUE #1393: connect assumes store prop is a redux store

(If this is a _usage question_, please **do not post it here**—post it on [Stack Overflow](http://stackoverflow.com/questions/tagged/redux) instead. If this is not a “feature” or a “bug”, or the phrase “How do I...?” applies, then it's probably a usage question.)

When you pass a `store` prop to a `connect`ed component, it attempts to use the value of the prop as a redux store. This is wreaking havoc on our codebase that has `store` as a commonly passed prop. I haven't been able to find any documentation for this behavior and I can't find a way to override it, though it does seem to be intentional:

https://github.com/reduxjs/react-redux/blob/master/src/components/connectAdvanced.js#L197

It even seems at one point there was a way to override this behavior: 

https://github.com/reduxjs/react-redux/blob/master/src/components/connectAdvanced.js#L115

I'm filing this as a bug because as far as I can tell, https://github.com/reduxjs/react-redux/pull/1000 removed this behavior and disabled the override key and then https://github.com/reduxjs/react-redux/pull/1209 added it back in, without reenabling the override key. My guess is that this wasn't intentional and hope that adding the override key back would be acceptable.

Repro is obvious, but I can add one if desired.

Either connect would not use the `store` prop at all, or the name would be overridable (likely preferred since it would be a less breaking change).

react-redux 7.1.1
redux 4.0.4
react/react-dom 16.8.3

aaronjensen:
Fixes #1393

### DIFF
--- a/src/components/connectAdvanced.js
+++ b/src/components/connectAdvanced.js
@@ -80,7 +80,6 @@ export default function connectAdvanced(
     // determines whether this HOC subscribes to store changes
     shouldHandleStateChanges = true,
 
-    // REMOVED: the key of props/context to get the store
     storeKey = 'store',
 
     // REMOVED: expose the wrapped component via refs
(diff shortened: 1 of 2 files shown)


markerikson:
Closing per comment in the issue.

Issue #1397, opened 2019-09-12
mapStateToProps is called when component is not due to render when a dispatch is triggered from componentDidUpdate


ykagan:
When a dispatch is triggered from a lifecycle method as a side-effect of a re-render from a previous dispatch, we wind up calling mapStateToProps on components that are never going to be rendered and are about to be unmounted.

https://github.com/reduxjs/react-redux/pull/1263 fixed a very similar issue when multiple dispatches are triggered simultaneously.

Minimal repro is here: https://codesandbox.io/embed/react-redux-issue-debug-rvr-8whlb

ckedar:
The issue is not because of top-down update. This can be verified by overriding the top-down update by providing `store={store}` prop to `App` and `Item` element.

When run with version `5.0.7`, the test case does not throw error, however if we add `console.log` before `invariant` we can see that `mapStateToProps` does throw `invariant` error, just that this error gets swallowed as component does not render subsequently.
 
In `7.1.1`, for some reason, the error caught during `mapStateToProps` is re-thrown during un-mounting, from `unsubscribeWrapper`. This seems to be unnecessary.

karolinkas:
I there a workaround for this?

lukas1994:
Any updates? Having the same issue.

markerikson:
I'm going to close this as a WONTFIX for now, largely because we're encouraging folks to stop using `connect` (even though it'll be fully supported in v8).  This is a real bit of behavior that I can understand is annoying, but I don't think there's a meaningful change we can make at this point, and I don't intend to spend time fixing it myself.

If someone still feels strongly about this please comment, or even better file a PR to improve behavior here.

lfarroco:
My workaround when encountering this kind of issue is wrapping my dispatches with `batch` to have a single re-render.
`await`ing actions to be fully dispatched can also help

PR #1398, opened 2019-09-16
Added test to improve support of top-down update behaviour

arhelmus:
Hey, in recent issue we found test case which violates principle of top-down state update behaviour when state change propagates through mapStateToProps: https://github.com/reduxjs/react-redux/issues/1397.

This PR is adding reproducible test-case to highlight this issue, I believe this is important due to statement in this part of codebase:
```
if (lastThrownError) {
    // It's possible that we caught an error due to a bad mapState function, but the
    // parent re-rendered without this component and we're about to unmount.
    // This shouldn't happen as long as we do top-down subscriptions correctly, but
    // if we ever do those wrong, this throw will surface the error in our tests.
    // In that case, throw the error from here so it doesn't get lost.
    throw lastThrownError
}
```

I would be grateful if you will help me to find how to fix this test case in react-redux or at least merge this test to make sure that issue will be fixed eventually.

### DIFF
--- a/test/components/connect.spec.js
+++ b/test/components/connect.spec.js
@@ -1960,6 +1960,52 @@ describe('React', () => {
       })
     })
 
+    it('should not notify nested components after they are unmounted from componentDidUpdate called dispatch', () => {
+      @connect(state => state)
+      class Parent extends Component {
+        componentDidUpdate() {
+          const { removed } = this.props
+          if (!removed) {
+            store.dispatch({
+              type: 'remove'
+            })
+          }
+        }
+        render() {
+          const { removed } = this.props
+          return <div>{!removed && <Child />}</div>
+        }
+      }
+
+      const mapStateToProps = jest.fn(state => state)
+      @connect(mapStateToProps)
+      class Child extends Component {
+        render() {
+          return <div>{this.props.count}</div>
+        }
+      }
+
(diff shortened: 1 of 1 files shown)


RrNn:
hope this gets merged, so we can get a fix. I've been in a little hell because of this http://bit.ly/2mq7QrK in an earlier version.

markerikson:
If this test case is truly showing a problem with our implementation, then I would really like someone to take some time to investigate the actual _cause_ of the problem.  The failing test case is helpful, but kind of useless on its own.

I unfortunately don't have time to investigate this issue myself right now.

timdorr:
I'm actually not sure what the problem is here. Why do we care if mSTP is called on a removed component? It should be a pure function, so there are no side effects to calling it an extra time. The actual render never happens, which is the important part.

ykagan:
> I'm actually not sure what the problem is here. Why do we care if mSTP is called on a removed component? It should be a pure function, so there are no side effects to calling it an extra time. The actual render never happens, which is the important part.

It becomes a problem when mSTP is written under the assumption that the props passed in are valid props for this component. For example its common to do something like
```
state.items.map((item, i) => <WrappedComponent index={i} />)
```
in the parent component. and
```
const item = state.items[props.i];
```
in the child mSTP. This will cause an exception when item i is removed from items but mSTP is called anyway.

PR #1399, opened 2019-09-16
Add createActionsHook factory method.

tracker1:
I know the `useActions` method was removed for various reasons.  However, this factory method would be incredibly useful for `actions` modules created by developers using `react-redux` in their own applications.

This `createActionsHook` will allow developers to easily expose a `useActions` export from their own actions modules.

### DIFF
--- a/docs/api/hooks.md
+++ b/docs/api/hooks.md
@@ -212,8 +212,48 @@ export const App = () => {
 
 ## Removed: `useActions()`
 
+## `createActionsHook()`
 
+```js
+const useActions : Function = createActionsHook(actions : Object);
+```
 
+This hook factory can be used in with your action creators in order to expose a `useActions()` hook that can be used in your components.
+
+#### Example
+
+```jsx
+// actions.js
+import { createActionsHook } from 'react-redux'
+
+export const doSomething = e => ({  
+  type: 'admin-button-action',
+  payload: e.target.value,
+})
+
+export const useActions = createActionsHook({ doSomething })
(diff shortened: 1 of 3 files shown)


markerikson:
Hi.  Appreciate the PR, but given that we don't export an official `useActions` hook, I don't see a reason to export a factory for one either.  This can be done in userland if you actually want it.  Thanks anyway!

tracker1:
@markerikson the main reason to integrate this, is I've used this method across dozens of action modules... the factory method is used a little differently than useActions would be... it allows for easy creation of `useActions` to be created in the `actions.js`.

for example...
```js
// actions.js
import { createActionsHook } from 'react-redux';

export const doSomething = (e) => ({ type:'do-something', payload: e.target.value });

export const useActions = createActionsHook({ doSomething });

// component.js
import { useActions } from './actions.js'

export default () => {
  const actions = useActions([])
  return (<button value="somevalue" onClick={actions.doSomething)>text</button>)
}
```

This has far more wide reaching utility than `useActions` itself...

markerikson:
There's a few reasons why I don't see a need to add this:

- Per my comment, we aren't officially supporting `useActions()` to begin with
- The "factory" hook here doesn't look like it's actually doing much useful.  You'd get the same behavior if you passed a static object into the `useActions()` recipe we have in the docs.
- There's nothing stopping you from having this in your own code in your own projects.
- There's nothing special about the implementation that requires this to be added as an official API.

PR #1407, opened 2019-09-30
Fix React version check to support 16.10

timdorr:
Since `parseFloat(React.version)` resolves to `16.1`, that breaks our version check. So, I've switched to checking against the `forwardRef` API's existence, which was also added in 16.3.

### DIFF
--- a/src/components/Provider.js
+++ b/src/components/Provider.js
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types'
 import { storeShape, subscriptionShape } from '../utils/PropTypes'
 import warning from '../utils/warning'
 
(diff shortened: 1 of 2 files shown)


samkline:
Note there are still two deprecated lifecycle methods that are causing warnings, both in  `src/components/connectAdvanced.js`: componentWillUpdate and componentWillReceiveProps.

timdorr:
[cWRP is deleted](https://github.com/reduxjs/react-redux/blob/cd67cb30baaa6375c78fd5ed4008436af98d216f/src/components/connectAdvanced.js#L271) and cWU is only in dev mode, so can be safely ignored.

PR #1409, opened 2019-10-01
Avoid warnings in React >=16.3

samkline:
Following up on https://github.com/reduxjs/react-redux/pull/1407, there are two issues still causing warnings on the 5.x branch:

* In development, componentWillUpdate is used. Even though this is dev-only, it causes verbose warnings that can mask other issues when working in a large project.
* componentWillReceiveProps is incorrectly deleted from `Connect`'s static properties, not from the prototype, so the delete is actually a no-op.

This PR addresses both issues. Thanks!


samkline:
I just realized this caused some tests to fail, so I'll close this PR while I investigate it.

PR #1412, opened 2019-10-01
Don't change reference returned by useSelector if equality Fn passes

salazarm:
Don't change the reference returned by useSelector if equality fn passes

Why? (Performance)
Many developers inline their selector functions. The current implementation makes it so that even if they implement the equalityFn to always return true the reference returned by the selector will change because the selector reference changed. This can cause components receiving these objects to re-render if their props aren't primitives! 

To get around this I created a version of useSelector that uses a selector reference that won't change eg:
```
export const useSelectorPerformant = <T, V>(
  selector: (state: T) => V,
  comparator?: (a: V, b: V) => boolean,
): V => {
    const selectorRef = useUpdatingRef(selector);
    const fixedSelector = useRef((rootState: T) => {
      return selectorRef.current(rootState);
    })
    return useSelector<T, V>(fixedSelector.current, comparator);
}

export const useConstantSelectorPerformant = <T, V>(
  selector: (state: T) => V,
  comparator?: (a: V, b: V) => boolean,
): V => {
    const selectorRef = useRef(selector);
    const fixedSelector = useRef((rootState: T) => {
      return selectorRef.current(rootState);
    })
    return useSelector<T, V>(fixedSelector.current, comparator);
}
````

However, Ideally Redux can just run the equalityFn to check if the reference should change.

For reference I tested this behavior by checking if the reference returned by the following selector is changing and in the current implementation it is.
```
  const obj = useSelector(() => ({ someKey: [{id: 1, name: 'test'}] }, () => true)
```

### DIFF
--- a/src/hooks/useSelector.js
+++ b/src/hooks/useSelector.js
@@ -49,6 +49,9 @@ function useSelectorWithStoreAndSubscription(
       latestSubscriptionCallbackError.current
     ) {
       selectedState = selector(store.getState())
+      if (equalityFn(selectedState, latestSelectedState.current)) {
+        selectedState = latestSelectedState.current;
+      }
     } else {
       selectedState = latestSelectedState.current
     }
--- a/src/hooks/useSelector.js
+++ b/src/hooks/useSelector.js
@@ -50,7 +50,7 @@ function useSelectorWithStoreAndSubscription(
     ) {
       selectedState = selector(store.getState())
       if (equalityFn(selectedState, latestSelectedState.current)) {
-        selectedState = latestSelectedState.current;
+        selectedState = latestSelectedState.current
       }
     } else {
       selectedState = latestSelectedState.current

### CODE REVIEW

File: src/hooks/useSelector.js [hunk 1]
```diff
@@ -49,6 +49,9 @@ function useSelectorWithStoreAndSubscription(
       latestSubscriptionCallbackError.current
     ) {
       selectedState = selector(store.getState())
+      if (equalityFn(selectedState, latestSelectedState.current)) {
+        selectedState = latestSelectedState.current;
```
salazarm:
```suggestion
        selectedState = latestSelectedState.current
```

dai-shi:
There might be a hypothetical issue when the equalityFn is changed in the second render.

BTW, I noticed that the equalitfyFn in checkForUpdates is from the stale closure.
It should either read from lastEaualityFn ref or warn(error in DEV) if equalityFn is not stable (not feasible?).

salazarm:
> There might be a hypothetical issue when the equalityFn is changed in the second render.

Do tell. Technically we could get around that with a ref but what issues could arise? if there's not a similar issue with the selector I wonder why we care if the selector reference changes?

salazarm:
I guess I should keep my selectors "pure". I could move calculating these objects to my reducer along with my equality checks and then select them directly from the store. But then that means to use the equality function from redux you need to useCallback your selector which is basically what I did above.

markerikson:
I'm not clear on what the actual problem is that this is trying to fix.

"Inlining a selector" can still return the same reference each time if it's a fairly trivial selector:

```js
const todos = useSelector(state => state.todos);
```

Even though we're defining a new selector function every time, we're still returning the same array reference out of the state (assuming the actions aren't resulting in updates to `state.todos`.

In the example shown, the fact that the selector is "inlined" is irrelevant - it's the fact that _it's returning a new object reference every time_ that's the issue:

```js
 const obj = useSelector(() => ({ someKey: [{id: 1, name: 'test'}] }));
```

Doesn't matter if that function was defined once outside the component or inline, it's always returning a new reference, and so that would force a re-render on every dispatched action.

Do you have any other concrete examples of what you think is actually a problem?  If not, at the moment I'm not seeing a real issue that has to be fixed.

dai-shi:
> other concrete examples

https://codesandbox.io/s/interesting-shtern-nhnkp
Here's a codesandbox of the original issue as far as I understand.
It's not a problem. Just extra re-renders.

> Technically we could get around that with a ref but what issues could arise?

It's hypothetical, but the equalityFn in checkUpdates is from the first render, and the equalityFn in  render is from the latest render, and those can be different. Not just referential difference, but the function itself. For example, one could specify refEqual at the first render (which is default) and then later specify shallowEqual.
So, something like:
```js
  const [myEqual, setMyEqual] = useState(null);
  const person = useSelector(mySelector, myEqual);
  useEffect(() => {
    setMyEqual(shallowEqual);
  }, []);
```
but, this wouldn't be something likely to happen...

salazarm:
> In the example shown, the fact that the selector is "inlined" is irrelevant - it's the fact that it's returning a new object reference every time that's the issue:
```const obj = useSelector(() => ({ someKey: [{id: 1, name: 'test'}] }));```
Doesn't matter if that function was defined once outside the component or inline, it's always returning a new reference, and so that would force a re-render on every dispatched action.


The behavior is different if the selector is inlined vs if its not. When the selector is inlined the reference will change. When the selector is not inlined then the reference does not change if the equality Fn passes.

dai-shi:
(Oh, I didn't notice that part: inlined vs not.)
Here's the modified example with a predefined selector.
https://codesandbox.io/s/muddy-snowflake-wot4k

salazarm:
>  Not just referential difference, but the function itself. For example, one could specify refEqual at the first render (which is default) and then later specify shallowEqual.

I don't see any problem with the equalityFn function changing on the redux side though. Why do we even need to track it? We should just care about whether it returns true or not.

dai-shi:
> Why do we even need to track it?

Because it technically can change. Here's the example. I don't think it happens practically.
https://codesandbox.io/s/pensive-bhabha-nxt3y

salazarm:
Why do we care if it changes though?

markerikson:
@salazarm : I'm saying that the selector function with that hardcoded object is _always_ returning a new object reference, regardless of how the equality function does its comparison.  The equality function is a completely separate point of concern.

salazarm:
>  : I'm saying that the selector function with that hardcoded object is always returning a new object reference, regardless of how the equality function does its comparison.

And I'm saying that is incorrect. The reference does not change when the equality function returns true when the selector is inlined. @dai-shi gave the example showing this behavior right here: https://codesandbox.io/s/pensive-bhabha-nxt3y

markerikson:
@salazarm : you're missing my point.

I'm skipping how `useSelector` behaves.  I'm saying that literally, `() => ({})` always returns a new object.  Period.

salazarm:
@markerikson Actually you missed my point! Obviously that selector function returns a new reference every time. Nobody is arguing that! I'm talking about the reference returned by `useSelector`. Right now the behavior is inconsistent depending on whether your selector is inlined or not and that difference in behavior is what I am trying to address.

markerikson:
Your original statement was:

> Many developers inline their selector functions. The current implementation makes it so that even if they implement the equalityFn to always return true the reference returned by the selector will change because the selector reference changed.

I was attempting to respond to that.

salazarm:
I see. Here is my corrected statement, bolded is the correction:

> Many developers inline their selector functions. The current implementation makes it so that even if they implement the equalityFn to always return true the reference returned by **useS**elector will change because the selector reference changed.

salazarm:
Also FWIW, the equalityFn argument is most useful in the case where we're returning object literals. The other case where it's useful is when the reducer is not properly managing references (changing references when nothing changed, or not changing references when things did change). 

I think if we're supporting an equalityFn then we should support the full use case which to me includes keeping references the same (along with skipping unnecessary renders).

salazarm:
I guess the equality function could be expensive... perhaps its best to just use a memoized selector. We should be clear though that the equality function does not prevent references from changing it only prevents re-renders.

PR #1413, opened 2019-10-01
Docs: Fix Code Example for API `connect()`

duncanleung:
The docs: [Using React Redux > Connect: Extracting Data with mapStateToProps](https://react-redux.js.org/next/using-react-redux/connect-mapstate#the-number-of-declared-arguments-affects-behavior)

Specify the following behavior in the code example:
```js
function mapStateToProps(state) {
  console.log(state) // state
  console.log(arguments[1]) // undefined
}
const mapStateToProps = (state, ownProps = {}) => {
  console.log(state) // state
  console.log(ownProps) // undefined
}
```

These docs for `connect()` show a code example that is not consistent with above example, nor the explanation: [API Reference > connect()](https://react-redux.js.org/next/api/connect#the-arity-of-maptoprops-functions)

> `ownProps` **is not passed** to `mapStateToProps` and `mapDispatchToProps` if the formal definition of the function contains one mandatory parameter (function has length 1). For example, **functions defined like below won't receive `ownProps` as the second argument**

### DIFF
--- a/docs/api/connect.md
+++ b/docs/api/connect.md
@@ -581,7 +581,7 @@ function mapStateToProps(state) {
 
 const mapStateToProps = (state, ownProps = {}) => {
   console.log(state) // state
-  console.log(ownProps) // {}
+  console.log(ownProps) // undefined
 }
 ```
 


timdorr:
Thanks!

markerikson:
I'm pretty sure the prior docs were correct, and this example is wrong.  Here's why.

First, let's see what happens when we define a function with a default argument, and check the `.length`:

```js
const fn1 = (a, b={}) => {}
console.log(fn1.length) // 1
```

Since a `mapState` with a default argument for `ownProps` would have a length of 1, `connect` will indeed call it as `mapState(state)`, not `mapState(state, wrapperProps)`.

That means that the `ownProps` argument will be undefined, and therefore, it would fall back to the empty object default argument that the user actually specified.  

Looking at the diff, the prior example did use to say `console.log(ownProps) // {}`.  That would be correct, and so this needs to be reverted.

Appreciate the attempt to improve things, though!

duncanleung:
@markerikson @timdorr 

Should I make the code example consistent in [Using React Redux > Connect: Extracting Data with mapStateToProps > The Number of Declared Arguments Affects Behavior](https://react-redux.js.org/using-react-redux/connect-mapstate#the-number-of-declared-arguments-affects-behavior)

The same code example is in those docs, but still shows:

```js
function mapStateToProps(state) {
  console.log(state) // state
  console.log(arguments[1]) // undefined
}

const mapStateToProps = (state, ownProps = {}) => {
  console.log(state) // state
  console.log(ownProps) // undefined <- should be // {} ?
}
```

markerikson:
Yeah, maybe tweak that example to clarify it (the incoming value is undefined, which thus results in the default argument value being used).

PR #1415, opened 2019-10-02
Docs: Fix Code Example for Arity of `mapStateToProps`

duncanleung:
Related: #1413

Makes the code example for `mapStateToProps` consistent with [Using React Redux > Connect: Extracting Data with mapStateToProps](https://react-redux.js.org/next/using-react-redux/connect-mapstate#the-number-of-declared-arguments-affects-behavior).

If the incoming value of `ownProps` is undefined, the default argument value will be used.


markerikson:
Thanks!

PR #1418, opened 2019-10-07
import createContext as named import

cristianbote:
By switching to a named import would help with the tree-shake, and also authors could use `react-redux` without the need to export a default `React` as part of the compatibility layer. The functionality does not change.

For example, in the case of `preact-redux` the compatibility layer for bridging `react-redux` with `preact` would not have to use the default export for React, at all. That would result in a better tree-shake result and smaller bundlesize.

Looking forward!

### DIFF
--- a/src/components/Context.js
+++ b/src/components/Context.js
@@ -1,5 +1,5 @@
-import React from 'react'
+import { createContext } from 'react'
 
-export const ReactReduxContext = React.createContext(null)
+export const ReactReduxContext = createContext(null)
 
 export default ReactReduxContext


markerikson:
Do you have any examples of how this would actually improve tree-shaking?  We already import multiple pieces of React throughout the codebase.  I would find it very surprising if changing this one import actually resulted in any improvement.

timdorr:
[Rollup can tree-shake it](https://rollupjs.org/repl/?version=1.23.1&shareable=JTdCJTIybW9kdWxlcyUyMiUzQSU1QiU3QiUyMm5hbWUlMjIlM0ElMjJtYWluLmpzJTIyJTJDJTIyY29kZSUyMiUzQSUyMmltcG9ydCUyMColMjBhcyUyMGFuc3dlciUyMGZyb20lMjAnLiUyRmFuc3dlci5qcyclM0IlNUNuJTVDbmV4cG9ydCUyMGRlZmF1bHQlMjBmdW5jdGlvbiUyMCgpJTIwJTdCJTVDbiU1Q3Rjb25zb2xlLmxvZyglMjAndGhlJTIwYW5zd2VyJTIwaXMlMjAnJTIwJTJCJTIwYW5zd2VyLkZvbyUyMCklM0IlNUNuJTdEJTIyJTJDJTIyaXNFbnRyeSUyMiUzQXRydWUlN0QlMkMlN0IlMjJuYW1lJTIyJTNBJTIyYW5zd2VyLmpzJTIyJTJDJTIyY29kZSUyMiUzQSUyMmV4cG9ydCUyMGNvbnN0JTIwRm9vJTIwJTNEJTIwMjQlM0IlNUNuZXhwb3J0JTIwY29uc3QlMjBCYXIlMjAlM0QlMjA0NCUzQiUyMiUyQyUyMmlzRW50cnklMjIlM0FmYWxzZSU3RCU1RCUyQyUyMm9wdGlvbnMlMjIlM0ElN0IlMjJmb3JtYXQlMjIlM0ElMjJjanMlMjIlMkMlMjJuYW1lJTIyJTNBJTIybXlCdW5kbGUlMjIlMkMlMjJhbWQlMjIlM0ElN0IlMjJpZCUyMiUzQSUyMiUyMiU3RCUyQyUyMmdsb2JhbHMlMjIlM0ElN0IlN0QlN0QlMkMlMjJleGFtcGxlJTIyJTNBbnVsbCU3RA==)

Also, React doesn't do true ES module exports, so it's not going to be traditionally tree-shakable anyways. Not that there's much to shake out anyways, as it's all pretty intertwined internally. Most actual "tree-shaking" is just DCE, which is done by a minifier. And that is all control flow dependent, not ES module based.

So, this isn't going to do much for us. Thanks for looking at it, though!

cristianbote:
Hey @markerikson and @timdorr really appreciate taking time to look into this! 🙏 

@markerikson I started to do the changes and ran the output dist files through a `gzip-size-cli` and I _was_ really optimistic. The savings in this case -- actually I had to change a few more imports 😀 , was `-0.02kB`. So, if `react-redux` would switch to named exports from `react` that would be it. Not _that_ big of a change to make any difference.

@timdorr Indeed. React doesn't do true ES exports. Also, rollup can do tree-shake but I think not in React's case? See [here](https://rollupjs.org/repl/?version=1.23.1&shareable=JTdCJTIybW9kdWxlcyUyMiUzQSU1QiU3QiUyMm5hbWUlMjIlM0ElMjJtYWluLmpzJTIyJTJDJTIyY29kZSUyMiUzQSUyMmltcG9ydCUyMFJlYWN0JTIwZnJvbSUyMCcuJTJGcmVhY3QuanMnJTNCJTVDbiU1Q25leHBvcnQlMjBkZWZhdWx0JTIwZnVuY3Rpb24lMjAoKSUyMCU3QiU1Q24lNUN0Y29uc29sZS5sb2coJTIwJ3RoZSUyMGFuc3dlciUyMGlzJTIwJyUyMCUyQiUyMFJlYWN0LmNyZWF0ZUNvbnRleHQlMjApJTNCJTVDbiU3RCUyMiUyQyUyMmlzRW50cnklMjIlM0F0cnVlJTdEJTJDJTdCJTIybmFtZSUyMiUzQSUyMnJlYWN0LmpzJTIyJTJDJTIyY29kZSUyMiUzQSUyMmV4cG9ydCUyMGNvbnN0JTIwY3JlYXRlRWxlbWVudCUyMCUzRCUyMDI0JTNCJTVDbmV4cG9ydCUyMGNvbnN0JTIwY3JlYXRlQ29udGV4dCUyMCUzRCUyMDQ0JTNCJTVDbiU1Q25jb25zdCUyMFJlYWN0JTIwJTNEJTIwJTdCJTVDbiU1Q3RjcmVhdGVFbGVtZW50JTJDJTVDbiU1Q3RjcmVhdGVDb250ZXh0JTVDbiU3RCUzQiU1Q24lNUNuZXhwb3J0JTIwZGVmYXVsdCUyMFJlYWN0JTNCJTIyJTJDJTIyaXNFbnRyeSUyMiUzQWZhbHNlJTdEJTVEJTJDJTIyb3B0aW9ucyUyMiUzQSU3QiUyMmZvcm1hdCUyMiUzQSUyMmNqcyUyMiUyQyUyMm5hbWUlMjIlM0ElMjJteUJ1bmRsZSUyMiUyQyUyMmFtZCUyMiUzQSU3QiUyMmlkJTIyJTNBJTIyJTIyJTdEJTJDJTIyZ2xvYmFscyUyMiUzQSU3QiU3RCU3RCUyQyUyMmV4YW1wbGUlMjIlM0FudWxsJTdE)

All in all, I think that switching to named exports would add no value at all, because of how `react` does its exports.

Again, wanna thank you for your time and keep up the good work!

PR #1419, opened 2019-10-08
Use the same condition for checking if SSR in useSelector.js as in connectAdvanced.js

LeoDanielsson:
In connectAdvanced.js, window.document.createElement is checked to determine whether the code is running server side. In useSelector only window !== undefined is checked.

Many projects polyfill window on the server, which causes the useLayoutEffect warning when using useSelector. Changing to the more specific condition fixes this.


timdorr:
Thanks!

PR #1425, opened 2019-10-14
Adds react-dom and react-native as optional peer dependencies

arcanis:
This diff adds `react-dom` and `react-native` as **optional** peer dependencies.

The `react-redux` package currently published doesn't list them, which is invalid. For example, here's what Yarn v2 is throwing:

image

Optional peer dependencies are supported by all three package managers and will silence warning that would otherwise appear if a consumer was omitting them.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,16 @@
   },
   "peerDependencies": {
     "react": "^16.8.3",
+    "react-dom": "^16.8.3",
     "redux": "^2.0.0 || ^3.0.0 || ^4.0.0-0"
+  },
+  "peerDependenciesMeta": {
+    "react-dom": {
+      "optional": true
+    },
+    "react-native": {
+      "optional": true
+    }
   },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


timdorr:
See these PRs:

#1261
#1282
#1371
#1366

timdorr:
Whoops, this is actually a dupe of #1390

arcanis:
@timdorr I'm about to merge [a PR](https://github.com/yarnpkg/berry/pull/531/) in Yarn to make peerDependenciesMeta imply peerDependencies (only for Yarn). This should make it possible to remove the warning in both Yarn and npm. Would that be acceptable?

timdorr:
If we can leave it out of peerDeps still, that would be great! 

The concern I have is for those that haven't yet upgraded their client. npm 6.11 has only been bundled with node for a few weeks now and obviously isn't available on an LTS release yet. For those users, adding the peerDep would make it warn on install.

PR #1426, opened 2019-10-14
Remove unused `latestStoreState` field

Hypnosphi:
1. This field doesn't actually affect anything: `action.payload` is always a fresh object anyway
2. Holding a reference to some arbitrary state from the past (namely, the one that triggered the last update for given component) leads to huge memory leaks if the state object is large and there are lots of connected components.


timdorr:
Thanks!

PR #1427, opened 2019-10-14
Fix spelling in hooks error message

ricokahler:
I was writing some tests for a hook that throws inside of a selector and I ran into this little suggestion via my code spell checker:

image

So here we are! 😅


markerikson:
Thanks!

ricokahler:
(well now my test is gonna break haha)

Thanks for the quick feedback

Issue #1428, opened 2019-10-16
Connected component is rendered with stale state


ckedar:
**Do you want to request a _feature_ or report a _bug_?**

Bug

**What is the current behavior?**

Under certain circumstances, a connected component is rendered with stale state.

One such case occurs when you have a stateful `Child` in a `Container`, both of them `connect`ed and child mutates both store and local state at the same time.

**steps to reproduce**

	import React from "react";
	import ReactDOM from "react-dom";
	import { createStore } from "redux";
	import { connect, Provider } from "react-redux";

	function counter(state = 0, action) {
	  return action.type === "INCR" ? state + 1 : state;
	}
	const store = createStore(counter);

	let Child = ({ state, dispatch }) => {
	  console.log("Child.render. state =", state, " store.getState() =", store.getState());
	  if (state !== store.getState()) {
	    window.alert(`Got stale state ${state} !== ${store.getState()}`);
	  }
	  const [, setLocalState] = React.useState(0);
	  const incrHandler = () => {
	    setLocalState(localState => localState + 1);
	    dispatch({ type: "INCR" });
	  };
	  return (
	    <div>
	      Child = {state}
	      <button onClick={incrHandler}>Incr</button>
	    </div>
	  );
	};
	Child = connect(state => {
	  console.log("Child. mapStateToProps");
	  return { state };
	})(Child);

	let Container = ({ state }) => {
	  console.log("Container.render. state =", state);
	  return (
	    <div>
	      Container = {state}
	      <Child />
	    </div>
	  );
	};
	Container = connect(state => ({ state }))(Container);

	function App() {
	  return (
	    <Provider store={store}>
	      <Container />
	    </Provider>
	  );
	}

	const rootElement = document.getElementById("root");
	ReactDOM.render(<App />, rootElement);

https://codesandbox.io/s/suspicious-merkle-0kzcg

**What is the expected behavior?**

*  Component should never be rendered with stale state.
*  Also, on state and store mutation, the component should be rendered only once.

**Versions **

```
  "dependencies": {
    "react": "16.8.6",
    "react-dom": "16.8.6",
    "react-redux": "7.1.1",
    "redux": "4.0.4"
  },

```

PR #1429, opened 2019-10-16
Component should always be rendered with latest state

ckedar:
Added test to detect component being rendered with stale state.
Issue: https://github.com/reduxjs/react-redux/issues/1428

### DIFF
--- a/src/components/connectAdvanced.js
+++ b/src/components/connectAdvanced.js
@@ -410,7 +410,10 @@ export default function connectAdvanced(
     }
 
     // If we're in "pure" mode, ensure our wrapper component only re-renders when incoming props have changed.
-    const Connect = pure ? React.memo(ConnectFunction) : ConnectFunction
+    // const Connect = pure ? React.memo(ConnectFunction) : ConnectFunction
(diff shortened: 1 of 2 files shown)


src/components/connectAdvanced.js
```diff
@@ -410,7 +410,10 @@ export default function connectAdvanced(
     }
 
     // If we're in "pure" mode, ensure our wrapper component only re-renders when incoming props have changed.
-    const Connect = pure ? React.memo(ConnectFunction) : ConnectFunction
+    // const Connect = pure ? React.memo(ConnectFunction) : ConnectFunction
```
markerikson: This change is completely incorrect and needs to be reverted.

test/components/connect.spec.js
```diff
@@ -3304,4 +3304,47 @@ describe('React', () => {
       expect(thrownError).toBe(null)
     })
   })
+  it('component should always be rendered with latest state', () => {
```
markerikson: What specifically does this test add that our existing tests don't cover?

PR #1439, opened 2019-10-30
WIP: Static Types doc page

JNaftali:
Static types doc page - see https://github.com/reduxjs/react-redux/issues/1001#issuecomment-546744689

### DIFF
--- a/docs/using-react-redux/static-types.md
+++ b/docs/using-react-redux/static-types.md
@@ -1,0 +1,106 @@
+---
+id: static-types
+title: Type safety and React-Redux
+hide_title: true
+sidebar_label: Type safety and React-Redux
+---
+
+# Type safety and React-Redux
+
+The [`react-redux` type definitions](https://npm.im/@types/react-redux) export some helpers to make it easier to write typesafe interfaces between your redux store and your React components.
+
+## Typing the useSelector hook
+
+If you manually type your selector functions, `useSelector` will return the same type as your selector
+
+```ts
+// selectIsOn always returns a boolean, so isOn is typed as a boolean
+const isOn = useSelector(selectIsOn)
+```
+
+Otherwise `useSelector` requires you to manually type the state argument every time you use it.
+
+```ts
+const isOn = useSelector((state: MyStateType) => state.isOn)
+```
+
+If you'd like, you can define a typed `useSelect` hook using a helper type
+
+```ts
+// reducer.ts
+import { useSelector, TypedUseSelectorHook } from 'react-redux'
+
+interface MyStateType {
+  isOn: boolean
+}
+
+export const useTypedSelector: TypedUseSelectorHook<MyStateType> = useSelector
+
+// my-component.tsx
+import { useTypedSelector } from './reducer.ts'
+
+const isOn = useSelector(state => state.isOn)
+```
+
+## Typing the `connect` higher order component
+
+React-redux exposes a helper type, `ConnectedProps`, that can extract the return type of mapStateToProps and mapDispatchToProps.
+
+```ts
+import { connect, ConnectedProps } from 'react-redux'
+
+interface MyStateType {
+  isOn: boolean
+}
+
+const connector = connect(
+  (state: MyStateType) => ({
+    isOn: state.isOn
+  }),
+  (dispatch: any) => ({
+    toggleOn: () => dispatch({ type: 'TOGGLE_IS_ON' })
+  })
+)
+
+type PropsFromRedux = ConnectedProps<typeof connector>
+```
+
+The return type of `ConnectedProps` can then be used to type your props object.
+
+```tsx
+interface Props extends PropsFromRedux {
+  backgroundColor: string
+}
+
+export const MyComponent = connector((props: Props) => (
+  <div style={{ backgroundColor: props.backgroundColor }}>
+    <button onClick={props.toggleOn}>
+      Toggle is {props.isOn ? 'ON' : 'OFF'}
+    </button>
+  </div>
+))
+```
+
+If you are using untyped selectors or for some other reason would rather, you can also pass type arguments to `connect` and use those same types to create your own props type.
+
+```ts
+interface StateProps {
+  isOn: boolean
+}
+
+interface DispatchProps {
+  toggleOn: () => void
+}
+
+interface OwnProps {
+  backgroundColor: string
+}
+
+type Props = StateProps & DispatchProps & OwnProps
+
+const connector = connect<
+  StateProps,
+  DispatchProps,
+  OwnProps
+>(/* arguments as above */)
+```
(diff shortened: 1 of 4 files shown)


JNaftali:
Explanations definitely need fleshing out and polishing. Uploading as a WIP to get feedback. Any feedback would be appreciated but primarily looking for:
(a) Document structure - do I need more sections? Should the existing ones be in a different order?
(b) Examples - Are they detailed enough/cover enough use cases? Are they clear enough?

JNaftali:
Also: couldn't figure out how to get the page working with docusaurus, so I didn't do anything in that area 😬 .

Next thing I'm gonna do is read #1336 - if that's getting finished up part of me wants to wait for that to be merged in and then integrate what I'm writing into that.

docs/using-react-redux/static-types.md
```diff
@@ -0,0 +1,106 @@
...
+
+// my-component.tsx
+import { useTypedSelector } from './reducer.ts'
+
+const isOn = useSelector(state => state.isOn)
+```
+
+## Typing the `connect` higher order component
+
+React-redux exposes a helper type, `ConnectedProps`, that can extract the return type of mapStateToProps and mapDispatchToProps.
```
markerikson: I'd like some explanatory stuff here about how `connect` is a HOC, and there's three separate sources of props: parent component, `mapState`, and `mapDispatch`.  

We should show a couple examples of "manually" typing `connect` and the plain component before introducing the `ConnectedProps<T>` technique.

Then again, I could see an argument for doing those _after_, or maybe even skipping them if we want to just highlight this technique.

docs/using-react-redux/static-types.md
```diff
@@ -0,0 +1,106 @@
...
+
+interface MyStateType {
+  isOn: boolean
+}
+
+const connector = connect(
+  (state: MyStateType) => ({
+    isOn: state.isOn
+  }),
+  (dispatch: any) => ({
```
markerikson: I'd prefer to use the "object shorthand" here for `mapDispatch` vs a function.

docs/using-react-redux/static-types.md
```diff
@@ -0,0 +1,106 @@
...
+React-redux exposes a helper type, `ConnectedProps`, that can extract the return type of mapStateToProps and mapDispatchToProps.
+
+```ts
+import { connect, ConnectedProps } from 'react-redux'
+
+interface MyStateType {
+  isOn: boolean
+}
+
+const connector = connect(
```
markerikson: I'd want to highlight that normally we do the `connect(mapState, mapDispatch)(MyComponent)` double-function-call all in one step at the end, but for this to work we break it up into two steps (and particularly doing the first part _before_ the component definition) so we can infer the props from the `connector` function.

docs/using-react-redux/static-types.md
```diff
@@ -0,0 +1,106 @@
...
+  })
+)
+
+type PropsFromRedux = ConnectedProps<typeof connector>
+```
+
+The return type of `ConnectedProps` can then be used to type your props object.
+
+```tsx
+interface Props extends PropsFromRedux {
```
markerikson: Y'know, I keep forgetting you can extend interfaces.

docs/using-react-redux/static-types.md
```diff
@@ -0,0 +1,106 @@
...
+  backgroundColor: string
+}
+
+export const MyComponent = connector((props: Props) => (
```
markerikson: Let's separate out the component definition just for clarity.

docs/using-react-redux/static-types.md
```diff
@@ -0,0 +1,106 @@
...
+  <div style={{ backgroundColor: props.backgroundColor }}>
+    <button onClick={props.toggleOn}>
+      Toggle is {props.isOn ? 'ON' : 'OFF'}
+    </button>
+  </div>
+))
+```
+
+If you are using untyped selectors or for some other reason would rather, you can also pass type arguments to `connect` and use those same types to create your own props type.
```
markerikson: "would rather".... what?

docs/using-react-redux/static-types.md
```diff
@@ -0,0 +1,106 @@
...
+  toggleOn: () => void
+}
+
+interface OwnProps {
+  backgroundColor: string
+}
+
+type Props = StateProps & DispatchProps & OwnProps
+
+const connector = connect<
```
markerikson: Okay, so this shows manually typing `connect`, but we should show using `Props` explicitly for the component when it's declared separately.  (Feel free to omit the actual rendering logic here.)

markerikson:
Paging @wgao19 .  Thoughts on the existing content?  Can you do something similar for the Flow part?

docs/using-react-redux/static-types.md
```diff
@@ -0,0 +1,106 @@
+---
+id: static-types
+title: Type safety and React-Redux
+hide_title: true
+sidebar_label: Type safety and React-Redux
+---
+
+# Type safety and React-Redux
+
+The [`react-redux` type definitions](https://npm.im/@types/react-redux) export some helpers to make it easier to write typesafe interfaces between your redux store and your React components.
```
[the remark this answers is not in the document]
JNaftali: You don't think the `TypedUseSelectorHook` type counts?

docs/using-react-redux/static-types.md
JNaftali: Would you rather I say `type Props = PropsFromRedux & {`? I've a slight preference for types, but the codebase I work on mostly uses interfaces. I've no idea which would be easier to understand

docs/using-react-redux/static-types.md
JNaftali: "would rather not rephrase an awkwardly worded sentence"

JNaftali:
Revised some things according to your feedback, @markerikson. In addition to the little things, I swapped the order of the last section around and added a bit about type hoisting. Not sure if that bits necesssary

docs/using-react-redux/static-types.md
markerikson: That's the syntax I've been using, but not sure if it matters that much.

docs/using-react-redux/static-types.md
JNaftali: Tried flipping it around so that the example of manually typing connect comes first. Not sure what meaningful variations on this that I should cover.

markerikson:
@wgao19 @endiliey @yangshun The only way I can get this to show up in the main sidebar is if I explicitly put the page into `versioned-docs/v7.1`, and use a specific versioned ID.  Any idea why just adding it to `/docs` and `sidebar.json` doesn't get it to show up in the v7.1 version?

I have to say that I find the versioning stuff very confusing.

markerikson:
In any case, I'm happy with the way this looks now, so I'll merge it.  I'd still like to see if there's a better way to add new pages with the versioning, though.

markerikson:
Thanks to @JNaftali for getting this put together!

PR #1444, opened 2019-11-05
Fix RN batching and effect behavior

### LINKED ISSUE #1313: Store change not immediately visible to component
This code used to work in react-redux v5.0.7:

1. componentDidUpdate dispatch action, then call reducer to update store
2. reducer updates store 
3. componentDidUpdate called again with new store value from selector
 
Step 3 no longer works in react-redux ^6.0.1:
selector not called

in 7.1.0, selector is called but the value not picked up by component props

### LINKED ISSUE #1437: Dispatch more than once in non-batched update cause incorrect result with `useSelector`

markerikson:
Fixes #1313 and fixes #1437 (hopefully!).

Per https://github.com/facebook/react/issues/14927 , `useLayoutEffect()` warns in SSR.  We tried to feature-detect that by checking if `window` exists, and if so, falling back to `useEffect`:

https://github.com/reduxjs/react-redux/blob/dbfcc97720cc962c8468f48d06a280d1dbf4d832/src/components/connectAdvanced.js#L40-L41

In #1283 , that was found to be insufficient if someone polyfilled `window`, so the check was made more explicit:

https://github.com/reduxjs/react-redux/blob/dcf2cb0864fffa58d914146cd08f2402ef66b1a9/src/components/connectAdvanced.js#L40-L45

We then duplicated that over into `useSelector` as well.

Unfortunately, it looks like that feature detection breaks in React Native, and we've been falling back to `useEffect` the whole time.  This has led to subtle timing bugs popping up, and we (I) ignored the issue reports largely because we couldn't reproduce them in a web environment.  In addition, while several folks said that something had broken with `connect` between 7.0.3 and 7.1.0, I only looked at the one tiny change inside of `connect` itself, not at the `useIsomorphicLayouteffect` change, and didn't catch that all the reports were in RN projects.

This PR hopefully fixes all those, by:

- Adding a `useIsomorphicLayoutEffect.native.js` file that always uses `useLayoutEffect` with RN
- Adding a test suite for RN-specific batching behavior
- Asserting that `useIsomorphicLayoutEffect === useLayoutEffect` in an RN environment
- Specifically using the repro app from #1437 as a test case to verify this works right


markerikson:
@MingruiZhang : what renderer are you using?

markerikson:
Okay, let's go with this.

aleclarson:
I just published `react-layout-effect` with your solution, so it can be shared between libraries.

https://www.npmjs.com/package/react-layout-effect

I'll gladly add you as a maintainer if you want to use it in `react-redux`.

PR #1450, opened 2019-11-06
fix performance createListenerCollection

leshkovichpvl:
createListenerCollection has a bad performance for more listeners.
In my implementation, performance is much better

before: 
Screen Shot 2019-11-06 at 18 13 44

after:
Screen Shot 2019-11-06 at 18 18 26

### DIFF
--- a/src/utils/Subscription.js
+++ b/src/utils/Subscription.js
@@ -7,18 +7,16 @@ import { getBatch } from './batch'
 const nullListeners = { notify() {} }
 
 function createListenerCollection() {
-  var batch = getBatch()
-  // the current/next pattern is copied from redux's createStore code.
-  // TODO: refactor+expose that code to be reusable here?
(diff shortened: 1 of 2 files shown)


timdorr:
Can you provide some numbers on how this performs better?

leshkovichpvl:
> Can you provide some numbers on how this performs better?
I added screenshots to make it clear.

timdorr:
OK, I cleaned things up a bit.

We used the previous current/next form copied from Redux core because we wanted to ensure adding a subscriber while dispatching wouldn't trigger an edge case. However, since this code is nested inside of Subscription anyways, we actually don't need these guarantees.

Thanks for the PR!

leshkovichpvl:
I'm sorry, I made a mistake. This PR must be revert. Because it led to the wrong order of calling parent/child selectors.😟

Issue #1351, opened 2019-07-08
Concurrent Mode


brunolemos:
**Do you want to request a _feature_ or report a _bug_?**

Feature

**What is the expected behavior?**

Full compatibility with Concurrent Mode

**Details**

We should be close to a Concurrent Mode release, and people are already experimenting with it on their projects. I am not sure but it seems react-redux is not compatible yet, based on [this tweet](https://twitter.com/dai_shi/status/1148233182547857410) and [this pr](https://github.com/reduxjs/react-redux/pull/1350).

markerikson:
There's still not ETA on when Concurrent Mode will be released, and there's also a general lack of clarity on what exactly it will include, what the impacts are, and what the points of incompatibility may be.  

As [Dan said in issue #1164](https://github.com/reduxjs/react-redux/issues/1164#issuecomment-457196712):

> We need to be clear that there are several “levels” of compatibility with new features of React. They’re not formalized anywhere yet but a rough sketch for a library or technique X could be:

> - X breaks in sync mode
> - X works in sync mode but breaks in concurrent mode
> - X works in concurrent mode but limits its DX and UX benefits for the whole app
> - X works in concurrent mode but limits its UX benefits for the whole app
> - X works in concurrent mode but limits its DX and UX benefits for features written in X
> - X works in concurrent mode but limits its UX benefits for features written in X
> - X works in concurrent mode and lets its users take full advantage of its benefits

> This is not a strict progression and there’s a spectrum of tradeoffs. (For example, maybe there is some temporary visual inconsistencies such as different like counts, but no crashes are guaranteed.)

> But we need to be more precise about where React Redux is, and where it aims to be.


Until Concurrent Mode is actually out and completely documented, we can really only speculate on the potential interactions, and there's a lot of other stuff that I'd rather spend my time worrying about.

I'll leave this open as a placeholder, but it's not an immediate priority.

dai-shi:
I don't mean any rush, but while it's not an immediate priority, let's do some experiment.
I made a small repo for reproduction of "tearing" (as far as I understand).
https://github.com/dai-shi/will-this-react-global-state-work-in-concurrent-mode

It would be nice if someone could review it as I might be biased.
Please file issues there about this tool or any related discussion, not to disturb this issue here.

gcloeval:
Hello redux team,

[...]

PR #1455, opened 2019-11-08
fix CM tearing 

salvoravida:
fix https://github.com/reduxjs/react-redux/issues/1351

>  Hi folks.
> My idea is to change forceRender into useSelector hook, let instead set a state, and return that state.
> This should transform redux state changes into react useState changes that should batched and eventually paused form react concurrent mode

### DIFF
--- a/src/components/connectAdvanced.js
+++ b/src/components/connectAdvanced.js
@@ -1,6 +1,6 @@
 import hoistStatics from 'hoist-non-react-statics'
 import invariant from 'invariant'
-import React, { useContext, useMemo, useRef, useReducer } from 'react'
+import React, { useContext, useMemo, useRef, useState } from 'react'
 import { isValidElementType, isContextConsumer } from 'react-is'
 import Subscription from '../utils/Subscription'
 import { useIsomorphicLayoutEffect } from '../utils/useIsomorphicLayoutEffect'
@@ -8,7 +8,6 @@ import { useIsomorphicLayoutEffect } from '../utils/useIsomorphicLayoutEffect'
 import { ReactReduxContext } from './Context'
 
 // Define some constant arrays just to avoid re-creating these
-const EMPTY_ARRAY = []
 const NO_SUBSCRIPTION_ARRAY = [null, null]
 
 const stringifyComponent = Comp => {
@@ -18,13 +17,6 @@ const stringifyComponent = Comp => {
     return String(Comp)
   }
 }
-
-function storeStateUpdatesReducer(state, action) {
-  const [, updateCount] = state
-  return [action.payload, updateCount + 1]
-}
-
-const initStateUpdates = () => [null, 0]
 
 export default function connectAdvanced(
   /*
@@ -232,9 +224,9 @@ export default function connectAdvanced(
       // We need to force this wrapper component to re-render whenever a Redux store update
       // causes a change to the calculated child component props (or we caught an error in mapState)
       const [
-        [previousStateUpdateResult],
+        previousStateUpdateResult,
         forceComponentUpdateDispatch
-      ] = useReducer(storeStateUpdatesReducer, EMPTY_ARRAY, initStateUpdates)
+      ] = useState({ reduxStore: store.getState() })
 
       // Propagate any mapState/mapDispatch errors upwards
       if (previousStateUpdateResult && previousStateUpdateResult.error) {
@@ -246,6 +238,8 @@ export default function connectAdvanced(
       const lastWrapperProps = useRef(wrapperProps)
       const childPropsFromStoreUpdate = useRef()
       const renderIsScheduled = useRef(false)
+
+      const latestReduxStore = useRef()
 
       const actualChildProps = usePureOnlyMemo(() => {
         // Tricky logic here:
@@ -265,8 +259,15 @@ export default function connectAdvanced(
         // This will likely cause Bad Things (TM) to happen in Concurrent Mode.
         // Note that we do this because on renders _not_ caused by store updates, we need the latest store state
         // to determine what the child props should be.
-        return childPropsSelector(store.getState(), wrapperProps)
-      }, [store, previousStateUpdateResult, wrapperProps])
+        return childPropsSelector(
+          latestReduxStore.current || previousStateUpdateResult.reduxStore,
+          wrapperProps
+        )
+      }, [
+        latestReduxStore.current,
+        previousStateUpdateResult.reduxStore,
+        wrapperProps
+      ])
 
       // We need this to execute synchronously every time we re-render. However, React warns
       // about useLayoutEffect in SSR, so we try to detect environment and fall back to
@@ -322,6 +323,8 @@ export default function connectAdvanced(
 
           // If the child props haven't changed, nothing to do here - cascade the subscription update
           if (newChildProps === lastChildProps.current) {
+            latestReduxStore.current = latestStoreState
+
             if (!renderIsScheduled.current) {
               notifyNestedSubs()
             }
@@ -335,11 +338,10 @@ export default function connectAdvanced(
             renderIsScheduled.current = true
 
             // If the child props _did_ change (or we caught an error), this wrapper component needs to re-render
+            latestReduxStore.current = undefined
             forceComponentUpdateDispatch({
-              type: 'STORE_UPDATED',
-              payload: {
-                error
-              }
+              error,
+              reduxStore: latestStoreState
             })
           }
         }
--- a/src/components/connectAdvanced.js
+++ b/src/components/connectAdvanced.js
@@ -239,9 +239,6 @@ export default function connectAdvanced(
       const childPropsFromStoreUpdate = useRef()
       const renderIsScheduled = useRef(false)
 
-      const latestReduxState = useRef()
-      const latestReduxStateIsValid = useRef(false)
-
       const actualChildProps = usePureOnlyMemo(() => {
         // Tricky logic here:
         // - This render may have been triggered by a Redux store update that produced new child props
@@ -257,16 +254,10 @@ export default function connectAdvanced(
         }
 
         return childPropsSelector(
-          latestReduxStateIsValid.current
-            ? latestReduxState.current
-            : previousStateUpdateResult.reduxState,
+          previousStateUpdateResult.reduxState,
           wrapperProps
         )
-      }, [
-        latestReduxStateIsValid.current,
-        previousStateUpdateResult.reduxState,
-        wrapperProps
-      ])
+      }, [previousStateUpdateResult.reduxState, wrapperProps])
 
       // We need this to execute synchronously every time we re-render. However, React warns
       // about useLayoutEffect in SSR, so we try to detect environment and fall back to
@@ -322,8 +313,11 @@ export default function connectAdvanced(
 
           // If the child props haven't changed, nothing to do here - cascade the subscription update
           if (newChildProps === lastChildProps.current) {
-            latestReduxState.current = latestStoreState
-            latestReduxStateIsValid.current = true
+            forceComponentUpdateDispatch(prev => {
+              prev.error = error
+              prev.reduxState = latestStoreState
+              return prev
+            })
 
             if (!renderIsScheduled.current) {
               notifyNestedSubs()
@@ -338,8 +332,6 @@ export default function connectAdvanced(
             renderIsScheduled.current = true
 
             // If the child props _did_ change (or we caught an error), this wrapper component needs to re-render
-            latestReduxStateIsValid.current = false
-            latestReduxState.current = undefined
             forceComponentUpdateDispatch({
               error,
               reduxState: latestStoreState
--- a/src/components/connectAdvanced.js
+++ b/src/components/connectAdvanced.js
@@ -226,7 +226,7 @@ export default function connectAdvanced(
       const [
         previousStateUpdateResult,
         forceComponentUpdateDispatch
-      ] = useState({ reduxStore: store.getState() })
+      ] = useState({ reduxState: store.getState() })
 
       // Propagate any mapState/mapDispatch errors upwards
       if (previousStateUpdateResult && previousStateUpdateResult.error) {
@@ -239,7 +239,8 @@ export default function connectAdvanced(
       const childPropsFromStoreUpdate = useRef()
       const renderIsScheduled = useRef(false)
 
-      const latestReduxStore = useRef()
+      const latestReduxState = useRef()
+      const latestReduxStateIsValid = useRef(false)
 
       const actualChildProps = usePureOnlyMemo(() => {
         // Tricky logic here:
@@ -255,17 +256,15 @@ export default function connectAdvanced(
           return childPropsFromStoreUpdate.current
         }
 
-        // TODO We're reading the store directly in render() here. Bad idea?
-        // This will likely cause Bad Things (TM) to happen in Concurrent Mode.
-        // Note that we do this because on renders _not_ caused by store updates, we need the latest store state
-        // to determine what the child props should be.
         return childPropsSelector(
-          latestReduxStore.current || previousStateUpdateResult.reduxStore,
+          latestReduxStateIsValid.current
+            ? latestReduxState.current
+            : previousStateUpdateResult.reduxState,
           wrapperProps
         )
       }, [
-        latestReduxStore.current,
-        previousStateUpdateResult.reduxStore,
+        latestReduxStateIsValid.current,
+        previousStateUpdateResult.reduxState,
         wrapperProps
       ])
 
@@ -323,7 +322,8 @@ export default function connectAdvanced(
 
           // If the child props haven't changed, nothing to do here - cascade the subscription update
           if (newChildProps === lastChildProps.current) {
-            latestReduxStore.current = latestStoreState
+            latestReduxState.current = latestStoreState
+            latestReduxStateIsValid.current = true
 
             if (!renderIsScheduled.current) {
               notifyNestedSubs()
@@ -338,10 +338,11 @@ export default function connectAdvanced(
             renderIsScheduled.current = true
 
             // If the child props _did_ change (or we caught an error), this wrapper component needs to re-render
-            latestReduxStore.current = undefined
+            latestReduxStateIsValid.current = false
+            latestReduxState.current = undefined
             forceComponentUpdateDispatch({
               error,
-              reduxStore: latestStoreState
+              reduxState: latestStoreState
             })
           }
         }
--- a/src/hooks/useSelector.js
+++ b/src/hooks/useSelector.js
@@ -13,9 +13,7 @@ function useSelectorWithStoreAndSubscription(
   store,
   contextSub
 ) {
-  const [reduxState, forceRender] = useState(store.getState())
-  const latestReduxState = useRef()
-  const latestReduxStateIsValid = useRef(false)
+  const [reduxState, setReduxState] = useState({ value: store.getState() })
 
   const subscription = useMemo(() => new Subscription(store, contextSub), [
     store,
@@ -33,9 +31,7 @@ function useSelectorWithStoreAndSubscription(
       selector !== latestSelector.current ||
       latestSubscriptionCallbackError.current
     ) {
-      selectedState = selector(
-        latestReduxStateIsValid.current ? latestReduxState.current : reduxState
-      )
+      selectedState = selector(reduxState.value)
     } else {
       selectedState = latestSelectedState.current
     }
@@ -62,8 +58,10 @@ function useSelectorWithStoreAndSubscription(
         const newSelectedState = latestSelector.current(newReduxState)
 
         if (equalityFn(newSelectedState, latestSelectedState.current)) {
-          latestReduxState.current = newReduxState
-          latestReduxStateIsValid.current = true
+          setReduxState(prev => {
+            prev.value = newReduxState
+            return prev
+          })
           return
         }
 
@@ -75,9 +73,7 @@ function useSelectorWithStoreAndSubscription(
         // changed
         latestSubscriptionCallbackError.current = err
       }
-      latestReduxStateIsValid.current = false
-      latestReduxState.current = undefined
-      forceRender(newReduxState)
+      setReduxState(() => ({ value: newReduxState }))
     }
 
     subscription.onStateChange = checkForUpdates
--- a/src/hooks/useSelector.js
+++ b/src/hooks/useSelector.js
@@ -14,6 +14,7 @@ function useSelectorWithStoreAndSubscription(
   contextSub
 ) {
   const [reduxState, forceRender] = useState(store.getState())
+  const latestReduxState = useRef()
 
   const subscription = useMemo(() => new Subscription(store, contextSub), [
     store,
@@ -31,7 +32,7 @@ function useSelectorWithStoreAndSubscription(
       selector !== latestSelector.current ||
       latestSubscriptionCallbackError.current
     ) {
-      selectedState = selector(reduxState)
+      selectedState = selector(latestReduxState.current || reduxState)
     } else {
       selectedState = latestSelectedState.current
     }
@@ -58,6 +59,7 @@ function useSelectorWithStoreAndSubscription(
         const newSelectedState = latestSelector.current(newReduxState)
 
         if (equalityFn(newSelectedState, latestSelectedState.current)) {
+          latestReduxState.current = newReduxState
           return
         }
 
@@ -69,7 +71,7 @@ function useSelectorWithStoreAndSubscription(
         // changed
         latestSubscriptionCallbackError.current = err
       }
-
+      latestReduxState.current = undefined
       forceRender(newReduxState)
     }
 
--- a/src/hooks/useSelector.js
+++ b/src/hooks/useSelector.js
@@ -1,4 +1,4 @@
-import { useReducer, useRef, useMemo, useContext } from 'react'
+import { useRef, useMemo, useContext, useState } from 'react'
 import invariant from 'invariant'
 import { useReduxContext as useDefaultReduxContext } from './useReduxContext'
 import Subscription from '../utils/Subscription'
@@ -13,7 +13,7 @@ function useSelectorWithStoreAndSubscription(
   store,
   contextSub
 ) {
-  const [, forceRender] = useReducer(s => s + 1, 0)
+  const [reduxState, forceRender] = useState(store.getState())
 
   const subscription = useMemo(() => new Subscription(store, contextSub), [
     store,
@@ -31,7 +31,7 @@ function useSelectorWithStoreAndSubscription(
       selector !== latestSelector.current ||
       latestSubscriptionCallbackError.current
     ) {
-      selectedState = selector(store.getState())
+      selectedState = selector(reduxState)
     } else {
       selectedState = latestSelectedState.current
     }
@@ -53,8 +53,9 @@ function useSelectorWithStoreAndSubscription(
 
   useIsomorphicLayoutEffect(() => {
     function checkForUpdates() {
+      const newReduxState = store.getState()
       try {
-        const newSelectedState = latestSelector.current(store.getState())
+        const newSelectedState = latestSelector.current(newReduxState)
 
         if (equalityFn(newSelectedState, latestSelectedState.current)) {
           return
@@ -69,7 +70,7 @@ function useSelectorWithStoreAndSubscription(
         latestSubscriptionCallbackError.current = err
       }
 
-      forceRender({})
+      forceRender(newReduxState)
     }
 
     subscription.onStateChange = checkForUpdates
--- a/src/hooks/useSelector.js
+++ b/src/hooks/useSelector.js
@@ -15,6 +15,7 @@ function useSelectorWithStoreAndSubscription(
 ) {
   const [reduxState, forceRender] = useState(store.getState())
   const latestReduxState = useRef()
+  const latestReduxStateIsValid = useRef(false)
 
   const subscription = useMemo(() => new Subscription(store, contextSub), [
     store,
@@ -32,7 +33,9 @@ function useSelectorWithStoreAndSubscription(
       selector !== latestSelector.current ||
       latestSubscriptionCallbackError.current
     ) {
-      selectedState = selector(latestReduxState.current || reduxState)
+      selectedState = selector(
+        latestReduxStateIsValid.current ? latestReduxState.current : reduxState
+      )
     } else {
       selectedState = latestSelectedState.current
     }
@@ -60,6 +63,7 @@ function useSelectorWithStoreAndSubscription(
 
         if (equalityFn(newSelectedState, latestSelectedState.current)) {
           latestReduxState.current = newReduxState
+          latestReduxStateIsValid.current = true
           return
         }
 
@@ -71,6 +75,7 @@ function useSelectorWithStoreAndSubscription(
         // changed
         latestSubscriptionCallbackError.current = err
       }
+      latestReduxStateIsValid.current = false
       latestReduxState.current = undefined
       forceRender(newReduxState)
     }
--- a/test/components/Provider.spec.js
+++ b/test/components/Provider.spec.js
@@ -246,13 +246,13 @@ describe('React', () => {
         store.dispatch({ type: 'APPEND', body: 'c' })
       })
 
-      expect(childMapStateInvokes).toBe(2)
-      expect(childCalls).toEqual([['a', 'a'], ['ac', 'ac']])
+      expect(childMapStateInvokes).toBe(3)
+      expect(childCalls).toEqual([['a', 'a'], ['a', 'ac'], ['ac', 'ac']])
 
       // setState calls DOM handlers are batched
       const button = tester.getByText('change')
       rtl.fireEvent.click(button)
-      expect(childMapStateInvokes).toBe(3)
+      expect(childMapStateInvokes).toBe(5)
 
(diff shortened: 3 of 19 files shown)


salvoravida:
this should fix this demo
https://github.com/reduxjs/react-redux/issues/1351#issuecomment-509893599

dai-shi:
This is probably a similar approach to my old implementation of reactive-react-redux.

It is important to keep consistency during intermediate updates, so you want to keep the test unchanged.
I think it's doable if you take the same approach as use-subscription.

It would actually be interesting to develop user-subscription based react-redux,
although I'm not motivated because I would like to push [my approach](https://github.com/dai-shi/use-context-selector).
I think the redux team is waiting for [the core support](https://github.com/reactjs/rfcs/pull/119), so they might not be so interested with the approach with use-subscription. But, anyway forking and experimenting should be good.

salvoravida:
finally it fix even useTransition and isPending

https://codesandbox.io/s/test-redux-in-concurrent-mode-with-react-redux-pytbh

salvoravida:
> This is probably a similar approach to my old implementation of reactive-react-redux.
> 
> It is important to keep consistency during intermediate updates, so you want to keep the test unchanged.
> I think it's doable if you take the same approach as use-subscription.
> 
> It would actually be interesting to develop user-subscription based react-redux,
> although I'm not motivated because I would like to push [my approach](https://github.com/dai-shi/use-context-selector).
> I think the redux team is waiting for [the core support](https://github.com/reactjs/rfcs/pull/119), so they might not be so interested with the approach with use-subscription. But, anyway forking and experimenting should be good.

the only test that i fix, is the one where the parent force the same value to the child, that need 2 render to CORRECTLY process all store updates.  reading store on render is an hack to have 1 render less on that "rarely usecase" but that break the normal queque from Redux to useState.

eddyw:
See https://github.com/reduxjs/react-redux/issues/1351#issuecomment-551802697
If component is wrapped within `React.memo`, it doesn't work 😛

src/hooks/useSelector.js
```diff
@@ -31,7 +31,7 @@ function useSelectorWithStoreAndSubscription(
       selector !== latestSelector.current ||
       latestSubscriptionCallbackError.current
```
eddyw: To be CM friendly, could we do something like:
```javascript
try {
 ...
} catch(e) {
  if (typeof e === 'object' && e !== null && typeof e.then === 'function') throw e
  let errorMessage = ...
```
So we could do this, for example:
```javascript
  const status = useSelector(s => {
    if (s.status === STATUS.PENDING) throw promise;
    return s.status;
  })
}
```
In React docs, it's using something like:
```const value = resource.read()```
where `read()` either returns the value or throws so it can bail out before the selector is set. Not sure if here is the best place to do it, but yeah, with CM not everything "thrown" is an error.

Comment (salvoravida) [reply src/hooks/useSelector.js]: sorry i think that thow promise to comunicate with Suspense does NOT matter with useSelector, that has to select a state without tearing on render. that's another problem.

**Comment by salvoravida** [comment]
> See [#1351 (comment)](https://github.com/reduxjs/react-redux/issues/1351#issuecomment-551802697)
> If component is wrapped within `React.memo`, it doesn't work 😛
here is the fix https://codesandbox.io/s/test-redux-in-concurrent-mode-with-react-redux-salvoravidareact-redux-9jo9z

but please note that is a bug of the current "experimental react" with "SimpleMemoComponent"
infact force memo Non simple it works.

rreact internal

```jsx
function updateMemoComponent(current$$1, workInProgress, Component, nextProps, updateExpirationTime, renderExpirationTime) {
  if (current$$1 === null) {
    var type = Component.type;

    if (isSimpleFunctionComponent(type) && Component.compare === null && // SimpleMemoComponent codepath doesn't resolve outer props either.
    Component.defaultProps === undefined) {
      var resolvedType = type;

      {
        resolvedType = resolveFunctionForHotReloading(type);
      } // If this is a plain function component without default props,
      // and with only the default shallow comparison, we upgrade it
      // to a SimpleMemoComponent to allow fast path updates.


      workInProgress.tag = SimpleMemoComponent;
      workInProgress.type = resolvedType;

      {
        validateFunctionComponentInDev(workInProgress, type);
      }

      return updateSimpleMemoComponent(current$$1, workInProgress, resolvedType, nextProps, updateExpirationTime, renderExpirationTime);
    }
```

i will open a issue on React.

So this "issue" is not related to useSelector!

markerikson:
I'm not actually sure what this PR is trying to accomplish.  Can you clarify?

Also, whatever it's doing only seems to involve `useSelector`, and not `connect`.

salvoravida:
> I'm not actually sure what this PR is trying to accomplish. Can you clarify?
> 
> Also, whatever it's doing only seems to involve `useSelector`, and not `connect`.

1) https://github.com/dai-shi/will-this-react-global-state-work-in-concurrent-mode
tearing when dispatching fast action and slow updates, please @dai-shi can you add your thought?

2) fix support dispatch action within "startTransaction"
here demo ok : https://codesandbox.io/s/test-redux-in-concurrent-mode-with-react-redux-pytbh
here demo KO : https://codesandbox.io/s/test-redux-in-concurrent-mode-with-react-redux-crbfn
isPending is not working

salvoravida:
the problem was reading store on render phase.

markerikson:
The point of `forceRender` is that the Redux store notification process is occurring outside of React, and therefore React has no way to know that a re-render is needed.  Therefore, we need to _tell_ React that it should queue up a re-render, and we are literally "forcing a render".

Now, the exact mechanics of that may be up for discussion.

For _any_ potential change, I would specifically like to see some unit tests that fail under the existing implementation, and pass under a new implementation.

salvoravida:
> The point of `forceRender` is that the Redux store notification process is occurring outside of React, and therefore React has no way to know that a re-render is needed. Therefore, we need to _tell_ React that it should queue up a re-render, and we are literally "forcing a render".
> 
> Now, the exact mechanics of that may be up for discussion.
> 
> For _any_ potential change, I would specifically like to see some unit tests that fail under the existing implementation, and pass under a new implementation.

yes i know. the only change is that reduxStore.getState() must be done ONLY on useEffect and non on render otherwise there is the tearing effect when react render 50 items in a random way.

anyway, ok i will update with 2 test that simulates the 2 demos

markerikson:
Sure.  The other half of this is that we need an equivalent implementation for `connect()`, as well.

dai-shi:
>  I'm not sure if we ever actually added the useSelector implementations to our benchmarks repo.

@markerikson I did. Please check out the PR and I can adjust if you have suggestions.

dai-shi:
@salvoravida In my experience, if you read store in render, you need to deal with the case that the store state is change from the time when checkForUpdates is called. That's what use-subscription is dealing with.

salvoravida:
> Sure. The other half of this is that we need an equivalent implementation for `connect()`, as well.

done, fixed connect too.
Hope i have some time on weekend to add the CM tests.

Meanwhile if someone would like to try, here is a temp build : @salvoravida/react-redux 7.1.6 🎉

src/components/connectAdvanced.js
```diff
@@ -322,6 +323,8 @@ export default function connectAdvanced(
 
           // If the child props haven't changed, nothing to do here - cascade the subscription update
           if (newChildProps === lastChildProps.current) {
+            latestReduxStore.current = latestStoreState
```
markerikson: This line doesn't look right. We shouldn't be assigning refs while rendering - that's definitely _not_ safe for Concurrent Mode.

test/hooks/useSelector.spec.js
```diff
@@ -350,13 +350,11 @@ describe('React', () => {
...
 
-          let sawInconsistentState = false
+          let lastRenderInconsistentState = false
 
           const Child = ({ parentCount }) => {
             const result = useSelector(({ count }) => {
-              if (count !== parentCount) {
-                sawInconsistentState = true
-              }
+              lastRenderInconsistentState = count !== parentCount
```
MrWolfZ: I don't understand this change. The whole purpose of this test was to ensure the selector never sees inconsistent state, not just on the latest render.

test/components/Provider.spec.js
```diff
@@ -246,13 +246,13 @@ describe('React', () => {
         store.dispatch({ type: 'APPEND', body: 'c' })
       })
 
-      expect(childMapStateInvokes).toBe(2)
-      expect(childCalls).toEqual([['a', 'a'], ['ac', 'ac']])
+      expect(childMapStateInvokes).toBe(3)
+      expect(childCalls).toEqual([['a', 'a'], ['a', 'ac'], ['ac', 'ac']])
```
MrWolfZ: this changes the whole point of the test. this test ensures that `mapState` never sees inconsistent state. with this change you are asserting the exact opposite

src/hooks/useSelector.js
```diff
@@ -68,8 +71,8 @@ function useSelectorWithStoreAndSubscription(
         // changed
         latestSubscriptionCallbackError.current = err
       }
-
-      forceRender({})
+      latestReduxState.current = undefined
+      forceRender(newReduxState)
```
MrWolfZ: I'm not sure this achieves what you want. The state that you set here is only used when the selector changes on the next render or an error was thrown above. In all other cases it is still `latestSelectedState.current` that gets used for rendering. And `latestSelectedState.current` is not safe in CM, since it may have been updated while a render that depends on an old value is suspended.

MrWolfZ: When I implemented this originally, I experimented with putting `latestSelectedState` in a `useState` instead of using a ref, but I didn't go through with that approach for a reason I can't remember sadly.

Looking at it now with fresh eyes I think the solution may be to store both, the latest full store state as well as the latest selected state in a `useState` and update each depending on whether calling the selector succeeded or not. This way there should never be a reason to access a state ref during rendering.

MrWolfZ: All of that said, I believe that the CM issues are more fundamental. Basically, any code that uses refs is very likely to be susceptible to tearing or other issues. In this hook for example, imagine someone changes the selector. However, the render that changes the selector is suspended. During suspension a store update is processed. Since `checkForUpdates` uses the selector from the latest committed render, it will use a stale selector for processing the update, which may return in missed updates.

MrWolfZ: The core problems remain the same, i.e. that store updates and rendering happen in two different "logical threads" so to speak that are prone to unfortunate interleaving. I believe the only proper way to fix this is to use the context to propagate the store value, which will hopefully become feasible once context selectors are implemented.

dai-shi: Let me put a link how I did approach to the CM issue (at least one of issues). https://github.com/dai-shi/reactive-react-redux/issues/29

src/components/connectAdvanced.js
salvoravida: Hi, that assignment is NOT on the render phase, is on the subscription Callback.
We have 2 case when store changes
1) the change produce a selector change -> we enqueque a component render with useState.setState(newReduxStore)
2) the change DOES not produce a selector change -> we "save" the new reduxState, because we have to always remember the last "reduxState event callback change" if there is an update from others reason like prop changes.

src/hooks/useSelector.js
salvoravida: Hi, that assignment is NOT on the render phase, is on the subscription Callback.
We have 2 case when store changes

the change produce a selector change -> we enqueque a component render with useState.setState(newReduxStore)
the change DOES not produce a selector change -> we "save" the new reduxState, because we have to always remember the last "reduxState event callback change" if there is an update from others reason like prop changes.

test/components/Provider.spec.js
salvoravida: Because we NO more read the last state in the render phase, the "parent-child-issue" is resolved with 2 render, as every change of the store make 2 event-callback one on the parent and one on the child.
Before the issue was resolved with the "hack" of reading always the fresh Store state, even if we are on the "preview" scheduled refresh came out from "checkForUpdate" callback.

That hack while resolve the "parent-issue" with 1 render less, broke the changes normal queue on the the componente, that produce tearing on CM where every component could rendered in a random way, jumping the "changhes queue"

please try the demo of @dai-shi with my version and with old one. you will see the differences on a fast store changes, with slow render.

src/hooks/useSelector.js
salvoravida: > The core problems remain the same, i.e. that store updates and rendering happen in two different "logical threads" so to speak that are prone to unfortunate interleaving. I believe the only proper way to fix this is to use the context to propagate the store value, which will hopefully become feasible once context selectors are implemented.

that will come back to v6.x performance issue. hum.

test/hooks/useSelector.spec.js
salvoravida: Because we NO more read the last state in the render phase, the "parent-child-issue" is resolved with 2 render, as every change of the store make 2 event-callback one on the parent and one on the child.
Before the issue was resolved with the "hack" of reading always the fresh Store state, even if we are on the "preview" scheduled refresh came out from "checkForUpdate" callback.

That hack while resolve the "parent-issue" with 1 render less, broke the changes normal queue on the the componente, that produce tearing on CM where every component could rendered in a random way, jumping the "changhes queue"

salvoravida: please try the demo of @dai-shi with my version and with old one. you will see the differences on a fast store changes, with slow render.

test/components/Provider.spec.js
dai-shi: If you are developing your own fork, it's fine and I think it's one possible approach.
But, if you would like to support all existing RR apps (which is necessary for the official RR), you want to keep the backward compatibility. As I said, it should be technically possible. Seems like you are used to the code base, I think you can do it. I'm not 100% sure though.

salvoravida: @dai-shi i'm not dev my own fork, otherwise i havent opened a pr. that said i think that was a mistake reading fresh store state on render, to have 1 less render on that particular use case. Moreover is of couse back compatible with 1 render more.
Anyway i will try to explain better with a flow chart as soon as i have time. That's all.

salvoravida: The Exactly point is: when a componet render it MUST read the storeState that was from HIS last storeChangeCallback!. Otherwise reading in render store.getState() could be newer than HIS callback queque. THAT'S THE POINT. Think more about that. test the demo.

src/components/connectAdvanced.js
markerikson: My bad, sorry.

test/components/Provider.spec.js
markerikson: Can you clarify what you mean by "more renders"?  It sounds like you're saying we will cause additional render cycles to occur, which is going to be bad for performance.

salvoravida:
> @salvoravida In my experience, if you read store in render, you need to deal with the case that the store state is change from the time when checkForUpdates is called. That's what use-subscription is dealing with.

this pr does not read store on render, but on callback handler.

dai-shi:
> this pr does not read store on render, but on callback handler.

Sorry, that's my misunderstanding from your original comment. I will try to understand again how it's working.

Have you come up with any idea to keep the test unchanged?

salvoravida:
Hi folks, i have added dai-shi CM tests on /test/cm !
Good News ! @dai-shi update your site with react-redux :D
`
PASS  __tests__/all_spec.js (12.341s)
  react-redux
    √ check1: updated properly (8053ms)
    √ check2: no tearing during update (1ms)
    √ check3: ability to interrupt render
    √ check4: proper update after interrupt (2108ms)

Test Suites: 1 passed, 1 total
Tests:       4 passed, 4 total
`

dai-shi:
Please calm down.
Reading useSelector.js carefully, I think I understand what you are trying to do.
It will be more CM friendly than the current RR, if it could be done right.
To my knowledge, we can't use useRef to keep single source of truth.
I could probably make a failing test, but it would be hard and I wouldn't dare to do it.

So, there are two issues in my mind. One is the changed test case, and the other is the useRef. I'm not sure but they might be from the single issue.

Based on your idea, I would think something like this might work.
```js
const useSelector = (selector) => {
  const [selected, setSelected] = useState(selector(store.getState()));
  if (selected !== selector(store.getState())) {
    setSelected(selector(store.getState()));
  }
  ...
};
```
Hmm, I'm not sure if this solves the first issue...
(I'm not even sure if the same technique can be applied to `connect`)

You want to try this? It seems like you are so exited (which is good), and I'm not sure I can convince you.

src/hooks/useSelector.js
```diff
@@ -31,7 +32,7 @@ function useSelectorWithStoreAndSubscription(
       selector !== latestSelector.current ||
       latestSubscriptionCallbackError.current
     ) {
-      selectedState = selector(store.getState())
+      selectedState = selector(latestReduxState.current || reduxState)
```
eddyw: Is this safe?
Imagine store state is a.. string:
```const reducer = (state = '', action) => { ... }```
If it's a falsy value `0 || null || undefined || ''`, then it'd always use `reduxState` and not `latestReduxState.current`. In most cases, this shouldn't really happen since it's unusual to have a store that holds a single value but still ? 😛 (same in connect)

Comment (markerikson) [reply src/hooks/useSelector.js]: Yeah, this seems like a bug.

Comment (salvoravida) [reply src/hooks/useSelector.js]: yeah fixed https://github.com/reduxjs/react-redux/pull/1455/commits/eb6f8cf1742a6a319dac4a459e5513a1e6e50c7e

File: src/components/connectAdvanced.js (hunk 2 above)
Comment (salvoravida) [reply src/components/connectAdvanced.js]: this should be resolved, if you agree.

**Comment by salvoravida** [comment]
> Please calm down.
> Reading useSelector.js carefully, I think I understand what you are trying to do.
> It will be more CM friendly than the current RR, if it could be done right.
> To my knowledge, we can't use useRef to keep single source of truth.
> I could probably make a failing test, but it would be hard and I wouldn't dare to do it.
> 
> So, there are two issues in my mind. One is the changed test case, and the other is the useRef. I'm not sure but they might be from the single issue.
> 
> Based on your idea, I would think something like this might work.
> 
> ```js
> const useSelector = (selector) => {
>   const [selected, setSelected] = useState(selector(store.getState()));
>   if (selected !== selector(store.getState())) {
>     setSelected(selector(store.getState()));
>   }
>   ...
> };
> ```
> 
> Hmm, I'm not sure if this solves the first issue...
> (I'm not even sure if the same technique can be applied to `connect`)
> 
> You want to try this? It seems like you are so exited (which is good), and I'm not sure I can convince you.



> Please calm down.
> Reading useSelector.js carefully, I think I understand what you are trying to do.
> It will be more CM friendly than the current RR, if it could be done right.
> To my knowledge, we can't use useRef to keep single source of truth.
> I could probably make a failing test, but it would be hard and I wouldn't dare to do it.
> 
> So, there are two issues in my mind. One is the changed test case, and the other is the useRef. I'm not sure but they might be from the single issue.
> 
> Based on your idea, I would think something like this might work.
> 
> ```js
> const useSelector = (selector) => {
>   const [selected, setSelected] = useState(selector(store.getState()));
>   if (selected !== selector(store.getState())) {
>     setSelected(selector(store.getState()));
>   }
>   ...
> };
> ```
> 
> Hmm, I'm not sure if this solves the first issue...
> (I'm not even sure if the same technique can be applied to `connect`)
> 
> You want to try this? It seems like you are so exited (which is good), and I'm not sure I can convince you.

There are so many points that we should discuss and find all an agreement, step by step, without mixing different points.

i will try to start with these::

1) First of all, we should all agree that we CANNOT read store.getState() into render phase, because that could be "reading from the future" in CM when there are many slow-render-component, dispatches with LOW priority and Scheduler interrupt ROOT_RENDER and commit partial renders.
That is "TEARING" effect. 

2) Secondo point is that storeChanges MUST be enquequed to useState with setState, so that every render can read reduxState from *hisCallback and his queue*

this fix also the others @dai-shi CM tests.

This is also why i do NOT think that your last example of useSelector can work in CM!
-

**Comment by dai-shi** [comment]
I understand we are not on the same page.
- [First Goal] My goal is to *at least* make it work in CM. That is to make my CM tests (excluding check3) pass without changing `useSelector.spec.js`.
- [Future Goal] Your goal is far ahead of it, and try to support state branching. That would be challenging and we might not be able to keep backward compatibility.

> we should all agree that we CANNOT read store.getState() into render phase

I disagree if I stand for [First Goal]. For [Future Goal], yes.

**Comment by salvoravida** [comment]
> I understand we are not on the same page.
> 
> * [First Goal] My goal is to _at least_ make it work in CM. That is to make my CM tests (excluding check3) pass without changing `useSelector.spec.js`.
> * [Future Goal] Your goal is far ahead of it, and try to support state branching. That would be challenging and we might not be able to keep backward compatibility.
> 
> > we should all agree that we CANNOT read store.getState() into render phase
> 
> I disagree if I stand for [First Goal]. For [Future Goal], yes.



> I understand we are not on the same page.
> 
> * [First Goal] My goal is to _at least_ make it work in CM. That is to make my CM tests (excluding check3) pass without changing `useSelector.spec.js`.
> * [Future Goal] Your goal is far ahead of it, and try to support state branching. That would be challenging and we might not be able to keep backward compatibility.
> 
> > we should all agree that we CANNOT read store.getState() into render phase
> 
> I disagree if I stand for [First Goal]. For [Future Goal], yes.

you cannot have the [FG] CM safe (no tearing) while you want to read on render phase!

the test that i have changhed is safe, was an hack before to force read store on render phase.
to explain better this, i will prepare new tests. just wait! :)

**Comment by dai-shi** [comment]
> you cannot have the [FG] CM safe (no tearing) while you want to read on render phase!

Both are [FG]... I assume [First Goal]. I think we can based on my snippet (which is from use-subscription which is basically for external store for CM safe).

**Comment by eddyw** [comment]
@dai-shi 
> [Future Goal] Your goal is far ahead of it, and try to support state branching. That would be challenging and we might not be able to keep backward compatibility.

I don't think it's trying to do that. There will always be state branching as long as it uses `useState` or `useReducer`, and it'll happen whenever you set a value within a transition. In this case, dispatch an action within a transition. However, that's a completely unrelated issue to "fixing tearing"

If you think in `useContextSelector`, if you set the value passed to the provider within a transition, it will have this state branching as well.

I'm trying to describe this in the other thread. Also, it'd be backwards compatible of course, because existing RR apps just don't use `useTransition` and `Suspense` also supports `React.lazy`

**Comment by dai-shi** [comment]
@eddyw Thanks for the clarification. I agree fixing tearing is nothing to do with branching.
And I think it can be solved by reading store state in render and fallback to re-render if it's changed during renders. (BTW, your description of the research in the other thread is very interesting.)

**Comment by salvoravida** [comment]
fixed new @dai-shi CM tests. now reduxState is 100% "streamed" from Redux to Components, without reading from the future or jumping queue callbacks!  CM should be fine.

File: src/hooks/useSelector.js [hunk 7]
```diff
@@ -53,10 +53,15 @@ function useSelectorWithStoreAndSubscription(
...
   useIsomorphicLayoutEffect(() => {
     function checkForUpdates() {
+      const newReduxState = store.getState()
       try {
-        const newSelectedState = latestSelector.current(store.getState())
+        const newSelectedState = latestSelector.current(newReduxState)
 
         if (equalityFn(newSelectedState, latestSelectedState.current)) {
+          setReduxState(prev => {
+            prev.value = newReduxState
```
dai-shi: @salvoravida As I said in other thread, I'm not confident this works for the future. Do you ever feel like just trying my previous snippet???

salvoravida: For what i have understand, useRef is not CM safe. its a shared mem between branches. we need an useStateRef, a way to save data to current branch states queue.
I think it is safe, just says React to not re-render, but store a value on that branch.

i dont' think that reading value from the future is ok in CM and event streaming, but i will try your idea asap. 
Do you have tried?

markerikson: This line is mutating the existing state value, though, which is definitely _not_ CM-safe.

salvoravida: no, because it is executed only on next update, before next render.

salvoravida: on every branch, so every branch has it's own reduxState queue changes. that's why there is NO more tearing, even if with last test, that simulate this edge case.

markerikson: Mmm... I'm going to need to see some examples and take some time to understand why this may be a good idea.  Why would we not want to just spread the state object here?

markerikson: Also, while I don't know how all the internals of React are implemented, what makes you say that every branch has a separate state object?  Can you point to where you're seeing that in the React source?

salvoravida: becouse we do not want a re-render as reduxState changes, but selectedState NO.

salvoravida: but we also want to Stream "all" redux changes to iternaal component states (that is handled bu react) without missing ones, or without reading from the future, because of that.

salvoravida: i think that we no more need nested subscriptions. "100%" queque of reduxChanges streamed to components, means that react can "branches" states when need, every comp has his own queue

markerikson: I'm really not convinced this is a technically correct approach.  I'll have to look at it more later.

markerikson: Also, nesting subscriptions is specifically for ensuring that a component renders with the latest props from its parent, it doesn't have anything to do with CM or Suspense.  That's still necessary.

salvoravida: instead i think that the component should not "jump" renders for that case, but process his queue of reduxChanges that will fix slate pros on next render. (but only 1 commit due to batches)

this way we avoid reading from future that is bad in CM. 
But yeah, do some tests by your own .)

dai-shi: @salvoravida One note: As I understand, by mutating state in callback that redux can call at any time, what you are doing now is equivalent to reading store.getState(). So, I don't think you avoid reading from future. (If I understand "read from future" correctly.)

salvoravida: No. That's like Redis queue FIFO. On ReduxState Changes, we "capture the state" (note where we read the state) and enqueue to the component, that will evaluete it at any time, but "inqueue".

Note2: useState.setState === useReducer.Dispatch Action. read ReactInternals :D

salvoravida: In few words: we "broadcast" reduxChanges to React branches. OneToMany.

dai-shi: That was my bad. It's not updating it in callback, but in the updater function.

test/components/Provider.spec.js
MrWolfZ: @markerikson I believe this is not due to an additional render. let me try to outline the order of events here:

1. the store updates
2. the parent sees the change first (due to tiered subscription) and renders
3. since the parent rendered and changed the props, the child renders as well. with this change, during the render the child will see the last state it knows about, which is the previous store state (since due to tiered subscription the child didn't see the update yet). funnily enough this is actually the inverse of the stale props issue, since now we actually see newer parent props with an old state instead of old props with a new state
4. once the render of the parent finishes the child finally sees the state update and re-renders itself with consistent props and state

it feels to me that tiered subscription and CM might just be incompatible

@salvoravida as @dai-shi pointed out, this discussion is not about the correctness of your change, but about preserving the existing behavior for backwards compatibility. there are likely tons of apps out there that have `mapStateToProps` written in a way that depends on the props and state being consistent. that's why it is important we think of a way to keep the existing behavior. in the `useSelector` hook we took the approach of simply ignoring these situations and dealing with errors by forcing a re-render. that was a deliberate decision we could make because it was a new API. `connect` is sadly different

src/hooks/useSelector.js
eddyw: I don't understand what this does 🙈 
I removed this and:
```javascript
          setReduxState(prev => {
            prev.value = newReduxState
            return prev
          })
```
I tried it locally, still passed the tests. Am I missing something?

At first I though you were using it as a custom `ref` that can optionally cause a re-render but I'm confused by this:
```
In few words: we "broadcast" reduxChanges to React branches. OneToMany, without lost "frames" !
```
Maybe `reduxState` can be "branched", but the selected state is still a ref. Is that the point?

salvoravida: Have you run CM tests? on /test/cm. on SyncRender there is 1 queue only so it's irrilevant.

Every Branch has it's own States queue. so states are Branch safe, Ref are shared beetween renders braches! the are tecnicaly "this" on the component instances.
States are Copied and enqueued on every branched that has different speed and timeline of execution! so the are branch safe.

salvoravida: this is what i call statesRef, here are an implementation.
```jsx
import { useState, useCallback } from 'react';

export const useStateRef = initialState => {
  const [state, setState] = useState(() => ({
    current: typeof initialState === 'function' ? initialState() : initialState
  }));

  const setStateRef = useCallback(
    (setter, forceRender) => {
      setState(prev => {
        const current = typeof setter === 'function' ? setter(prev.current) : setter;
        if (forceRender) return { current };
        prev.current = current;
        return prev;
      });
    },
    [setState]
  );
  return [state, setStateRef];
};
```
As CM becomes the standard, i wish React will have a native useStateRef.
It will help libraries stay in sync with React branches.
Maybe i should open an RFC?

eddyw: Yes, I removed those 3 lines, run build, then jest:
```javascript
 PASS  __tests__/all_spec.js (13.458s)
  react-redux
    ✓ check1: updated properly (8836ms)
    ✓ check2: no tearing during update (2ms)
    ✓ check3: ability to interrupt render
    ✓ check4: proper update after interrupt (2406ms)
```
That's why it's difficult to reason about.

salvoravida: have you update to last @dai-shi tests? i do not think so. get a pull :D

salvoravida: https://github.com/salvoravida/react-redux/commits/fix-use-selector-cm

salvoravida: @dai-shi can confirm?

eddyw: `Already up to date.`
I cloned the repo just about 30min ago, so I'm sure it's up to date 😅

salvoravida: i will check. there should be something wrong on copying daishi test from his project. that fix was for his new test with.

salvoravida: https://github.com/dai-shi/will-this-react-global-state-work-in-concurrent-mode/issues/3#issuecomment-552300031

eddyw: hm, so it's more or less like...:
```javascript
const [refState, forceUpdateRef] = React.useState({ current: null })
React.useImperativeHandle(refState, () => 'some value', [ ...optional_deps ])
```
Isn't this just like a short cut?, I mean, if keeping state in ref and just using `forceUpdate`. If it's within a transition, `forceUpdate` will be scheduled for later but the ref will still change which I think this does. Using `useState` for refs is just like merging a ref with forceUpdate into one line of code:
```javascript
const ref = useRef()
const forceUpdate = useReducer(c => c + 1, 0)[1]
somewhere {
   ref.current = someValue
   forceUpdate()
}
...
const [ref, forceUpdate] = useState({ current: null })
somewhere {
   ref.current = someValue
   forceUpdate({ ...ref })
}
```
Both `forceUpdate` will schedule "for later" if within transition.

salvoravida: Yes but using ref, another rendering brahch could update the same value before or after, like shared mem between 3d. Instead states ref are duplicated on every branch and every state ref updater will work on its own branch rendering.

eddyw: Idk, this is how I see it:
```
queue no transition => [
  setReduxState(p => { p.value = reduxState 1; return p }),
  setReduxState(p => { p.value = reduxState 2; return p }), // before p.value is reduxState 1
  setReduxState(p => { p.value = reduxState 3; return p }), // before p.value is reduxState 2
]
start transition queue => [
  later setReduxState(p => { p.value = reduxState 4; return p }), // previous 'p' is unknown
]
queue outside transition => [
  setReduxState(p => { p.value = reduxState 5; return p }), // before p.value is reduxState 3
]
stop transition queue => [
  run previous:
    initial state: reduxState 5 <<<< this is previous 'p' now
    setReduxState(p => { p.value = reduxState 4; return p }), // <<< inconsistent state
]
```
Now, current @dai-shi tests using transition are not exactly right. It's testing dispatching in transition.. but, since there is no Suspense, no component is waiting for anything, so there is no need to queue for later.

Sometimes there may be some flickering on screen, when you use transition but nothing causes interrupting of rendering (like throw promise on Suspense), I've seen on some demos I was testing with, I think this is a bug related to what Dan mentioned on twitter, maybe? Point is, you **can't** test using transition without causing interrupt on render (throw on Suspense), otherwise it is pointless to wrap it in `startTransition`.

dai-shi: I'm not following the discussion in deep, but:

> without causing interrupt on render (throw on Suspense)

My understanding is a bit different. We can cause interrupt with heavy computation without throwing promise. That's the whole point of my tests.

P.S. Did I ever test using transition? Just today, I made new checks using `useTransition` based on @eddyw 's demo. No Suspense. No throwing promises. Is it pointless?

eddyw: Never mind, ignore the above 🙈 
I've copied the tests over this PR locally which did have `useTransition` but you finished committing later, so I had them broken

markerikson:
This PR is stale, and it's unlikely we're going to make any CM-specific changes to v7 at this point. Instead, we're more likely to rewrite things into a v8 release that builds on top of `useMutableSource` somehow.

PR #1466, opened 2019-11-14
Convert footer.js to function component

thiskevinwang:
# Description

The (maybe by now) classic _stateless class - to - function component_ switcherino.

(Please close if not a priority!)


timdorr:
Thanks!

Issue #1470, opened 2019-11-22
Investigate tree-shaking


markerikson:
As hooks continue to gain adoption, it's possible that some folks might want to write apps that _only_ use our hooks API.  I'm unsure whether `connect` actually tree-shakes or not.  

It would be helpful if someone could do some testing to see if `connect` tree-shakes when only the hooks are used, and if it doesn't, figure out why not and what changes are needed to enable that.  (On the flip side, I'd also be curious if the hooks tree-shake when you only use `connect`.)

Andarist:
https://github.com/reduxjs/react-redux/blob/118601605244f3e4a1f5a0b953faa863c8a623f7/src/index.js#L14

Cannot be tree-shaken (unless a particular minifier goes deeper and figures out that this is needed only if set `batch` stays referenced), but it is expected. Would be cool to figure out how to do this lazily but it's a super minor thing.

https://github.com/reduxjs/react-redux/blob/118601605244f3e4a1f5a0b953faa863c8a623f7/src/utils/shallowEqual.js#L1

Top-level getter also cannot be tree-shaken, arguably this is a minor thing as well - but ideally, you should inline this into the function.

https://github.com/reduxjs/react-redux/blob/118601605244f3e4a1f5a0b953faa863c8a623f7/src/utils/useIsomorphicLayoutEffect.js#L12

Similarly this - also a short leftover, but cannot be removed. A recommendation here would be making it a function, and put `/* #__PURE__ */` right before the call:
```
const isHopefullyDomEnvironment = () =>
  typeof window !== 'undefined' &&
  typeof window.document !== 'undefined' &&
  typeof window.document.createElement !== 'undefined'

export const useIsomorphicLayoutEffect = /* #__PURE__ */ isHopefullyDomEnvironment()
  ? useLayoutEffect
  : useEffect
```

https://github.com/reduxjs/react-redux/blob/118601605244f3e4a1f5a0b953faa863c8a623f7/src/hooks/useDispatch.js#L41

Top-level function call prevents this from being tree-shaked. The recommendation would be a PURE annotation before the call.

https://github.com/reduxjs/react-redux/blob/118601605244f3e4a1f5a0b953faa863c8a623f7/src/hooks/useSelector.js#L2

Most likely this is not written in ESM and this might prevent tree-shaking it. The recommendation would be to just migrate to using if+throw explicitly without somewhat quirky API wrapper - the React team is doing the same thing from what I read (migrating away from this pattern).

[...]

PR #1472, opened 2019-11-23
Remove invariant in favor of doing NODE_ENV checks directly

Andarist:
Relates to https://github.com/reduxjs/react-redux/issues/1470

Removes tiny (but distributes solely in CJS format and thus deoptimizing) dependency. It also follows suit of https://github.com/facebook/react/issues/16753 - this should be more approachable to newcomers, because `invariant` is a little bit quirky - in example I often found myself wondering if I should use passing or failing condition when working with various OSS projects using this pattern. This change makes it easier to immediately know how this works.

This only helps webpack as Rollup is smart enough to remove this dependency on its own. Results on my playground:
```diff

Rolluped:
    3330
Rolluped & uglified:
    2388
Webpacked:
-   3294
+   3071

```


timdorr:
Thanks!

PR #1473, opened 2019-11-23
Pass @babel/runtime version to @babel/plugin-transform-runtime

Andarist:
It doesn't change anything right now, but it's a future-proof change that will allow `@babel/plugin-transform-runtime` to add imports to `@babel/runtime` helpers based on currently used version of `@babel/runtime` package - instead of falling back to list of helpers available in initial Babel 7 release.

### DIFF
--- a/.babelrc.js
+++ b/.babelrc.js
@@ -9,6 +9,14 @@ module.exports = {
     ['@babel/proposal-object-rest-spread', { loose }],
     '@babel/transform-react-jsx',
     cjs && ['@babel/transform-modules-commonjs', { loose }],
-    ['@babel/transform-runtime', { useESModules: !cjs }],
-  ].filter(Boolean),
+    [
+      '@babel/transform-runtime',
+      {
+        useESModules: !cjs,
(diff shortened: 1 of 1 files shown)


.babelrc.js
```diff
@@ -9,6 +9,14 @@ module.exports = {
     ['@babel/proposal-object-rest-spread', { loose }],
     '@babel/transform-react-jsx',
     cjs && ['@babel/transform-modules-commonjs', { loose }],
-    ['@babel/transform-runtime', { useESModules: !cjs }],
-  ].filter(Boolean),
+    [
+      '@babel/transform-runtime',
+      {
+        useESModules: !cjs,
+        version: require('../package.json').dependencies[
```
timdorr: This is at the root, so it shouldn't be looking up a directory.

Andarist: right 😅this is how last-second changes end up 😉 just fixed it

timdorr:
While I kind of wish the plugin did this itself, this seems like a good stopgap.

PR #1474, opened 2019-11-24
Rethrowing original errors thrown in selector functions

### LINKED ISSUE #1433: Do not alter exception thrown from selector
bug

useSelector extracts `.message` field from thrown exception in https://github.com/reduxjs/react-redux/blob/master/src/hooks/useSelector.js#L60 hiding original exception with stack trace making it hard to pinpoint exact error location.

Is it possible to re-throw original error or, at least print it using console.error?

r3dm1ke:
When throwing error from selector function, append original error stacktrace (if present) to new error. Should fix #1433

### DIFF
--- a/src/hooks/useSelector.js
+++ b/src/hooks/useSelector.js
@@ -35,14 +35,11 @@ function useSelectorWithStoreAndSubscription(
       selectedState = latestSelectedState.current
     }
   } catch (err) {
-    let errorMessage = `An error occurred while selecting the store state: `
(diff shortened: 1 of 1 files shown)


timdorr:
No, the issue we have is that we catch one error and throw a different error. [My suggestion](https://github.com/reduxjs/react-redux/issues/1433#issuecomment-550008800) is probably the way to go here.

timdorr:
Cool, thanks for putting this together!

PR #1476, opened 2019-11-28
Docs: fix code example in static-types.md

jhujhul:
The code example showing how to use a typed useSelect does not use useTypedSelector but the useSelect hook.


timdorr:
Thanks!

PR #1483, opened 2019-12-10
unnecessary extra semi colon in docs

Amoodaa:
in docs at https://react-redux.js.org/using-react-redux/connect-mapstate#ownprops-optional
```;<Component id={123} />```


timdorr:
This is likely a side-effect of older versions of Prettier. Thanks!

PR #1488, opened 2019-12-30
Deleted dead link to "Redux with react" to eliminate redundancy

MrRyanFloyd:
The link at the end of the docs is dead. No need to have a link to "Using Redux with React" when that's what that doc page is about.


timdorr:
Thanks, but this page does still exist: https://redux.js.org/basics/usage-with-react

I'll update it directly on master.

PR #1496, opened 2020-01-08
fix useIsomorphicLayoutEffect in Node

### LINKED ISSUE #1492: useIsomorphicLayoutEffect flag set wrongly

`useIsomorphicLayoutEffect` is set to true when running on a Node.js instance, where the window object is defined.

This is a common scenario when you need to emulate a subset of a web browser for tests or DOM manipulation, then you define the object perhaps with JSDom.

It's resulting in console warnings and issues when hydrating the SSR code.

`useIsomorphicLayoutEffect` flag is set to false when running in Node.js, always, not relying only on the window object being undefined.

Ref: https://github.com/reduxjs/react-redux/blob/master/src/utils/useIsomorphicLayoutEffect.js#L12

**Used versions**

```bash
react@16.8.4
react-redux@7.1.3
```
This used to work without issues, in `react-redux@6.0.0`.

patricksmms:
Fix condition to work in Node even when the window object is defined

Closes #1492

### DIFF
--- a/src/utils/useIsomorphicLayoutEffect.js
+++ b/src/utils/useIsomorphicLayoutEffect.js
@@ -1,5 +1,13 @@
 import { useEffect, useLayoutEffect } from 'react'
 
+const isBrowser = typeof window !== 'undefined' &&
+  typeof window.document !== 'undefined' &&
+  typeof window.document.createElement !== 'undefined'
+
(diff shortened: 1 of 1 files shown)


markerikson:
Is there any kind of a unit test we can add to better cover this scenario?  Do we have any tests that cover this yet?

PR #1498, opened 2020-01-13
Mark Redux as an optional peer dependency

xtydev:
**TL;DR:** With the introduction of [`@reduxjs/toolkit`](/reduxjs/redux-toolkit), we can now use it in place of `redux`, thus making both peer dependencies **optional**.

As suggested by @markerikson in https://github.com/reduxjs/redux-toolkit/issues/238#issuecomment-547237742,
> One other small downside is that React-Redux says it wants Redux as a peer, and logs a warning if only RSK is installed.

Currently, this warning may very well give the false impression that `@reduxjs/toolkit` must be installed **alongside** `redux` for it to work properly.

Ideally, an `UNMET PEER DEPENDENCY` warning should by displayed **iff** neither `redux`  nor `@reduxjs/toolkit` is installed, but neither `npm` nor `yarn` has a way to specify such a requirement AFAIK. That leaves us no choice but to mark both as optional (similar to the [`react-dom`-vs-`react-native` situation](https://github.com/reduxjs/react-redux/pull/1390#issuecomment-528940544)), if we are to avoid being misleading.

In the future, leaving `redux` as is might pose a bigger problem still with `npm` v7, which will install peer dependencies automatically, as outlined in the [npm CLI Roadmap](https://github.com/npm/cli/wiki/Roadmap#proper-peerdependencies-support) among other places:
> ### Proper `peerDependencies` Support
> Part of the installer rewrite involves taking a fresh look at how `peerDependencies` are handled (and, quite often, _not_ handled) to maximize the cases where the CLI does the right thing. Since npm v3, `peerDependencies` were not installed by default, putting the onus on users to resolve them. We'll be bringing back this functionality in version 7.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,14 +36,19 @@
     "coverage": "codecov"
   },
   "peerDependencies": {
-    "react": "^16.8.3",
-    "redux": "^2.0.0 || ^3.0.0 || ^4.0.0-0"
+    "react": "^16.8.3"
   },
   "peerDependenciesMeta": {
+    "@reduxjs/toolkit": {
+      "optional": true
+    },
     "react-dom": {
       "optional": true
     },
     "react-native": {
+      "optional": true
+    },
+    "redux": {
       "optional": true
     }
   },


timdorr:
What about the fact that RTK requests redux@^4, whereas React Redux is compatible with redux@^2-4? I suppose we should bump up to the same ^4 dependency. 

Another option: What if RTK re-exported React Redux? Sort of akin to react-router-dom re-exporting all of react-router.

markerikson:
RTK is supposed to be independent of React-Redux and usable with any UI layer, in the same way that the Redux core is.  So no, RTK can't depend on React-Redux at all.

Right now, using RTK + React-Redux shows a warning at install time, but we get away with it at runtime because of how package managers and bundlers end up handling things (nested dependencies get flattened to the root of `node_modules`, and you can have import statements for transitive packages without explicitly listing them in `package.json` because of that flattened hierarchy).

In terms of the actual source code, the only true dependency React-Redux has on Redux is the two places where we import `bindActionCreators` ( https://github.com/reduxjs/react-redux/blob/0c048f0646754de103cc2d9825762340bed06119/src/connect/mapDispatchToProps.js#L1 and https://github.com/reduxjs/react-redux/blob/3e53ff96ed10f71c21346f08823e503df724db35/src/utils/wrapActionCreators.js#L1 ).

The store shape aspect (ie, expecting `store` to be an object like `{dispatch, subscribe, getState}`) is more of an implicit / duck-typing kind of thing.

So, strictly speaking, React-Redux itself is indeed compatible with `redux@^2`, because all it needs is `bindActionCreators` + something store-like.

If we were to copy-paste `bindActionCreators` over, we could remove the dependency on Redux entirely, and mark both Redux and RTK as purely optional.  That said, not sure if it's the right approach or not.

package.json
```diff
@@ -36,15 +36,20 @@
...
   },
   "peerDependencies": {
-    "react": "^16.8.3",
-    "redux": "^2.0.0 || ^3.0.0 || ^4.0.0-0"
+    "react": "^16.8.3"
   },
   "peerDependenciesMeta": {
+    "@reduxjs/toolkit": {
+      "optional": true
+    },
```
arjungupta14, line 44:
```suggestion
    },ddd
```

markerikson:
FWIW, [`bindActionCreators` is all of 25-ish lines long](https://github.com/reduxjs/redux/blob/c3cbe2e0551835dc9c400949504b3ba4d884b984/src/bindActionCreators.js).  I vote we just copy-paste it over.  We could even make the shape checks be dev-only to shave a few bytes.

For that matter, in theory you should only get that if you're using `connect` anyway. If you're only using the hooks API, that should never get imported.

timdorr:
Just did that over in #1705

xty:
Well done, guys

Issue #1502, opened 2020-01-20
Investigate adding a proxy-based selector API


markerikson:
Both our `connect` and `useSelector` APIs rely on the user manually specifying what pieces of state to extract to determine re-renders.

@dai-shi has done some amazing work with using ES6 Proxies to manage tracking the pieces of state that are actually needed by a component, particularly in his https://github.com/dai-shi/reactive-react-redux library.

I've said several times that I kinda see that lib as a sort of alpha test for whether an approach like that should be added to React-Redux itself.  Now that our hooks API has been stable for a while, it might be time to start discussing that idea more seriously.

This may also have some relevance to the Concurrent Mode questions over in #1351 .

Ideally, anything we add would be completely tree-shakeable by default, so that it doesn't increase bundle size unless the API is actually used.

dai-shi:
One of the biggest concerns was about performance because Proxies have some overhead.
During last year, I spent much effort on it and reached a reasonable point.
There's a benchmark result here: https://github.com/reduxjs/react-redux-benchmarks/pull/26

dai-shi:
As for CM, there's some relevance because reactive-react-redux tries [it](https://github.com/dai-shi/reactive-react-redux/issues/29).
But technically, proxy-based API (useTrackedState in rrr) is nothing to do with CM, which means we can apply it to the current react-redux v7.

timdorr:
Can we just adopt useTrackedState with some minor changes? @dai-shi Would you be OK with that?

dai-shi:
You mean like react-redux v7.2? Yes, that should be possible. I'm certainly OK with it.

salvoravida:
hi! @dai-shi 
first of all, great work with rrr!

i would like to get a point:
* it is not clear from the user-point-of-view which are the main advg of using `useTrackedState` instead of `useSelector`

in summary, we are replacing 
//useSelector.js
```jsx
  if (equalityFn(newSelectedState, latestSelectedState.current)) {
          return
        }
```
with
//useTrackedState.js
```jsx
if (
        latestTracked.current.state !== nextState &&
        isDeepChanged(
          latestTracked.current.state,
          nextState,
          latestTracked.current.affected,
          latestTracked.current.cache
        )
      ) {
        forceRender()
      } 
```
the main drawback is that user cannot use selectors, and so derivated states.

am i missing something?

markerikson:
Yeah, like I said in the PR:

> We'll need documentation:
> - How does this get used?
[...]

PR #1503, opened 2020-01-20
useTrackedState from reactive-react-redux

dai-shi:
For #1502.

This brings useTrackedState from [reactive-react-redux](https://github.com/dai-shi/reactive-react-redux) and adjusts for react-redux store context.

TODOs:
- [x] check tree-shakable
- [x] custom context support
- [x] spec for useTrackedState
- [x] docs


markerikson:
Cool.

Primary points of concern:

- Does this get properly tree-shaken in typical CRA / RN setups if it isn't used?
- We'll need documentation:
    - How does this get used?
    - Why would you want to use it?
    - What are the differences in behavior compared to `useSelector`?
    - What are the limitations in browser support?

dai-shi:
> tree-shaken

I made a simple counter + textbox app and built with CRA.

### useTrackedState + useSelector

```
wc build/static/js/*.js
       2    2847  140002 build/static/js/2.14348f2c.chunk.js
       1     143    7306 build/static/js/main.592541a8.chunk.js
       1      37    1569 build/static/js/runtime-main.3ed87fa7.js
       4    3027  148877 total
```

### useSelector only

```
wc build/static/js/*.js
       2    2847  140002 build/static/js/2.14348f2c.chunk.js
       1     103    5171 build/static/js/main.0b18769f.chunk.js
       1      37    1569 build/static/js/runtime-main.3ed87fa7.js
       4    2987  146742 total
```

### useTrackedState only

```
wc build/static/js/*.js
       2    2847  140002 build/static/js/2.14348f2c.chunk.js
       1     114    6405 build/static/js/main.dbc372e9.chunk.js
       1      37    1569 build/static/js/runtime-main.3ed87fa7.js
       4    2998  147976 total
```

dai-shi:
I'm not super good at documentation. Would appreciate for any help (not only for contents but also for grammar.)

src/hooks/useTrackedState.js
```diff
@@ -0,0 +1,92 @@
...
+  const affected = new WeakMap()
+  const latestTracked = useRef(null)
+  useIsomorphicLayoutEffect(() => {
+    latestTracked.current = {
+      state,
+      affected,
+      cache: new WeakMap()
+    }
+  })
+  useIsomorphicLayoutEffect(() => {
```
markerikson: While it's not an immediate concern, I'm noticing that we now have 3 different parts of our codebase where we're subscribing to the store in a hook.  That _may_ be worth looking at to see if there's some way to abstract it.

docs/api/proxy-based-tracking.md
```diff
@@ -0,0 +1,145 @@
+---
```
markerikson: This file is not currently added to the Docusaurus sidebar, and so it's not showing up in the docs.

Having said that: for now, let's put the content into the existing "Hooks" docs page.  Longer-term, we might want to consider breaking that page into separate pages per hook, but let's just keep it all in one page at the moment.

docs/api/proxy-based-tracking.md
```diff
@@ -0,0 +1,145 @@
...
+id: proxy-based-tracking
+title: Proxy-based Tracking
+sidebar_label: Proxy-based Tracking
+hide_title: true
+---
+
+# Proxy-based Tracking
+
+This document describes about `useTrackedState` hook.
```
markerikson: This line isn't necessary, especially if we move this content into a separate `useTrackedState` header section in the "Hooks" page.

docs/api/proxy-based-tracking.md
```diff
@@ -0,0 +1,145 @@
...
+
+## How does this get used?
+
+`useTrackedState` is a hook that can be used instead of `useSelector`.
```
markerikson: I'd like to rephrase this intro paragraph.  Something like:

> `useTrackedState` allows components to read values from the Redux store state.  It is similar to `useSelector`, but uses an internal tracking system to detect which state values are read in a component, without needing to define a selector function.

docs/api/proxy-based-tracking.md
```diff
@@ -0,0 +1,145 @@
...
+It doesn't mean to replace `useSelector` completely.
+It gives a new way of connecting Redux store to React.
+
+> **Note**: It's not completely new in the sense that there already exists a library for `connect`: [beautiful-react-redux](https://github.com/theKashey/beautiful-react-redux)
```
markerikson: Drop this link for now.  We could maybe move it to the bottom as a note, like "If you want to try something similar for `connect`, see `beautiful-react-redux`.

docs/api/proxy-based-tracking.md
```diff
@@ -0,0 +1,145 @@
...
+
+The usage of `useTrackedState` is extremely simple.
```
markerikson: I'd like to avoid using phrases like "simple" and "intuitive" in the docs if possible.

docs/api/proxy-based-tracking.md
```diff
@@ -0,0 +1,145 @@
...
+export const TodoListItem = props => {
+  const state = useTrackedState()
+  const todo = state.todos[props.id]
+  return <div>{todo.text}</div>
+}
+```
+
+## Why would you want to use it?
+
+> For beginners: Far easier to understand Redux and R-R without the notion of selectors
```
markerikson: Drop the quote blocks here.  Would also like to see some longer explanations for each of these points.

docs/api/proxy-based-tracking.md
```diff
@@ -0,0 +1,145 @@
...
+
+```js
+const young = useTrackedState().person.age < 11;
+```
+
+Whereas with useTrackedState, a component re-renders whenever the `age` value is changed.
+
+### Caveats
+
+Proxy-based tracking may not work 100% as expected.
```
markerikson: I'd prefer to phrase this like:

> Proxy-based tracking has limitations.

docs/api/proxy-based-tracking.md
```diff
@@ -0,0 +1,145 @@
...
+const state = { count: 0 }
+
+// this works with polyfill.
+state.count
+
+// this won't work with polyfill.
+state.foo
+```
+
+So, if the state shape is defined initiall and never changed, it should be fine.
```
markerikson:
```suggestion
So, if the state shape is defined initially and never changed, it should be fine.
```

dai-shi: Thanks.

docs/api/proxy-based-tracking.md
dai-shi: 👍

docs/api/proxy-based-tracking.md
dai-shi: Makes sense.

docs/api/proxy-based-tracking.md
dai-shi: Dropping for now.

docs/api/proxy-based-tracking.md
dai-shi: 🤔

docs/api/proxy-based-tracking.md
dai-shi: Let me copy&paste. 😄

dai-shi:
> This file is not currently added to the Docusaurus sidebar, and so it's not showing up in the docs.
> 
> Having said that: for now, let's put the content into the existing "Hooks" docs page. 

I think I did it so, but it's not showing in [the preview](https://deploy-preview-1503--react-redux-docs.netlify.com/api/hooks). Is anything wrong?

joelmoss:
Love this! My only concern is there seems to be too many caveats, and holes in which to fall through if not fully aware of how it works.

dai-shi:
I understand your concern. There are some ideas to mitigate caveats, but they are not free either. My question would be how big those holes are in real world apps.

joelmoss:
For me, I think it would be all worth it, but any time there is a hole, it is inevitable that someone will fall in, and not take that fall well at all 😕

@dai-shi how have users of reactive-react-redux handled this? Have you experienced many reports or complaints about the caveats?

dai-shi:
There are not many users of reactive-react-redux, so I can't tell. That's the point of having this in react-redux.
Those caveats are more or less theoretical, and someone expressed concerns but not real issues.
But for the second caveat, I personally hit it. I was so shocked to find it out.
The third one is hypothetical. It may only slow down a bit. Or, it may behave unexpectedly in some middleware.
It would be nice if we could detect falling into the hole. Yeah, that might be worth considering.

joelmoss:
> It would be nice if we could detect falling into the hole. Yeah, that might be worth considering.

Yeah, think that would be a good idea.

The first caveat should be easy to track and to throw an error/warning. Not sure about the others, but if not possible to track and warn, then maybe create a few eslint rules?

joelmoss:
> It would be nice if we could detect falling into the hole. Yeah, that might be worth considering.

Yeah, think that would be a good idea.

The first caveat should be easy to track and to throw an error/warning. Not sure about the others, but if not possible to track and warn, then maybe create a few eslint rules?

dai-shi:
The first one is actually difficult if you consider custom hooks. can eslint rules detect it?

For the second one, I have an idea with heuristics, only effective in DEV. I'll tackle this first.

For the third one, we have several options: a) allow leaking proxies and deproxify before wrapping with a new proxy, b) disallow it and detect it in DEV before wrapping with a new proxy, c) disallow it and detect it in DEV with a patched `dispatch`, d) deproxify always with a patched `dispatch`. Of course, `dispatch` is not only the place to leak proxies, but it should be a typical path.

dai-shi:
> For the second one, I have an idea with heuristics, only effective in DEV. I'll tackle this first.

I was too naive. My first trial reveals that it comes with false positives quite easily. I don't think we can do much at the hooks level.

One thing it would be helpful is to show "what path is tracked" in console. But, this means developers have to understand what is "tracking by proxies".

We have to decide one important point, whether this hook is for beginners or not.

I have assumed this hook is for beginners from the start. If this is the case, I have a crazy idea. We can make the hook so that it will never forget the tracked path by default. That will erase the second caveat. In such a case, I could add an functionality to opt-out the default behavior (meaning, to clear the tracked path in each render) for optimization, **or** just encourage the use of `useSelector` for more (predictable) performance.

Feedback welcome, including request for clarification.

dai-shi:
Let me summarize my thoughts about caveats with possible ideas to mitigate them.

### Caveat 1: Proxied states are referentially equal only in per-hook basis

Unless we use custom hooks, it's not very likely to be misused (and we could detect it by eslint rules.)
With custom hooks, it's hard to detect.
In either case, we would need to explicitly state that the tracked state is special.

### Caveat 2: An object referential change doesn't trigger re-render if an property of the object is accessed in previous render

This is a pitfall that's hard to find if one falls into.
Per my previous comment, we should add some debugging information so that developers can "see" how it is tracking, which should allow finding pitfalls.

### Caveat 3: Proxied state shouldn't be used outside of render

As I said, there are several paths. For this one, I need more feedback.
I shouldn't have said "shouldn't be used", the fact is it's not uncertain what would happen.
Let's just focus on two separate approaches:
1) We allow passing proxied state to `dispatch` and unwrap it before wrapping with a new proxy (as far as I remember, this is how `proxyequal` works)
2) We create another useDispatchForTrackedState, which unwraps proxied state before passing to `dispatch`. (Oh, but this doesn't work if we dispatch a thunk function with state in closure.)

dai-shi:
I've made a few improvements.

- useTrackedState can now show tracked path list with useDebugValue
  - This doesn't directly solve the caveats, but it will help finding issues, so hopefully somewhat mitigated.
- useTrackedState always tries to unwrap the previous proxy before wrapping with a new proxy
  - Basically, this will avoid multiple wrapping, which was a major issue (leading memory leaks) in the previous caveat 3.
  - With this change, passing tracked state to dispatch and reducer should be fine as long as there are no middlware...
  - Some middleware may conflict with this behavior, but nothing is confirmed so far.
- Docs are updated accordingly

dai-shi:
It turns out that this feature can be built on top of useSelector.
https://github.com/dai-shi/react-tracked/pull/71
I still think it would be nice to support it natively in react-redux,
but this add-on usage would also be nice and it works.

artem-malko:
@markerikson @dai-shi hi!)

I'd like to know, why this pull request was not merged?) Looks like useful feature. Di you need any help with it?

dai-shi:
@artem-malko [react-tracked](https://github.com/dai-shi/react-tracked) offers the same capability opt-in. If you are interested in it, check the (naive) example: https://react-tracked.js.org/docs/tutorial-redux-01

Also, check out [the proxy-memoize in the redux docs](https://redux.js.org/usage/deriving-data-selectors#proxy-memoize) for similar (not the same) capability as selectors.

What we can do (help) is to show some use cases, how powerful these approaches are, and possible limitations.

markerikson:
Given that we're working on v8 right now and it's all been migrated to TS, and that `proxy-memoize` now exists, I think it might be time to close this PR.

I'm still open to the idea of _possibly_ adding `useTrackedState`, I'm just not entirely sure it's necessary or that there's enough actual benefit.

@dai-shi , thoughts?

dai-shi:
Yeah, I guess `proxy-memoize` comes with less caveats, and fits well with the current (react-)redux coding style. `useTrackedState` would still be worth with simpler api, and one can use `react-tracked`. 

I close this PR with two alternatives.

## `react-tracked`

https://react-tracked.js.org has `createTrackedSelector` to support react-redux.
We can create the same `useTrackedState` implemented in this PR, from `useSelector`.
(btw, react-tracked does the same with `useContextSelector` for React Context.)

```jsx
import { useSelector } from 'react-redux';
import { createTrackedSelector } from 'react-tracked';

const useTrackedState = createTrackedSelector(useSelector);

const Component = () => {
  const { name, price } = useTrackedState();
  return <>{name}: {price}</>
}
```

## `proxy-memoize`

Redux docs: https://redux.js.org/usage/deriving-data-selectors#proxy-memoize

```jsx
import { useSelector } from 'react-redux';
import memoize from 'proxy-memoize';

const selectItem = memoize(state => ({ name: state.item.name, price: state.item.price }))

const Component = () => {
  const { name, price } = useSelector(selectItem);
  return <>{name}: {price}</>
}
```

artem-malko:
Thx for the info)

PR #1506, opened 2020-01-21
Fix memory leak issue with `UseEffect`

larrylin28:
There's a memory leak in `react-redux`'s usage of `useEffect`, here's the detail:

In the last `useIsomorphicLayoutEffect` usage in `connectAdvanced.js`, it returns a function, `unsubscribeWrapper`, which will be retained by React as `destroy` function of the effect. Since this `useEffect` only subscribes to `store`, `subscription` and `childPropsSelector`, which in most case won't change when store state updates. So this `useEffect` is never called again in following updates of `connected` component. So the effect in the `connected` component will always keep a reference to the `unsubscribeWrapper` created in first call. But this will lead to a memory leak in modern JS VM.

In modern JS VM such as V8(Chrome), the instance of function `unsubscribeWrapper` will retain a closure for context when it's created. Which means, all local variables in first call of each `connected` component will be retained by this instance of `unsubscribeWrapper`, even though they are not used by it at all. In this case, the context includes `actualChildProps`. Which means, every connected component, will in the end retain 2 copy of its props in the memory, one as its current prop, another is the first version of its props.

It can be huge impact of memory if a connected component has props retaining a reference to big chunk of data in store state that can be fully updated to another version(e.g. data parsed from cache, network response, etc). It will end up always retaining 2 copy of that data in memory, or more if there're more other `connected` components.

A better JS VM should optimize to not include unused variable in the closure, but as I tested in V8 and Hermes, they both doesn't have such optimisation to avoid this case.

This can be easy reproduced:
1. Have a connected component, reference one object in the store state.
2. Update the store state(add a version maker in the object to help identify the issue)
3. Use Chrome dev tools to take a heap snapshot.
4. Search for the object in the heap snapshot, you will find 2 version of it in the heap, one retained by connected wrapped component's props, one retained by a `destroy` in lastEffect of connected components.

By communicating with React community, a good solution suggested is to lift `useEffect` outside of the hook component in such kind of case.
And this is how this PR solve the problem.

### DIFF
--- a/src/components/connectAdvanced.js
+++ b/src/components/connectAdvanced.js
@@ -399,8 +399,6 @@ export default function connectAdvanced(
         return childPropsSelector(store.getState(), wrapperProps)
       }, [store, previousStateUpdateResult, wrapperProps])
 
-      // To avoid memory leak issue of function closure in useEffect, move useEffect functions out of component scope.
-
       // We need this to execute synchronously every time we re-render. However, React warns
       // about useLayoutEffect in SSR, so we try to detect environment and fall back to
       // just useEffect instead to avoid the warning, since neither will run anyway.
--- a/src/components/connectAdvanced.js
+++ b/src/components/connectAdvanced.js
@@ -23,8 +23,12 @@ function storeStateUpdatesReducer(state, action) {
   return [action.payload, updateCount + 1]
 }
 
-function useIsomorphicLayoutEffectWithArgs(create, deps, args) {
-  useIsomorphicLayoutEffect(() => create(...args), deps)
+function useIsomorphicLayoutEffectWithArgs(
+  effectFunc,
+  effectArgs,
+  dependencies
+) {
+  useIsomorphicLayoutEffect(() => effectFunc(...effectArgs), dependencies)
 }
 
 function captureWrapperProps(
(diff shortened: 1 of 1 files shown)


larrylin28:
cc @gaearon

timdorr:
Ah, closures strike again! Always making a fool out of the reference counter...

I think this is A-OK. Let me look through some things.

src/components/connectAdvanced.js
```diff
@@ -23,6 +23,127 @@ function storeStateUpdatesReducer(state, action) {
   return [action.payload, updateCount + 1]
 }
 
+function useIsomorphicLayoutEffectWithArgs(create, deps, args) {
```
timdorr: Can we swap the last two args and use slightly more descriptive naming?

```suggestion
function useIsomorphicLayoutEffectWithArgs(effectFunc, effectArgs, dependencies) {
```

The reason being that it reads more naturally to say that you're "calling this function with these args" as a contiguous phrase. Also, it's relatively idiomatic React/useEffect to have the dependency array at the end of the function call.

larrylin28:
@timdorr Applied the suggested change.

timdorr:
LGTM. Let's get another set of eyes on it and then we can see about merging it in.

markerikson:
I'll try to look through it more closely this evening.

markerikson:
Seems plausible enough.  I don't suppose there's any kind of a test we could add to cover this internal implementation change?

Also, does this affect `connect()` at all?

timdorr:
Not really, since it's hard to parse through heap snapshots without a whole bunch of other stuff going on.

markerikson:
I figured as much.

Can we at least have some kind of sandbox or something set up that shows the before and after behavior on this PR, just to confirm it really does make a difference?

markerikson:
Before we go any further on this PR, @jessidhia reported a bug in Reactiflux, which I've filed as #1508 .

We should try to investigate that one first, because she offered up a suggested refactoring of `useSelector` to fix it.  If that's correct, we would want to combine the changes from her gist and this PR - both fixing the race condition, _and_ the memory leak issue.

larrylin28:
@markerikson how does that goes?

markerikson:
Awright, let's get this in.

@larrylin28 just to check, does `useSelector` have a similar issue?

Issue #1512, opened 2020-01-29
Broken link


Ildarik:
https://react-redux.js.org/next/api/hooks

_The selector is approximately equivalent to the **_mapStateToProps argument to connect_**_ conceptually.
There are 2 `next` words in a link: https://react-redux.js.org/next/next/using-react-redux/connect-mapstate

Found that in markdown file it is a relative link: https://github.com/reduxjs/react-redux/blob/master/docs/api/hooks.md
How that works?)

timdorr:
Can you submit a PR?

PR #1513, opened 2020-01-29
Fix broken link

Ildarik:
#1512 

I tried locally and didn't find any issue with that (the link is correct). After adding one level up in relative path it works again and so I confused a little :)


Ildarik:
Didn't work in preview above :(

markerikson:
The original link itself is correct, the problem here has something to do with the versioning aspect somehow.

timdorr:
Cribbed the format from connect.md. I'll copy this over to the versioned docs so it works on all versions.

Issue #1454, opened 2019-11-08
Performance: createListenerCollection


leshkovichpvl:
On a large application, there are problems with the performance of createListenerCollection.

<img width="1914" alt="Screen Shot 2019-11-06 at 18 13 44" src="https://user-images.githubusercontent.com/13157694/68459466-8f018900-0216-11ea-8e2a-e5dca8003a23.png">

https://github.com/reduxjs/react-redux/pull/1450 I tried to solve this problem using a map, but this led to the wrong order of selector calls.

Any ideas on how to improve performance?

markerikson:
Can you provide more details on what the specific perf issue is with the existing implementation?

A reproduceable CodeSandbox example would help.

ckedar:
@leshkovichpvl, try this implementation and see if it helps:
```
const noop = () => {}

function createListenerCollection() {
  const batch = getBatch()
  let listeners = []
  let toBeRemoved = null

  return {
    clear() {
      listeners = CLEARED
    },

    notify() {
      toBeRemoved = [] // indication to unsubscribe() that removal from listeners should be deferred
      const count = listeners.length // capture the count now, so that any listeners added while we are notifiying are not processed in this round
      batch(() => {
        for (let i = 0; i < count; i++) {
          listeners[i]()
        }
      })

      // do the deferred removal
      for(let i = toBeRemoved.length-1; i >= 0; i--) {
        listeners.splice(toBeRemoved[i], 1)
      }
      toBeRemoved = null // unsubscribe() is now free to remove directly from listeners
    },

    get() {
      return listeners
    },

    subscribe(listener) {
      let isSubscribed = true
      listeners.push(listener)

      return function unsubscribe() {
        if (!isSubscribed || listeners === CLEARED) return
        isSubscribed = false

        const index = listeners.indexOf(listener);
        if(toBeRemoved) { 
          // notify() is working, deffer the removal
          listeners[index] = noop
          toBeRemoved.push(index)
        } else {
          listeners.splice(index, 1)
        }
      }
    }
  }
}

```

leshkovichpvl:
@ckedar, Unfortunately, this didn't help improve performance.

timdorr:
Without a real world example of this issue, we can't really do anything here. It looks like you're connecting a *lot* of components, so I might recommend refactoring to reduce your connected component count. That's going to be bad for performance in other ways that can't be fundamentally fixed by the library.

markerikson:
[...]

PR #1523, opened 2020-02-17
Optimize createListenerCollection

wurstbonbon:
Follow-up to https://github.com/reduxjs/react-redux/issues/1454. Using a linked list to store listeners for fast removal. For comparison unmounting 50k connected children under a single parent.
 
**Before:**
before

**Afterwards:**
afterwards

### DIFF
--- a/src/utils/Subscription.js
+++ b/src/utils/Subscription.js
@@ -4,46 +4,68 @@ import { getBatch } from './batch'
 // well as nesting subscriptions of descendant components, so that we can ensure the
 // ancestor components re-render before descendants
 
-const CLEARED = null
 const nullListeners = { notify() {} }
 
 function createListenerCollection() {
   const batch = getBatch()
-  // the current/next pattern is copied from redux's createStore code.
-  // TODO: refactor+expose that code to be reusable here?
-  let current = []
-  let next = []
+  let first = null
+  let last = null
 
   return {
     clear() {
-      next = CLEARED
-      current = CLEARED
+      first = null
+      last = null
     },
 
     notify() {
-      const listeners = (current = next)
       batch(() => {
-        for (let i = 0; i < listeners.length; i++) {
-          listeners[i]()
+        let listener = first
+        while (listener) {
+          listener.callback()
+          listener = listener.next
         }
       })
     },
 
     get() {
-      return next
+      let listeners = []
+      let listener = first
+      while (listener) {
+        listeners.push(listener)
+        listener = listener.next
+      }
+      return listeners
     },
 
-    subscribe(listener) {
+    subscribe(callback) {
       let isSubscribed = true
-      if (next === current) next = current.slice()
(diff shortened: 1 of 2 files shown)


src/utils/Subscription.js
```diff
@@ -4,46 +4,68 @@ import { getBatch } from './batch'
...
   return {
     clear() {
-      next = CLEARED
-      current = CLEARED
+      first = null
+      last = null
     },
 
     notify() {
-      const listeners = (current = next)
```
wurstbonbon: I removed current/next based on this comment https://github.com/reduxjs/react-redux/pull/1450#issuecomment-550382242

markerikson:
I would like to see the actual benchmarks you're using to compare these results first, and get a better understanding of the behavior characteristics may be different between the current implementation and this PR.

wurstbonbon:
This is the benchmark code:
```js
const store = createStore(
);

const App = connect(null, { increment })(
    class extends React.Component {
        state = { numChildren: 0 };

        toggleChildren = () => {
            this.setState(state => ({ numChildren: state.numChildren ? 0 : 50000 }));
        };

        render() {
            return (
                <div>
                    <button onClick={this.toggleChildren}>Toggle Children</button>
                    {[...Array(this.state.numChildren).keys()].map(i => (
                        <Child key={i} />
                    ))}
                </div>
            );
        }
    }
);

const Child = connect(state => ({ state }))(
    class extends React.Component {
        render() {
            return <div className="child">{this.props.state}</div>;
        }
    }
);

ReactDOM.render(
    <Provider store={store}>
        <App />
    </Provider>,
    document.getElementById("root")
);
```
What I'm profiling is the unmounting. This is obviously specifically designed to expose the quadratic complexity of the unsubscribe function.

markerikson:
Can you clarify what specifically about the current implementation is "quadratic"?

wurstbonbon:
This line: ```next.splice(next.indexOf(listener), 1)```

markerikson:
Okay, yeah, that definitely looks like an issue.

wurstbonbon:
This statement `next.splice(next.indexOf(listener), 1)` for every listener that is remove we need look at all other listeners. And the splice should also scale with size of the array.

wurstbonbon:
Screenshot 2020-02-17 at 22 54 16

timdorr:
LGTM. It may be a tiny bit less memory efficient (there is a closure around `listener` that keeps the ref alive), but I think that's an acceptable tradeoff when this gets noticeable on something like a route change.

markerikson:
Yeah, I vote we put this in, merge #1506 , and put out 7.2.0.

markerikson:
I'll publish shortly.

markerikson:
And https://github.com/reduxjs/react-redux/releases/tag/v7.2.0 is live.

PR #1526, opened 2020-02-22
docs(hooks): clarify useSelector re-runs on render

AndersDJohnson:
The documentation on `useSelector` was a little unclear as to when it re-runs on function component renders. It does not seem to re-run on re-renders when a selector reference doesn't change (as was hinted at below, but not above).

### DIFF
--- a/docs/api/hooks.md
+++ b/docs/api/hooks.md
@@ -40,7 +40,7 @@ Allows you to extract data from the Redux store state, using a selector function
 
 > **Note**: The selector function should be [pure](https://en.wikipedia.org/wiki/Pure_function) since it is potentially executed multiple times and at arbitrary points in time.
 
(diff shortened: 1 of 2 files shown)


markerikson:
Thanks.  Unfortunately, due to how we version the docs, this will only show up in the "Next" version section, and not the "7.1" version section.  Can you update the matching file under `versioned_docs` as well?

AndersDJohnson:
> Thanks. Unfortunately, due to how we version the docs, this will only show up in the "Next" version section, and not the "7.1" version section. Can you update the matching file under `versioned_docs` as well?

@markerikson Good call. Done in 579415c!

markerikson:
Yeah, I'm still really not happy with how versioning gets handled in Docusaurus, and I don't think it gets any better in v2.  This is exactly my concern with it - folks don't know to edit the `versioned_docs` files, and the `/docs` folder is actually "next" and not "current".

Thanks!

PR #1527, opened 2020-02-24
create tests for functions in mapDipatchToProps

phapdinh:
Hi I saw that there were no tests for the functions in mapDispatchToProps and thought I would help by adding some. Those functions can be tricky so I thought I would add to make sure they don't break


timdorr:
Thanks, but I'd rather not add small unit tests like this. I'm more concerned with integration tests, as per Kent C Dodds' methodology: https://kentcdodds.com/blog/write-tests/

Issue #1236, opened 2019-04-12
Deprecating connectAdvanced


timdorr:
I intend to remove the `connectAdvanced` API from React Redux, essentially replacing `connect` with its contents.

Why? It's an abstraction that we don't use ourselves, so it just becomes an indirection in the code. It is only minimally used in the community. It creates a more complex code base, which is a barrier to entry for many contributions. Plus, frankly, I just don't like it; it gives me [Java flashbacks](https://gph.is/13o8Kjq).

This is mainly to give the few folks that do use it a heads up. It won't go away until the next major release. However, I intend to add a deprecation warning at some point in the future. We'll do that _after_ Hooks, just so we're not dropping two big changes in one minor release. Plus, you can lock to that version to have your Hooks and your connectAdvanced at the same time. Sound fair?

If there are specific use cases, we can see about migrating them into the `connect` API, most likely as options. This isn't about completely throwing away a big chunk of functionality, just simplifying the code base a bit and making it easier to grok for everyone.

gaearon:
👍 I could never understand what it does or why it was added. It sounds like if you need this kind of complexity you should just vendor the connect code and maintain it yourself.

markerikson:
FWIW, the intended use case is for an end user to be able to customize how all the memoization and prop generation is handled, on their own.  All the logic for managing the subscription itself stays the same - it's really just changing from calling `mapState` and `mapDispatch` specifically, to `(storeState, dispatch) => childProps`, and `connect` implements its `mapState/mapDispatch` handling as one possible `(storeState, dispatch) => childProps` function.

If y'all are really dead set on removing it, we can, but I truly don't think it adds that much extra complexity.

Lemme tag three folks who I know have a particular interest in this: @jimbolla , @epeli , @mariusandra .

epeli:
Bummed about this of course but I understand the reasoning and I can live with. Actually just realized I can achieve everything I want with the `areStatePropsEqual` option of `connect()`.

Farewell `advancedConnect()`.

mariusandra:
This would be a bummer. [Kea](https://kea.js.org) is basically built around it. I'll need to see if I can work around it and will provide a more insightful reply next week, as I'm literally boarding a plane for a weekend getaway now.

Cheers!

jimbolla:
I'm not using Redux so I don't have an opinion on this.

mariusandra:
[...]

PR #1528, opened 2020-02-25
remove options parameter

phapdinh:
I checked the use of the wrapMapToPropsConstant function in the codebase and it is only used in wrapMapToProps and mapDispatchToProps. All three places, the function passed in doesn't take in a second parameter. This might be left over code from a previous refactor or maybe left over for some future feature that wants to pass options as a parameter into the function. However, I thought I bring it up to your attention.

### DIFF
--- a/src/connect/wrapMapToProps.js
+++ b/src/connect/wrapMapToProps.js
@@ -1,8 +1,8 @@
 import verifyPlainObject from '../utils/verifyPlainObject'
 
 export function wrapMapToPropsConstant(getConstant) {
-  return function initConstantSelector(dispatch, options) {
-    const constant = getConstant(dispatch, options)
+  return function initConstantSelector(dispatch) {
+    const constant = getConstant(dispatch)
 
     function constantSelector() {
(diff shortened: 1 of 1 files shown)


timdorr:
Nope, it's passed right here: https://github.com/reduxjs/react-redux/blob/5e0c50d814b08cbdd46f1fcce77b34210d546669/src/connect/selectorFactory.js#L104-L106

phapdinh:
@timdorr options is passed into the initMapDispatchToProps function, but the getConstant function doesn't use it. So your passing in an option parameter that is not being used in the codebase. I can update the PR to remove the options from the initMapDispatchToProps. However, the options is not being used by the nested function. This would be more apparent in typescript, but right now javascript doesn't throw an error if you pass in unnecessary parameters

timdorr:
Well, the long term plan is to drop connectAdvanced (#1236) and refactor all of these complex factory functions into something much simpler. So, this will go away one way or another. Thanks for bringing it up!

PR #1535, opened 2020-03-05
Fix link to connect API

allhailwesttexas:
Link to connect.md works in preview but is gives 404 for website Update internal link from `../api/connect.md` to `../api/connect` to fix.

### DIFF
--- a/docs/introduction/basic-tutorial.md
+++ b/docs/introduction/basic-tutorial.md
@@ -352,7 +352,7 @@ export default connect(
 )(Component)
 ```
 
-These four cases cover the most basic usages of `connect`. To read more about `connect`, continue reading our [API section](../api/connect.md) that explains it in more detail.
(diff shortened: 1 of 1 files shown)


allhailwesttexas:
Hm, maybe this is not the right fix. Will have another look.

markerikson:
Intra-page links should always point to the Markdown file name, not the final URL.

If the link is 404-ing, then there's something else that needs to be done to make the link correct.

timdorr:
Actually, this may be due to a bug in Docusaurus. We had to do the same sort of thing in dd17c64791baaebb1b50910f2febf3271416ac22. Can we see about upgrading to 2.0?

nickmccurdy:
FWIW I'm noticing that the Netlify deploy preview isn't updating in another pull request: #1543

allhailwesttexas:
Yes, I found that as well. Didn't have much luck with the Docusaurus previews unfortunately.

allhailwesttexas:
Closing as this is not the right fix for the problem.

PR #1537, opened 2020-03-05
Fixed mapDispatchToProps example with parentheses

SaveYourTime:
To return an object using the arrow function, wrap the return value in parentheses.


markerikson:
Good catch, thanks!

PR #1543, opened 2020-03-10
Add "Create an App" to quick start

nickmccurdy:
See reduxjs/redux#3726

### DIFF
--- a/docs/introduction/quick-start.md
+++ b/docs/introduction/quick-start.md
@@ -13,20 +13,6 @@ sidebar_label: Quick Start
 
 React Redux 7.1 requires **React 16.8.3 or later.**
 
-To use React Redux with your React app:
-
-```bash
-npm install react-redux
-```
-
-or
-
-```bash
-yarn add react-redux
-```
-
-You'll also need to [install Redux](https://redux.js.org/introduction/installation) and [set up a Redux store](https://redux.js.org/recipes/configuring-your-store/) in your app.
-
 ### Using Create React App
 
 The recommended way to start new apps with React Redux is by using the [official Redux+JS template](https://github.com/reduxjs/cra-template-redux) for [Create React App](https://github.com/facebook/create-react-app), which takes advantage of [Redux Toolkit](https://redux-toolkit.js.org/).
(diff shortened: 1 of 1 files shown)


nickmccurdy:
Deploy preview is out of date for some reason, looks like the quick start page from master

docs/introduction/quick-start.md
```diff
@@ -27,6 +27,16 @@ yarn add react-redux
 
 You'll also need to [install Redux](https://redux.js.org/introduction/installation) and [set up a Redux store](https://redux.js.org/recipes/configuring-your-store/) in your app.
 
+## Create an App
```
timdorr:
```suggestion
### Using Create React App
```

Let's make this an h3, so it's part of the overall installation section. Otherwise it looks like you do both this and the install command before it.

We should also move it above the vanilla install steps, so it's the first thing and fits in as the "recommended" option.

timdorr:
We have versioned docs, so you're editing the master set that need to be synced to the version directory occasionally.

markerikson:
Yeah, this is my ongoing frustration with the versioned docs thing, and it doesn't seem like DS v2 is gonna make that any better.

timdorr:
The docs on versioning are somewhat confusing as to where the docs files are supposed to live. I think if we nuke the `versioned_docs/7.2` path, it will pull from the root path for that "current" version. At least, that's what I think will happen. It's not super obvious, unfortunately.

docs/introduction/quick-start.md
nickmccurdy: I've updated the docs for [Redux](https://redux-toolkit.js.org/introduction/quick-start#create-a-react-redux-app) and [Redux Toolkit](https://redux.js.org/introduction/getting-started#redux-toolkit), and they still have the template recommendation after installation with the same heading level. Should I update those pages to be consistent?

nickmccurdy:
Let me know if I need to do anything to fix the deploy, I'm not familiar with Docusaurus' versioning

timdorr:
Let's merge this in for now and we can fix the versioning stuff elsewhere.

Issue #1552, opened 2020-03-23
internal usage of forwardedRef prevents a parent component from passing a forwardedRef prop to a connected component


dtschust:
**What is the current behavior?**

If a parent of a connected component passes in a prop `forwardedRef`, it is passed through as `ref` to the child component. This is happening because `connectAdvanced` is using `forwardedRef` internally, for the use case where `forwardRef` is set to true when `connect` is called.
 
[Here is what I feel is expected behavior](https://codesandbox.io/s/old-react-redux-demo-hvu5i), using `react-redux` v5.1.2 . The `App` component passes a prop `forwardedRef` to the `connect`ed `Component`, which then uses that `forwardedRef` prop to pass a reference to an `Input`. The behavior of this example is that clicking the button will focus the input.

[And here is the current behavior](https://codesandbox.io/s/new-react-redux-demo-lock1), using `react-redux` v7.1.3. In this example, the top two buttons do not focus on their inputs, because the `connect` HOC is intercepting the `forwardedRef` prop passed in by `App`. The third button works, following the exact same pattern but using a name other than `forwardedRef`, in this case `refToForward`.

**What is the expected behavior?**

I think `react-redux` should use a different internal name for this field, perhaps something like `__REACT_REDUX_FORWARDED_REF`, which would limit namespace collision. `forwardedRef` is a particularly likely name to collide on because [it's the name used in React's forwarding refs documentation](https://reactjs.org/docs/forwarding-refs.html#forwarding-refs-in-higher-order-components). I'm happy to make a PR for this change if you think this is a change worth making. The `forwardedRef` key is used internally only so I believe it would be a non-breaking change.

timdorr:
If you want to make a PR, that's cool. But given it's not enabled by default and is a fairly uncommon occurrence that you're forwarding a ref twice through a component, I'm going to close this issue out. I don't think we need to track this very rare edge case.

PR #1553, opened 2020-03-24
Rename internal forwardedRef usage

dtschust:
This addresses issue https://github.com/reduxjs/react-redux/issues/1552, where if a parent passes in a `forwardedRef` prop to a connected component (with default options), the connected child will not receive this prop. This change has the same issue with the prop `reactReduxForwardedRef`, but `forwardedRef` is a prop name commonly used and provided in [React's documentation](https://reactjs.org/docs/forwarding-refs.html#forwarding-refs-in-higher-order-components), so `reactReduxForwardedRef` is much less likely to be used elsewhere.

I'm happy to choose a name other than `reactReduxForwardedRef` if there are any suggestions, I went with something that would hopefully be clear that it is for internal use only and would be unlikely to be used by any other codebase.

A bit of context for why I'm proposing this change, if it helps. I migrated a large codebase from `react-redux` v5 to v7 recently. This codebase had a lot of instances of refs being passed down as a `forwardedRef` prop (per the suggestion in React's documentation). Many of our components are connected components in order to access feature flags. The upgrade broke all cases where a `forwardedRef` prop was passed through a connected component. I fixed this in the codebase by renaming all of these props `refToForward` and making a lint rule to forbid the use of `forwardedRef`. This was a little silly, but was a high confidence way to resolve the issue. I believe `connect` swallowing `forwardedRef` props is a very minor pain point and can cause confusion for developers, and since the fix isn't too involved I'm of the opinion it's worth doing.

It's your all's library though, and I understand if you don't want to merge this. Thanks for all the you do!

### DIFF
--- a/src/components/connectAdvanced.js
+++ b/src/components/connectAdvanced.js
@@ -272,12 +272,12 @@ export default function connectAdvanced(
     const usePureOnlyMemo = pure ? useMemo : callback => callback()
 
     function ConnectFunction(props) {
-      const [propsContext, forwardedRef, wrapperProps] = useMemo(() => {
+      const [propsContext, reactReduxForwardedRef, wrapperProps] = useMemo(() => {
         // Distinguish between actual "data" props that were passed to the wrapper component,
         // and values needed to control behavior (forwarded refs, alternate context instances).
         // To maintain the wrapperProps object reference, memoize this destructuring.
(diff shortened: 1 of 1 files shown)


timdorr:
Thanks!

markerikson:
Finally live in https://github.com/reduxjs/react-redux/releases/tag/v7.2.1

PR #1557, opened 2020-03-30
fix(docs): fixes broken link on connect-mapstate page

kutyepov:
Just noticed a broken link in the docs :)

### DIFF
--- a/website/versioned_docs/version-7.2/using-react-redux/connect-extracting-data-with-mapStateToProps.md
+++ b/website/versioned_docs/version-7.2/using-react-redux/connect-extracting-data-with-mapStateToProps.md
@@ -93,7 +93,7 @@ function mapStateToProps(state) {
 // component will receive: props.a, props.todos, and props.filter
 ```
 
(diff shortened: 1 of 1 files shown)


markerikson:
Appreciate the PR, but links should always be written pointing to the specific Markdown files, not the public URLs.  Can you fix the PR to do that?

kutyepov:
Hi @markerikson,

In that case, fix would be to decline this PR. Relative URLs do work if docs are read on github, but do not work on the website. On the website, we should point to the website itself, not to some .md files that do not exist there.

Also, just noticed another broken url: https://react-redux.js.org/using-react-redux/accessing-store#using-the-usestore-hook

If you click on `useStore hook` you will get 404 as well.

markerikson:
That's what I'm saying:

- The docs are meant to be read on the published docs website, not in the Github repo
- Intra-page URLs in the Markdown files should always point to Markdown files.  They will be processed at build time by Docusaurus and turned into relative HTTP URLs pointing to the correct pages.

Please feel free to submit PRs that do actually fix any broken URLs correctly - you can confirm that they work right by using either the local Docusaurus dev server, or the PR preview links.

kutyepov:
Can you please be a little more specific on "fix any broken URLs correctly"?

Below gif shows an issue I am trying to fix.

react-redux-404

markerikson:
So there's two different answers there.

The first is that we never got around to write that "Advanced Usage: Factory Functions" page :)  The actual potential fixes for that one are either A) write the docs page, or B) remove the link and change the sentence for now.

However, we _do_ have actual API reference content for `useStore`.

I'm saying that the correct fix for the latter would be something like `../api/hooks.md#useStore`.

PR #1558, opened 2020-03-31
Update basic-tutorial.md - convert class component to functional component

indatawetrust:


markerikson:
Appreciate the PR, but I don't think just changing a snippet here is the right approach.

What we need to do is update the docs to show use of our hooks API as the default approach, and leave this page as a secondary tutorial that shows how to use `connect`.

PR #1566, opened 2020-04-13
Feature: add object shorthand for mapStateToProps in connect

GregoryNative:
I found it useful for myself.

Same as we use shorthand for mapDispatchToProps in connect, it's also possible to do for mapStateToProps using reselect library. It's good 1 file library, and even mentioned in redux here:
https://redux.js.org/recipes/computing-derived-data#selectorsindexjs

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,5 @@ website/build/
 website/yarn.lock
 website/node_modules
 website/i18n/*
+
+.DS_Store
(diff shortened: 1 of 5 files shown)


markerikson:
Hi.  Appreciate the PR.  This was previously attempted in #323 , #723 , and #724 .  Ultimately, we decided that there was too much complexity in how this _could_ work, and that it was not a feature we wanted to add.

We will continue to maintain `connect` in general, but at this point we're recommending that folks use the hooks API as the default approach, so I don't see a need to add more features to `connect`.

In addition, our implementation has avoided any additional external dependencies like `reselect`, and we're not going to add another dependency for a rarely-used feature.

GregoryNative:
Np

PR #1567, opened 2020-04-15
removed a redundant object creation when using forceUpdate

vzaidman:
there's no need to pass an object to `forceRender` like this: `forceRender({})`

this is enough: `forceRender()` and saves us a redundant object creation.

notice the example in the official react docs:
https://reactjs.org/docs/hooks-faq.html#is-there-something-like-forceupdate


timdorr:
Thanks!

PR #1568, opened 2020-04-15
extract and improve force render to be a custom hook "useForceRender"

vzaidman:
1. now it's reusable
2. it won't re-create the function inside the "useReducer" function it uses internally for each hook on every single render, instead it has only one instance of this function now (`forceRenderFn`)
3. added a test to ensure it works as expected

### DIFF
--- a/src/utils/customHooks.js
+++ b/src/utils/customHooks.js
@@ -1,0 +1,5 @@
+import { useReducer } from 'react'
+
+// based on https://reactjs.org/docs/hooks-faq.html#is-there-something-like-forceupdate
+const forceRenderFn = s => s + 1
+export const useForceRender = () => useReducer(forceRenderFn, 0)[1]
--- a/src/hooks/useSelector.js
+++ b/src/hooks/useSelector.js
@@ -1,8 +1,9 @@
-import { useReducer, useRef, useMemo, useContext, useDebugValue } from 'react'
+import { useRef, useMemo, useContext, useDebugValue } from 'react'
 import { useReduxContext as useDefaultReduxContext } from './useReduxContext'
 import Subscription from '../utils/Subscription'
 import { useIsomorphicLayoutEffect } from '../utils/useIsomorphicLayoutEffect'
 import { ReactReduxContext } from '../components/Context'
+import { useForceRender } from '../utils/customHooks'
 
 const refEquality = (a, b) => a === b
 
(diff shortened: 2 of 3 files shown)


src/utils/customHooks.js
```diff
@@ -0,0 +1,5 @@
+import { useReducer } from 'react'
+
+// based on https://reactjs.org/docs/hooks-faq.html#is-there-something-like-forceupdate
+const forceRenderFn = s => s + 1
```
vzaidman, line 4: `forceRenderFn` is only created once instead of on each render of the hook

dai-shi:
Hi, just a comment from my experience. I used to use `useForceUpdate` a lot, but I later preferred the `useReducer` approach. (BTW, I now prefer avoiding forceUpdate at all, and no useRef.)
The inline `s => s + 1` doesn't have a closure, and it would never be a performance bottleneck because JS engine optimize it, I presume. The drawback of having `useForceUpdate` for me was 1) `exhaustive-deps` rule doesn't detect it as stable, and 2) it may (or may not) increase bundle size. Hope it helps.

vzaidman:
> doesn't have a closure, and it would never be a performance bottleneck

I believe newer browsers won't be hurt and older once would benefit.

> `exhaustive-deps` rule doesn't detect it as stable

Can you elaborate? There are no deps in `useReducer` or `useForceRender`

> it may (or may not) increase bundle size

Yea it indeed increased the minimized build by 33 bytes. not sure if it's considered a lot

dai-shi:
> Can you elaborate? There are no deps in useReducer or useForceRender

If you use eslint-plugin-react-hooks v2, and enable `exhaustive-deps`:

```js
const [, forceUpdate] = useReducer(c => c + 1, 0);
useEffect(() => {
  ...
  forceUpdate();
  ...
}, []); // no need to specify forceUpdate, because the eslint rule knows it's stable.
```

----

So, defining `const forceRenderFn = s => s + 1` and using `useReducer` seems good, which won't increase bundle size. It's not a strong opinion.

timdorr:
This isn't really a necessary change. It adds a tiny amount of cognitive overhead (now you have to look in another file) and doesn't have a practical performance or quality gain. Thank you for looking at this, but we're not going to add this at this time.

vzaidman:
> So, defining `const forceRenderFn = s => s + 1` and using `useReducer` seems good, which won't increase bundle size. It's not a strong opinion.

Thanks! This new knowledge is what I take from the pr. :)

PR #1570, opened 2020-04-20
Updated siteConfig.js

chinmaykude:
Updated the proper copyright symbol and the current year in the footer section.

Currently:
Copyright (c) 2015-present Dan Abramov and the Redux documentation authors.

With this change:
Copyright © 2015–2020 Dan Abramov and the Redux documentation authors.
(The changes in the end year would be dynamic)


timdorr:
That won't match the license of the repo: https://github.com/reduxjs/react-redux/blob/master/LICENSE.md

Putting a start year through the "present" is considered legally valid.

PR #1571, opened 2020-04-22
docs: Fix broken link in mapStateToProps.md

anthonyc1:


docs/using-react-redux/connect-extracting-data-with-mapStateToProps.md
```diff
@@ -92,7 +92,7 @@ function mapStateToProps(state) {
 // component will receive: props.a, props.todos, and props.filter
 ```
 
-> Note: In advanced scenarios where you need more control over the rendering performance, `mapStateToProps` can also return a function. In this case, that function will be used as the final `mapStateToProps` for a particular component instance. This allows you to do per-instance memoization. See the [Advanced Usage: Factory Functions](../api/connect.md) section of the docs for more details, as well as [PR #279](https://github.com/reduxjs/react-redux/pull/279) and the tests it adds. Most apps never need this.
+> Note: In advanced scenarios where you need more control over the rendering performance, `mapStateToProps` can also return a function. In this case, that function will be used as the final `mapStateToProps` for a particular component instance. This allows you to do per-instance memoization. See the [Advanced Usage: Factory Functions](../api/connect) section of the docs for more details, as well as [PR #279](https://github.com/reduxjs/react-redux/pull/279) and the tests it adds. Most apps never need this.
```
x86chi, line 95: I think the right path is [`../api/connect#factory-functions`](https://react-redux.js.org/api/connect#factory-functions)

Suggest change to `[Notes : Factory Functions](https://react-redux.js.org/api/connect#factory-functions)`

timdorr:
This doesn't work because it breaks the link on GitHub. It should work in both places.

PR #1577, opened 2020-05-02
Fix a borken link of redux doc 

x86chi:
> https://redux.js.org/api/bindactioncreators

Update `bindActionCreators` API Link


timdorr:
Thanks!

PR #1581, opened 2020-05-07
Fix error message typo in useSelector ('You must pass a selector...).

Pixelwelder:
This is a tiny edit, but the current wording did send me hunting for `useSelectors` in the source.


timdorr:
Huh, don't know how we missed that. Thanks!

PR #1593, opened 2020-06-16
Apply React.memo twice to ensure that props object reference stays the same

Hypnosphi:
See https://github.com/facebook/react/issues/18901

Ideally, it should be fixed on React side though

### DIFF
--- a/src/components/connectAdvanced.js
+++ b/src/components/connectAdvanced.js
@@ -471,7 +471,8 @@ export default function connectAdvanced(
     }
 
     // If we're in "pure" mode, ensure our wrapper component only re-renders when incoming props have changed.
-    const Connect = pure ? React.memo(ConnectFunction) : ConnectFunction
+    // Applying React.memo twice ensures that props object reference doesn't change unless some of the props has changed.
+    const Connect = pure ? React.memo(React.memo(ConnectFunction)) : ConnectFunction
 
     Connect.WrappedComponent = WrappedComponent
     Connect.displayName = displayName
--- a/src/components/connectAdvanced.js
+++ b/src/components/connectAdvanced.js
@@ -471,8 +471,10 @@ export default function connectAdvanced(
     }
 
     // If we're in "pure" mode, ensure our wrapper component only re-renders when incoming props have changed.
-    // Applying React.memo twice ensures that props object reference doesn't change unless some of the props has changed.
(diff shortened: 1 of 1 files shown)


markerikson:
Can you clarify exactly what behavior this is trying to fix?  I'm not sure I understand what the problem is.

Hypnosphi:
`props`, and therefore `wrapperProps` change by reference each time parent component re-renders, which leads to unnecessary calculations and effects

markerikson:
Can you put together a sandbox that shows how this specifically affects React-Redux?  Also, how does this change affect the React-Redux benchmarks repo results?

Hypnosphi:
Looks like it only adds some more calls to react-redux internal functions, there's no external effect thanks to an additional [check](https://github.com/reduxjs/react-redux/blob/master/src/connect/selectorFactory.js#L75)

Here's the sanndbox https://codesandbox.io/s/focused-goldberg-jres3?file=/src/App.js

If you open devtools on detached view ("Open in new window"), you can find `https://jres3.csb.app/node_modules/react-redux/es/components/connectAdvanced.js` file in Sources tab and place a debug point at line 183 which maps to https://github.com/reduxjs/react-redux/blob/master/src/components/connectAdvanced.js#L407

You can observe that this line is only executed if you click "update redux" after "update parent"

I will also check how my change affects React-Redux benchmarks

timdorr:
This won't have any effect because the props object is spread:

```js
<Connect {...props} />
```

React doesn't compare the props as an object instance, it does [a shallow comparison of their values](https://github.com/facebook/react/blob/d4fc2c1457bdee52367df3ec2dc9348a12ab85b3/packages/react-reconciler/src/ReactFiberBeginWork.new.js#L525).

If there's something else going on with React not keeping child props memoized between parent re-renders, that's something internal to React. We shouldn't mis-use their APIs to workaround their bug.

Hypnosphi:
> React doesn't compare the props as an object instance

But react-redux [does](https://github.com/reduxjs/react-redux/blob/master/src/components/connectAdvanced.js#L285)

Hypnosphi:
> This won't have any effect because the props object is spread:
>
> `<Connect {...props} />`

Looks like it's only the case for `forwardRef: true`. Can be solved by applying `memo` after `forwardRef`, not before

PR #1598, opened 2020-06-30
docs(hooks.md): clarify `useDispatch`, see #1468

### LINKED ISSUE #1468: Memoize useDispatch?
The docs say

> When passing a callback using dispatch to a child component, it is recommended to memoize it with useCallback, since otherwise child components may render unnecessarily due to the changed reference.

with the example code being

```js
  const incrementCounter = useCallback(
    () => dispatch({ type: 'increment-counter' }),
    [dispatch]
  )
```

I think this could use some clarification. I am understanding it to not be specific to useDispatch, and to just be the classic react thing of don't pass anonymous functions to children since they are different every render. 

If that is the case, it would probably be worth highlighting that it's not specific to useDispatch. If I am misunderstanding, then is there any reason that react-redux can't export some helper that calls useCallback for you? I imagine most people would not realize they need to do this.

papb:
This PR clarifies the memoization of `useDispatch` and how the generated `dispatch` function can be safely used as a dependency of an `useCallback` hook.

I wrote this thinking of what I would have liked to read earlier today. This little clarification would have saved some time.

Thanks for considering!

-----

This would resolve #1468 (in my opinion) if it wasn't already closed.

### DIFF
--- a/docs/api/hooks.md
+++ b/docs/api/hooks.md
@@ -219,7 +219,7 @@ const dispatch = useDispatch()
 
 This hook returns a reference to the `dispatch` function from the Redux store. You may use it to dispatch actions as needed.
 
-*Note: like [React's `useReducer`](https://reactjs.org/docs/hooks-reference.html#usereducer), the returned `dispatch` function identity is stable and won't change on re-renders.*
+*Note: like in [React's `useReducer`](https://reactjs.org/docs/hooks-reference.html#usereducer), the returned `dispatch` function identity is stable and won't change on re-renders (unless you change the `store` being passed to the `<Provider>`, which would be extremely unusual).*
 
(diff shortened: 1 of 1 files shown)


timdorr:
Thanks!

markerikson:
Unfortunately, I think this might have the same problem we've been running into with the versioning: it only updates the file under `/docs`, and not in the versioned folders, so it'll only show up under the "Next" version instead of the main page.

papb:
@markerikson Hmm, should I create mirror PRs targeting other branches? Which ones?

markerikson:
It's not "other branches" - it's the files under https://github.com/reduxjs/react-redux/tree/master/website/versioned_docs .

PR #1602, opened 2020-07-04
Removed unused variable in server-rendering test

keiya01:
I removed unused variable. 
Thank you.


timdorr:
Thanks, but we may adjust this test to check the layout later. Plus, keep it around doesn't hurt anything.

PR #1621, opened 2020-08-05
fix mapDispatchToProps ownProps example description

sag1v:
When using mapDispatchToProps we re-binding the props on component mount and not on re0rendering like the current docs mentions.

This fixes it.


markerikson:
Thanks!

PR #1626, opened 2020-08-30
Fix redundant check

jzendo:
Only need check x or y is non-zero(+0 or -0).


timdorr:
Thanks, but this isn't needed.

jzendo:
Simple is beauty.
😊

Issue #1583, opened 2020-05-08
Incorrect/dead links


jinhale:
From some pages including [accessing the store](https://react-redux.js.org/using-react-redux/accessing-store) there is at least one bad [link](https://react-redux.js.org/api/hooks.md#useStore).  

To correct the link remove the portion of the URL that is '.md'. This may need to be done for multiple pages.

timdorr:
This seems to be a common issue with Docusaurus. They _should_ be converted from the file links to the generated URLs when the site is built. Maybe we have some setting wrong?

josephMG:
https://github.com/facebook/docusaurus/releases/tag/v2.0.0-alpha.64
https://github.com/facebook/docusaurus/pull/3429
upgrade docusaurus should fix it.

Issue #1588, opened 2020-05-18
useStore link not found


cesarvarela:
The link from here: https://react-redux.js.org/using-react-redux/accessing-store#using-the-usestore-hook
Returns 404: https://react-redux.js.org/api/hooks.md#useStore
Seems the `.md` shouldn't be there.

timdorr:
Could you submit a PR to fix this? Thanks!

PR #1638, opened 2020-10-06
Update docusaurus

josephMG:
Fix https://github.com/reduxjs/react-redux/issues/1583 and https://github.com/reduxjs/react-redux/issues/1588 by upgrade docusaurus.

Refs:
1. https://github.com/facebook/docusaurus/releases/tag/v2.0.0-alpha.64
2. https://github.com/facebook/docusaurus/pull/3429


timdorr:
Hmm, do we want to upgrade to 2.0? Everything appears A-OK, but we would be using the legacy theme.

markerikson:
Yes, I _definitely_ want us to upgrade to DS v2!  Just a question of whether the upgrade correctly handles all the versioning aspects we're using from v1 or not.

markerikson:
Hmm.  This PR only seems to bump the DS version.  Based on the work upgrading the Redux core docs from DSv1 to v2, I would expect this also requires a number of changes to the Docusaurus config files as well.

@josephMG , can you review the Docusaurus config files from https://github.com/reduxjs/redux/tree/master/website and https://github.com/reduxjs/redux-toolkit/tree/master/website and see what config changes need to be applied here as well?  I assume that `docusaurus.config.js`, `sidebars.js`, and the `src/css/custom.css` files would need to be updated, at a minimum - possibly others too.

josephMG:
That's OK. You can just close my PR If it's not emergency. 
Thanks.

josephMG:
Yes I only upgrade version to DSv2. I think it still run DSv1 because I only modify `docusaurus-1.x` in [Docusaurus package](https://github.com/facebook/docusaurus/tree/master/packages).

@markerikson   Let me try it. I'm glad to help the rest of things to DSv2.

markerikson:
@josephMG hey, this is looking great so far!  The PR preview is definitely using DS v2 now, and it looks like the versioning is (hopefully?) cooperating.

Will need to get the theme colors updated to match the other sites, and both the RTK and Redux core docs have a bunch of extra custom CSS that I'd like to pull over.  Loving where this is going, and thank you!

josephMG:
@markerikson I fixed styles to match both the RTK and Redux. Btw, this [version](https://v2.docusaurus.io/docs/2.0.0-alpha.65/theme-classic#navbar-docs-version-dropdown) has only **versions dropdown** but not `/versions` link. I think [DSv2-alpha-66](https://v2.docusaurus.io/docs/theme-classic#navbar-docs-version-dropdown) will handle it.

markerikson:
Great!  It's looking fairly close. Only thing I really see atm is that the "features" section should have a white background, not `var(--ifm-blockquote-color)`.  Can you tweak that?

josephMG:
@markerikson Yes, I fixed features section color.

markerikson:
Awright, this looks good enough for now.  Let's merge it, and we can always fix any other issues later.

@josephmg, thank you for doing this! :)

PR #1655, opened 2020-10-30
Doc : Update ownProps example

harshilparmar:
This changes can make more clear understanding to future reader.I hope it helps 👍

PR #1657, opened 2020-10-30
Doc : Update ownProps examples for all versions

harshilparmar:
@markerikson As we discussed in #1655.I have changed Docusaurus's "versioned docs".


markerikson:
Obsolete now that we've dropped versioning.

Issue #1654, opened 2020-10-29
useSelector returns new value on each call even equalityFn returns true


otakustay:
**What is the current behavior?**

1. Open https://codesandbox.io/s/serene-franklin-pjkhu?file=/src/App.js
2. Press button
3. Inspect console log

We have some confusing logs in console like:

```
array equals: true 
selector value changed 
```

This indicates the return value of `useSelector` changes even its `equalityFn` returns `true`.

**What is the expected behavior?**

In `DisplayValueList` component, the value `mustChange` can change on every press, but the next `useSelector` should return a constant value because its selector function returns a static array and its `equalityFn` is set to an array equal function, the console log should be clean.

When the first `useSelector` is removed, the console output is clean as expected, this is even more confusing to me.

markerikson:
This is specifically because you're passing in a new selector function reference on each render pass.  We always re-run the selector if the selector is a new reference, because your code _could_ have changed what it's selecting entirely (ie, going from `state => state.todos` to `state => state.posts`).

If you were to extract the selector function definition outside the component, or memoize creating it, you'll see the same array result being returned.

otakustay:
Sorry but I still think `equalityFn` should work even selector is a new reference. When `equalityFn` is omitted and selector returns a reference equal object, `useSelector` doesn't  trigger a re-render.

In many cases it's hard to keep selector reference equal, mostly when we select state from props:

```js
const post = useSelector(s => {
  const post = s.posts[props.id];
  const author = s.users[post.author];
  return {...post, author};
});
```

I believe when user passes a `equalityFn`, this function should take responsibility to identify state equality between different selectors.

markerikson:
Mmm... I can see this as a potential point, yeah.  Not something I have time to change myself right now, but if you or someone else wants to put up a PR, we can look at it.

nightpool:
Hey @otakustay! Thank you for opening this issue—I just had the same confusing problem, where I assumed that the equalityFn would be applied in both cases! This was what my code looked like:

```js
   const items = useSelector(state => todoIds.map(id => state.entities.todos[id]), shallowEquals);
   useMemo(() => doSomethingExpensiveWith(items), [items]);
```

and I found to my surprise that the `doSomethingExpensive` was being called every time the component rendered! I had assumed that the shallowEquals, which I know is already memoizing the returned value when the store changes, would also be memoizing the returned value when the selector changed.

@otakustay Let me know if you need any help with that patch—would love to get this over the finish line!

[...]

PR #1660, opened 2020-11-07
Reuse latest selected state on selector re-run (#1654)

otakustay:
I'm not sure if it is a simple fix or a potential breaking change, but reuse latest selected state is quite important in preventing components from re-render frequently

### DIFF
--- a/src/hooks/useSelector.js
+++ b/src/hooks/useSelector.js
@@ -34,7 +34,7 @@ function useSelectorWithStoreAndSubscription(
       latestSubscriptionCallbackError.current
     ) {
       const newSelectedState = selector(storeState)
-      // ensure latest selected state is reused so that a curtom equality function can result in identical references
+      // ensure latest selected state is reused so that a custom equality function can result in identical references
       if (
         latestSelectedState.current === undefined ||
         !equalityFn(newSelectedState, latestSelectedState.current)
--- a/src/hooks/useSelector.js
+++ b/src/hooks/useSelector.js
@@ -33,7 +33,16 @@ function useSelectorWithStoreAndSubscription(
       storeState !== latestStoreState.current ||
       latestSubscriptionCallbackError.current
     ) {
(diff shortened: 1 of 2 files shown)


test/hooks/useSelector.spec.js
```diff
@@ -412,6 +412,35 @@ describe('React', () => {
 
           spy.mockRestore()
         })
+
+        it('reuse latest selected state on selector re-run', () => {
+          store = createStore(({ count } = { count: -1 }) => ({
+            count: count + 1,
+          }))
+
+          const alwaysEqual = () => true
```
jgibson02, line 421: Does this also maintain the latest state if the equalityFn is defined within the component or inline?

src/hooks/useSelector.js
```diff
@@ -33,7 +33,16 @@ function useSelectorWithStoreAndSubscription(
       storeState !== latestStoreState.current ||
       latestSubscriptionCallbackError.current
     ) {
-      selectedState = selector(storeState)
+      const newSelectedState = selector(storeState)
+      // ensure latest selected state is reused so that a curtom equality function can result in identical references
```
nightpool, line 37:
```suggestion
      // ensure latest selected state is reused so that a custom equality function can result in identical references
```

test/hooks/useSelector.spec.js
nightpool: Yes—it only cares what the equalityFn returns, not what its identity is

nightpool: However, having a non-pure equalityFn (i.e., one that closes over variables from its surrounding components) could have *very surprising results.

markerikson:
Seems reasonable, I think.

brianathere:
I'm encountering the same issue, and this would be a nice fix. Is there something I can do to help with merging this PR?

markerikson:
Yeah, let's do this.

PR #1663, opened 2020-11-11
Update quick-start.md

sahilpurav:
Updated the documentation with small correction of package version and also added a note about its support to React v17.


timdorr:
Thanks for ticking up that version. React 17 is covered by the `or later` part, so we don't need to call it out explicitly.

sahilpurav:
Make sense! Thanks.

PR #1683, opened 2021-01-26
Use useIsomorphicLayoutEffect in Provider for consistency

speakingcode:
Provider should use `useIsomorphicLayoutEffect` instead of directly calling `useEffect`, for consistency with `useSelector`

### DIFF
--- a/src/components/Provider.js
+++ b/src/components/Provider.js
@@ -1,7 +1,8 @@
-import React, { useMemo, useEffect } from 'react'
+import React, { useMemo } from 'react'
 import PropTypes from 'prop-types'
 import { ReactReduxContext } from './Context'
 import Subscription from '../utils/Subscription'
+import { useIsomorphicLayoutEffect } from '../utils/useIsomorphicLayoutEffect'
 
 function Provider({ store, context, children }) {
   const contextValue = useMemo(() => {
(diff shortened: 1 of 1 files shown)


timdorr:
This isn't a layout effect, so we don't need to elevate its priority to run before DOM manipulation.

markerikson:
Actually, this is something I wanted us to look at changing.

timdorr:
Okie dokie, we can look at it.

markerikson:
Bit of background for context.

Per https://stackoverflow.com/a/65660253/62937 , I recently debugged someone's project that was using `react-blessed` (a React renderer for CLI output), where the UI stopped updating after the first dispatch.

The real problem was that `react-blessed` was using an old version of `react-reconciler` that happened to have a bug and caused effects to run too early.  However, I also noted that:

-`Provider` is still calling `useEffect` instead of `useIsomorphicLayoutEffect`, which is indeed generally inconsistent with our other components
- There's no way to force use of `useLayoutEffect` when running under Node in cases where you might be using an alternate renderer instead of ReactDOM.

So, I was tossing around the idea of tweaking `Provider` to call `useIsomorphicLayoutEffect`, and also expose a util function from `alternate-renderers` to let users force that that point to `useLayoutEffect` in case that matters to them.

speakingcode:
Indeed that was my project. Took me a while to get around to this. I'm working on the API to force useLayoutEffect at the moment and will try to submit a PR today

PR #1686, opened 2021-02-01
Fixing function dependencies for useActions

pedroapfilho:
I've just refactored the function `useActions` on the examples to a more readable way (IMO)

from:

```js
import { bindActionCreators } from 'redux'
import { useDispatch } from 'react-redux'
import { useMemo } from 'react'

export function useActions(actions, deps) {
  return useMemo(
    () => {
      if (Array.isArray(actions)) {
        return actions.map((a) => bindActionCreators(a, dispatch))
      }
      return bindActionCreators(actions, dispatch)
    },
    deps ? [dispatch, ...deps] : [dispatch]
  )
}
```

to:

```js
import { bindActionCreators } from 'redux'
import { useDispatch } from 'react-redux'
import { useMemo } from 'react'

export function useActions(actions, deps = []) {
  return useMemo(() => {
    if (Array.isArray(actions)) {
      return actions.map((a) => bindActionCreators(a, dispatch))
    }
    return bindActionCreators(actions, dispatch)
  }, [dispatch, ...deps])
}
```

### DIFF
--- a/docs/api/hooks.md
+++ b/docs/api/hooks.md
@@ -404,17 +404,14 @@ import { bindActionCreators } from 'redux'
 import { useDispatch } from 'react-redux'
 import { useMemo } from 'react'
 
-export function useActions(actions, deps) {
+export function useActions(actions, deps = []) {
   const dispatch = useDispatch()
-  return useMemo(
-    () => {
-      if (Array.isArray(actions)) {
-        return actions.map((a) => bindActionCreators(a, dispatch))
-      }
-      return bindActionCreators(actions, dispatch)
-    },
-    deps ? [dispatch, ...deps] : [dispatch]
-  )
+  return useMemo(() => {
+    if (Array.isArray(actions)) {
+      return actions.map((a) => bindActionCreators(a, dispatch))
+    }
+    return bindActionCreators(actions, dispatch)
+  }, [dispatch, ...deps])
 }
 ```
 
--- a/docs/api/hooks.md
+++ b/docs/api/hooks.md
@@ -209,8 +209,6 @@ export const App = () => {
 }
 ```
 
-## Removed: `useActions()`
-
 ## `useDispatch()`
 
 ```js
--- a/website/versioned_docs/version-7.2/api/hooks.md
+++ b/website/versioned_docs/version-7.2/api/hooks.md
@@ -91,7 +91,7 @@ import React from 'react'
 import { useSelector } from 'react-redux'
 
 export const CounterComponent = () => {
-  const counter = useSelector(state => state.counter)
+  const counter = useSelector((state) => state.counter)
   return <div>{counter}</div>
 }
 ```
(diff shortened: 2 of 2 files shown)


markerikson:
There's actually a couple issues here:

- First, we use Docusaurus's "versioned docs" feature, so the "current" version of the docs is in `versioned_docs/7.2/` or something like that.  The files in `/docs/` are the "next" version, which isn't what appears by default.  The changes really need to be made over in the versioned folder.
- Second, we should remove that "Removed: `useActions`" header from everywhere.

pedroapfilho:
Good call:

* Should I keep the change on this folder too? Or only on the versioned one?

* I'll remove it, just need to know from where (if on the `docs` too or not)

markerikson:
I think it needs to be updated in both places, probably.

pedroapfilho:
Done @markerikson

docs/api/hooks.md
```diff
@@ -404,17 +402,14 @@ import { bindActionCreators } from 'redux'
 import { useDispatch } from 'react-redux'
 import { useMemo } from 'react'
 
-export function useActions(actions, deps) {
+export function useActions(actions, deps = []) {
```
ku8ar, line 405: according to: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters#evaluated_at_call_time
default parameters are evaluated at call time.  So this code creates new array with each call.

Such approach is not recommended for core libraries (unnecessary memory waste).

markerikson: I don't see how this makes any difference, because typically you're allocating a new array when you call the hook anyway: `useSomeHookWithDeps(arg1, [])`.

ku8ar: Allocating a new hook-deps array in your business code is a consensus between code readability and usability.

But here we are dealing with a library in which we are not sure how it is used. For example, someone can write 100 such hooks in the code. And this additionally creates 100 new arrays.

Therefore, it is much better to create one static array in a singleton:

```
const defaultDeps = [] // we only create a new array once

export function useActions (actions, deps = defaultDeps) {
```

All we want Redux to be the most efficient framework, right? So we should not remove any microoptimizations, that were here in this code (even if they were not added intentionally ;) )

docs/api/hooks.md
```diff
@@ -404,17 +402,14 @@ import { bindActionCreators } from 'redux'
...
-      return bindActionCreators(actions, dispatch)
-    },
-    deps ? [dispatch, ...deps] : [dispatch]
-  )
+  return useMemo(() => {
+    if (Array.isArray(actions)) {
+      return actions.map((a) => bindActionCreators(a, dispatch))
+    }
+    return bindActionCreators(actions, dispatch)
+  }, [dispatch, ...deps])
```
ku8ar, line 412: This is also a micro-optimization removal. Spread operator performs operations every time (even when deps wasn't given).
In some projects, deps are not given to useActions because most of the time they are not needed.

However, we can reduce the additional cost of this refactor (in V8) by replacing the elements in this new array:
```
[dispatch, ...deps] // bad
[...deps, dispatch] // good
```
If you don't believe it, fire up chrome profiler ;)

ku8ar:
For me this PR is incorrect. Title contains word "fixing" and the code only deals with a refactor (which de facto removes some pretty micro-optimizations).

markerikson:
Afraid I don't really see any benefit in this, especially since this is not part of our actual API anyway.

PR #1699, opened 2021-03-23
Automatically install @types/react-redux as a dependency

markerikson:
This PR:

- Adds a hard dependency on `@types/react-redux`, so that any user who installs the main lib automatically gets the types.

This is a trick I saw from Tanner Linsley and Kent C Dodds, and should resolve cases where users forgot to install the typedefs as well.  Given that we're not rewriting in TS until we work on a v8 release, and we aren't planning to bring the typedefs over here any time soon, we might as well make sure they're available all the time.

Also, I tweaked the Prettier config to ignore line endings, because suddenly ESLint claimed that every single line of code in my local environment had the wrong line endings :(

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,5 +1,6 @@
 {
   "semi": false,
   "singleQuote": true,
-  "tabWidth": 2
+  "tabWidth": 2,
+  "endOfLine": "auto"
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -7234,6 +7234,15 @@
         "@types/node": "*"
       }
     },
+    "@types/hoist-non-react-statics": {
+      "version": "3.3.1",
+      "resolved": "https://registry.npmjs.org/@types/hoist-non-react-statics/-/hoist-non-react-statics-3.3.1.tgz",
+      "integrity": "sha512-iMIqiko6ooLrTh1joXodJK5X9xeEALT1kM5G3ZLhD3hszxBdIEd5C75U834D9mLcINgD4OyZf5uQXjkuYydWvA==",
+      "requires": {
+        "@types/react": "*",
+        "hoist-non-react-statics": "^3.3.0"
+      }
+    },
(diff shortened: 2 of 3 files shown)


MBelniak:
@markerikson Why was it added to "dependencies" instead of "devDependencies"? Types are needed only for development. In the product I work on we have a tool that treats "dependencies" as something that potentially could go into the final bundle and it reports @types/react-redux, which of course will not be included in the bundle :/

timdorr:
@MBelniak Please read the post above by Mark. We're favoring developer experience for those types. In 8.0, we'll be written directly in TS and will have types embedded in the package.

MBelniak:
@timdorr devDependencies are installed as well. Anyway, I've just noticed that this was fixed here: https://github.com/reduxjs/react-redux/pull/1766/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519 :) Sorry for disturbing, it will be all fine in the next release :D

markerikson:
@MBelniak :

> devDependencies are installed as well

No, they're not, and that's exactly the point.

Look at our `package.json` file for v7.2.4:

https://github.com/reduxjs/react-redux/blob/v7.2.4/package.json#L57-L97

It includes Babel, Jest, ESLint, and many more.

I can assure you that when you `npm i react-redux`, _none_ of those get installed. Package managers _do not_ install `devDependencies` when you install a library. They install just `dependencies` (and in the case of NPM 7, `peerDependencies` as well, although that's a separate topic).

So, the whole reason we added `@types/react-redux` to `dependencies` is to ensure that when our users `npm i react-redux`, they automatically get the types installed as well.

MBelniak:
@markerikson OK, this is simply my misunderstanding. What I meant is that when you work on react-redux and run `npm install` or `yarn` you'll get devDependencies installed. But, here we talk about the case of installing react-redux and of course devDependencies won't be installed then. You're completely right, thanks for explaining.

PR #1702, opened 2021-03-27
Remove unused useState call in getting-started.md example

maxsynnott:


markerikson:
Yeah, that's left over from me copy-pasting the Redux template project code and then removing the other inputs.

PR #1705, opened 2021-03-30
Import bindActionCreators from redux

timdorr:
Closes #1498

This removes our peer dep on Redux itself. Technically, anything that implements the Redux API/behaviors should work A-OK with React Redux.

This also makes it easy to work with different version of Redux, if we eventually get to releasing 5.0. Less scary warnings from npm/yarn.


timdorr:
BTW, I got rid of the type guards, since [we're already doing that before calling this function](https://github.com/reduxjs/react-redux/blob/754c1059ded0c1ea3a9b6dc1d870e31c22d8c3b7/src/connect/mapDispatchToProps.js#L17). I also removed the function-to-function option, since I don't think that's a valid code path we use anywhere.

markerikson:
Hah, was _just_ about to ask about those two changes :)

Yeah, if we're already confident we've got the right shape, that should be fine.

markerikson:
Oh, did you bump NPM from v6 to v7? I've still been using v6 locally for everything.  I guess at some point I may have to bite the bullet and upgrade that on my machine.

markerikson (approved):

timdorr:
Yeah, I upgraded locally. It's backwards compatible with npm@6, so it shouldn't be a problem. If it starts flapping back and forth, I can revert back until the tooling catches up.

hornta:
I was just about to raise an issue about this, so glad I found out this was done 👍

markerikson:
Tentatively in the next few days?  Juggling a lot of things atm, but I'll see if I can get around to cutting 7.2.4.

PR #1709, opened 2021-04-09
Remove wrapActionCreators

xty:
This is a follow-up to #1705.

After I had a look at the changes and a bit of digging, I came to realize `wrapActionCreators` has not been used anywhere in the codebase since 5.x, so it's really just a dangling/unreachable file that should have been removed a long time ago.

Now, the remaining question is whether we should move `bindActionCreators` from `src/utils/bindActionCreators.js` into `src/connect/mapDispatchToProps.js` since it is now the only place that references `bindActionCreators`. My own take would be to leave it as is.


timdorr:
Thanks!

Issue #1652, opened 2020-10-27
Connect Component's displayName is not working properly


myNameIsDu:
Wrong name for the component that was packaged with Connect in React - Developer-Tools
![image](https://user-images.githubusercontent.com/53305259/97308780-35b5be00-189c-11eb-8650-3a13d94f439a.png)
I think that's the problem

```js
// If we're in "pure" mode, ensure our wrapper component only re-renders when incoming props have changed.
    const Connect = pure ? React.memo(ConnectFunction) : ConnectFunction

    Connect.WrappedComponent = WrappedComponent
    Connect.displayName = displayName
```
Displayname should be a property in the connectFunction
Like this
```js
    ConnectFunction.displayName=displayName
```
![image](https://user-images.githubusercontent.com/53305259/97309136-9ba24580-189c-11eb-8650-939f14f0e9c2.png)

timdorr:
Weird. This does look like a legit issue going back to 7.0.0, but I can't figure it out. 

That definitely fixes the issue for React Dev Tools, but we have tests for the displayName that do pass: https://github.com/reduxjs/react-redux/blob/1df5622da1324320d6a1b2135aeba914f1873078/test/components/connect.spec.js#L1695-L1730

I'm not sure why that isn't translating over to RDT. Perhaps Babel is mangling something? It looks like facebook/react#18026 is related.

myNameIsDu:
Thanks for your answer, I think it should be React - Developer-Tools don't support 'memoize Components' displayname, like this
```js
let Test =  (props) => {
  const dispatch = useDispatch()

    return (
        <div onClick={ () => dispatch({type:'TEST'})}>
            dispatch test------
        </div>
    )
}
const MemoizeTest = React.memo(Test);
MemoizeTest.displayName='memoize(test)'
```
![image](https://user-images.githubusercontent.com/53305259/97389327-fe372800-1914-11eb-824d-c0d1e0842eaf.png)

But the displayname for 'Memoize Components' does exist, so your test case passed

myNameIsDu:
Can I submit a fix for this issue?

timdorr:
Sure, a PR is always appreciated.

PR #1711, opened 2021-04-10
fix:displayName error displayed in development tool

myNameIsDu:
fix:#1652
fixed an issue with incorrect display in developer tools. But I can't think of a good test case. Can you give me some suggestions


timdorr:
I'm not sure if there's a good test case, since React DevTools gets to it via secret internal stuff that isn't a documented API and may change at any time. I'm not super-worried about testing this specifically. 

Thanks for the quick fix!

PR #1712, opened 2021-04-11
Removed Useless Symbol - Docs

PrinceRajRoy:
Found another useless semicolon while reading the docs,
These might be caused by, [prettier (docs)](https://prettier.io/docs/en/rationale.html#semicolons)


timdorr:
Thanks;

PR #1713, opened 2021-04-15
docs: add link to source in getting-started.md

Nasicus:
I had a look at the "An Existing React App" chapter today. I found the code examples great, but was missing the link to the full code. To be honest I only "stumbled" to it via another page later. I think it's helpful if it's linked directly in this chapter.


timdorr:
Thanks!

PR #1714, opened 2021-04-16
chore: Docusaurus with webpack 5

slorber:
Use an unreleased version of Docusaurus to test the webpack5 support works for this site.

See also https://github.com/facebook/docusaurus/pull/4089

Note: I'm not sure to understand how Netlify is set-up, you may want to cache the webpack5 cache with a similar config than https://github.com/reduxjs/redux/pull/4067

### DIFF
--- a/docs/tutorials/connect.md
+++ b/docs/tutorials/connect.md
@@ -2,6 +2,7 @@
 id: connect
 title: 'Tutorial: Connect API'
 hide_title: true
+sidebar_label: 'Tutorial: Connect API'
 ---
 
 # Tutorial: Using the `connect` API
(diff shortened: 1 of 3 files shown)


timdorr:
Poked at some bad front matter for you. Not sure why GitHub is continually asking me to approve your builds, as I'd figure it would remember the first time I say it's good. 

LGTM

slorber:
thanks

we recently added a more strict validation of frontmatter, do you think we should allow empty sidebar_label like before? What was the intention? Also why your first fix didn't work?

timdorr:
It didn't fall back to the title in the first fix. You can check out that revision and build it locally to see for yourself. Might be worth a bug fix. 

As for the expected behavior, I'd say `null` should behave similarly to `undefined`. That feels the most natural to me. I'm not sure why someone would add sidebar_label without filling it in, but that seems like a reasonable fallback.

slorber:
thanks for the feedback :)

Let me know if I can do anything else for this PR

PR #1717, opened 2021-05-06
embed d.ts typings

gugu:
**Why?**
1. DefinitelyTyped is worked fine when typescript was used by few enthusiasts, but now it is enormous and unmaintanable
2. DefinitelyTyped typings are written for every supported version of react-redux, currently it includes typings for v5 and v6
3. @types/react-redux includes their own dependencies, which are not in sync with react-redux

**What I did?**

1. I've copied index.d.ts from @types/react-redux, added "types" key and removed @types/react-redux dependency
2. I've tested that the module locally works file

**What I didn't do**

I didn't copy tests. Current index.d.ts is in tested state and I want to copy tests in a separate PR if you think adding typescript in devDependencies is fine for you


markerikson:
I don't think this is a thing we want to do right now. The typings are fine where they are.

When we work on React-Redux v8, we'll rewrite the codebase in TypeScript at that point.

PR #1726, opened 2021-06-02
useSelector micro-optimization

ku8ar:
benchmark: https://jsbench.me/exkpfyva4v/1

Version without trailing comma and without creating a new func during each rerender is:
- 20% faster on Safari
- 25% faster on Chrome
- 100% faster on Firefox

Zrzut ekranu 2021-06-2 o 23 30 58


dai-shi:
Mind blowing...

image

timdorr:
Thanks, but I'd rather have more readable code when the difference is going to be in nanoseconds.

PR #1733, opened 2021-06-13
test: Bump jest to 27.x

eps1lon:
Switches to modern timers (including missing mocks) which have better support in the latest testing library versions which you'll probably want to use to test React 18.


timdorr:
I just merged in the TL/R update, so maybe that will fix the issue here? I'm not seeing where it's coming from, because I only have the line of the error, not the full trace.

eps1lon:
> I just merged in the TL/R update, so maybe that will fix the issue here? I'm not seeing where it's coming from, because I only have the line of the error, not the full trace.

The react-native preset is incompatible with jest 27. facebook/react has the same issue. I'll do some digging to find out what the status of jest 27 support for the react-native jest preset is.

markerikson:
What's the actual issue this PR is supposed to fix?

eps1lon:
> What's the actual issue this PR is supposed to fix?

Using modern timers (including missing mocks) which have better support in the latest testing library versions which you'll probably want to use to test React 18.

NitinChronext:
I till see the same issue in React-Native with jest 27.1.1, Is resolution? This is really frustating.

`est suite failed to run

    TypeError: Cannot destructure property 'config' of 'options' as it is undefined.

      at Object.getCacheKey (node_modules/@jest/create-cache-key-function/build/index.js:62:12)
      at ScriptTransformer._getCacheKey (node_modules/@jest/transform/build/ScriptTransformer.js:280:41)
      at ScriptTransformer._getFileCachePath (node_modules/@jest/transform/build/ScriptTransformer.js:351:27)
      at ScriptTransformer.transformSource (node_modules/@jest/transform/build/ScriptTransformer.js:588:32)
      at ScriptTransformer._transformAndBuildScript (node_modules/@jest/transform/build/ScriptTransformer.js:758:40)
      at ScriptTransformer.transform (node_modules/@jest/transform/build/ScriptTransformer.js:815:19)`

test/react-native/batch-integration.tsx
```diff
@@ -51,7 +51,10 @@ describe('React Native', () => {
     return action.type === 'APPEND' ? prev + action.body : prev
   }
 
-  afterEach(() => rtl.cleanup())
```
eps1lon, line 54: This is automatically handled by the auto-cleanup from `@testing-library/react-native`. We just need to make sure we restore the original timers now.

markerikson:
Did finally switch to Jest 29 over in the `v9.0-integration` branch

PR #1734, opened 2021-06-13
Bump @testing-library/react to 11.2.7

eps1lon:
Updating to its latest version for a clearer picture when we a new major gets released.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,7 +5,7 @@
   "requires": true,
   "packages": {
     "": {
-      "version": "7.2.3",
+      "version": "7.2.4",
       "license": "MIT",
       "dependencies": {
         "@babel/runtime": "^7.12.1",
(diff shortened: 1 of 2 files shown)


package-lock.json
```diff
@@ -5,7 +5,7 @@
   "requires": true,
   "packages": {
     "": {
-      "version": "7.2.3",
+      "version": "7.2.4",
```
eps1lon, line 8: I rarely use `npm`. I don't know why it would change unreleated things in the lockfile when running `npm install @testing-library/react@latest`

timdorr: It's cleaning up a bunch of sub-dependencies that are no longer unique to a package (e.g., `foo` requiring `bar@1.2.x` and `baz` requiring `bar@1.1.x` but then being upgraded to support 1.2.x).

PR #1736, opened 2021-06-21
bindActionCreators use in hooks

jiangkangkang12345:


markerikson:
Filing random PRs without any discussion, or even any explanation of what this is supposed to do, is not helpful :(

Per the hooks docs, this is not a thing we want to add:

https://react-redux.js.org/api/hooks#recipe-useactions

Issue #1737, opened 2021-06-22
[v8] Port React-Redux to TypeScript


markerikson:
React-Redux is currently written in plain JS, and our types have been maintained by the community over in DefinitelyTyped at https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/react-redux/ .

Whenever we get around to writing a React-Redux version 8, we intend to port the library to be written in TypeScript directly.

While we don't yet have plans to work on a new version, it's likely that might happen in the next few months because of React 18 and the Concurrent Rendering changes in that release.

It would be helpful if we could start the work of porting React-Redux to TS now, even if it's just a draft on a branch, so that we have that complex work done ahead of time and can then look at any compat changes for React 18 separately without trying to do both at once.

We maintainers probably don't have the time to tackle this ourselves right now, but I'd love to see the community work on that port so we get it ready.

chuckdries:
I'm down to help! If no one else has done so by the time I get off work today, I'll spin up a branch and at least configure typescript so we can start working.

markerikson:
Tell you what. In the interest of keeping collaboration centralized, I'm going to go ahead and create an integration branch here in the main repo.  That way we can have PRs and code progress in one place.

markerikson:
Okay, I just created a new https://github.com/reduxjs/react-redux/tree/typescript-port branch .  Let's use that as the target for all PRs for this effort.

A couple points to consider:

- The goal for now is to exactly match current React-Redux v7.2.x behavior in both logic and TS.  In other words, we're not trying to add any new features or change any TS types, _for now_.  I want an exact match of current functionality as a baseline. _Then_ we can use that as a starting point for any future logic or types changes.  That does mean porting over all the ridiculously complicated helper types and overloads that are currently in the React-Redux typedefs.
- We might want to consider pulling over the ESBuild-powered build toolchain that we recently set up for Redux Toolkit.  It's slightly less critical here because we aren't going to have multiple entry points the way we now have in RTK, but it's something that we know works and is probably less complicated than using TSDX or similar.  See [the `scripts` folder in the RTK repo](https://github.com/reduxjs/redux-toolkit/tree/583823474d041773229a8e78b2ceba37f7616851/scripts) for the current source of the build toolchain.

vinay-deshmukh:
@markerikson 

Is there any plan / structure on how contributors should start converting .js to .ts? 
For example, first the `hooks` folder, then the `connect` folder.

[...]

PR #1739, opened 2021-06-25
Integrate TypeScript port

markerikson:
This is an ongoing integration branch for the TypeScript port work, per #1737 .


markerikson:
Awright. Tests aren't fully converted, but it's time to get this into `master`.

The `compressed-size` action is suddenly failing due to some kind of a Git-related issue, but that's not a blocker.

Merging!

PR #1741, opened 2021-06-25
Bump @testing-library/react to 12.0.0

eps1lon:
[`@testing-library/react@12.0.0` release](https://github.com/testing-library/react-testing-library/releases/tag/v12.0.0) includes a list of breaking changes. Seems like you're unaffected.

The next major will use concurrent rendering by default if React 18 is installed so I want to make sure any test failures are only happening due to switching to concurrent rendering.


timdorr:
Thanks!

PR #1742, opened 2021-06-26
Convert Context/Provider/Subscription to Typescript

Vannevelj:
Building on #1738, this converts `Context`, `Provider` and `Subscription` to Typescript.

Main things of note:
* `undefined` over `null` as it is more idiomatic in Typescript (able to use `name?: Type` as opposed to `name: Type | null`) -- only in two places though
* I've left `store` untyped as `any` for now since I've struggled with that type in the past, though happy to change it if you have a suggestion
* ~~Typescript is flagging some unreachable code, more info down the PR~~

I'm getting an error "Cannot find module 'babel-plugin-polyfill-corejs2'" when running any npm command so I haven't been able to run tests locally -- if this issue rings a bell then let me know what I should do.

It's a small PR to start with as I wanted to make sure we're aligned on the approach that should be taken.


src/utils/Subscription.ts
```diff
@@ -57,6 +61,7 @@ function createListenerCollection() {
         isSubscribed = false
 
         if (listener.next) {
+          //@ts-expect-error -- listener.next is always null
```
Vannevelj, line 64: Because `listener` is initialized on line 47 with `next: null`, Typescript interprets it as the following structure:

image


However when it gets to this line, it causes an error for potentially accessing this `null` field. It doesn't seem to realise that `listener.next` would prevent that from getting accessed.

image

I'm not entirely sure what's going on here -- either Typescript is wrong that `listener.next` is always `null` (though not sure how that would be), or Typescript is wrong that this `null` check isn't sufficient. Is there another option?

Vannevelj: I sorted this by explicitly marking the variable as type `Listener` -- that ensured Typescript correctly interprets it as type `Listener | null` rather than `null`.

rollup.config.js
```diff
@@ -7,7 +7,7 @@ import pkg from './package.json'
 
 const env = process.env.NODE_ENV
 
-const extensions = ['.js', '.ts', '.json']
+const extensions = ['.js', '.ts', '.tsx', '.json']
```
Vannevelj, line 10: I don't know rollup so I'm not sure if this is necessary, but it seemed reasonable

markerikson:
Sweet, more community contributions! :)  Lemme look this over

src/components/Context.ts
```diff
@@ -1,6 +1,11 @@
 import React from 'react'
+import type Subscription from '../utils/Subscription'
 
-export const ReactReduxContext = /*#__PURE__*/ React.createContext(null)
+export type ReduxContextProps = {
+  store: any
```
markerikson, line 5: :speech_balloon: Tell you what, let's use my favorite "stub this out" trick for converting stuff to TS.

~~Let's add a `type $FixTypeLater = any` somewhere, and use that here. That way we know this is something that needs to be addressed down the road.~~  

Or, actually... the TS typedefs for React-Redux should already have the right-ish types that we need, and they specifically reference the core Redux `Store` type here:

https://www.unpkg.com/browse/@types/react-redux@7.1.16/index.d.ts

```ts
export interface ReactReduxContextValue<SS = any, A extends Action = AnyAction> {
    store: Store<SS, A>;
    storeState: SS;
}
```


Hmm. Having said that, those types actually seem buggy, because we _don't_ put the `storeState` in there, and haven't seen v6.

Still, we ought to try to reuse those typedefs as much as possible, because a lot of this stuff has been figured out for us already :)  still a good idea to stub things out with `$FixTypeLater` in places, but we should be starting with these types as our baseline.

src/components/Provider.tsx
```diff
@@ -1,10 +1,10 @@
-import React, { useMemo } from 'react'
+import React, { ReactNode, useMemo } from 'react'
 import PropTypes from 'prop-types'
-import { ReactReduxContext } from './Context'
+import { ReactReduxContext, ReduxContextProps } from './Context'
 import Subscription from '../utils/Subscription'
 import { useIsomorphicLayoutEffect } from '../utils/useIsomorphicLayoutEffect'
 
-function Provider({ store, context, children }) {
+function Provider({ store, context, children }: { store: any, context?: React.Context<ReduxContextProps>, children: ReactNode }) {
```
markerikson, line 7: :hand: I've always preferred defining interfaces for component props separately, right above the component definition:

```ts
interface ProviderProps {
  store: $FixTypeLater;
  // etc
}

function Provider({store, context, children}: ProviderProps) {

}
```
```

markerikson (changes requested):
A great start! I'd like to see the PR updated to copy-paste types defined in the React-Redux typedefs, but with any edits needed to be correct (ie, the context value has `subscription` but not `storeState`

src/components/Context.ts
Vannevelj: I've added a new type `FixTypeLater` which is just an alias for `any` -- I'm not sure if I see all that much value in it (since you can just search for usages for `any` as well) but no strong feelings so don't mind adding it.

Good call on the @types -- I didn't think of using those. I've put them in and replaced the `storeState` field with `subscription`. Is this more like what you had in mind?

Vannevelj:
One thing I haven't done in this PR is remove the `PropTypes`:

https://github.com/reduxjs/react-redux/blob/a24b885bde8e68522be4196aa22ee71cfcdc6084/src/components/Provider.js#L37-L47

I don't think it serves any purpose anymore so happy to drop it if you'd like me to.

markerikson:
Yeah, I'm inclined to agree - I don't think they provide enough benefit at this point, and generally TS types for props have replaced the use of PropTypes from what I've seen.  Go ahead and remove them.

src/components/Context.ts
markerikson: Yep!

markerikson (approved):

markerikson:
Looks great, merging!

PR #1745, opened 2021-06-29
chore(mapStateToProps): port to typescript

tony-go:
Hi folks 👋 

⚠️ : Lot of `any` there. I have to make a second pass to add generics. But feel free to give me feedback anyway.

This pull request follows the typescript migration initiative (#1737).

files to migrate:
- `mapStateToProps`
- `wrapMapToProps`

### DIFF
--- a/src/connect/wrapMapToProps.ts
+++ b/src/connect/wrapMapToProps.ts
@@ -3,18 +3,23 @@ import { Dispatch } from 'redux'
 import { FixTypeLater } from '../types'
 import verifyPlainObject from '../utils/verifyPlainObject'
 
-export type MapToProps = {
-  (stateOrDispatch: FixTypeLater, ownProps?: unknown): FixTypeLater
-  dependsOnOwnProps: boolean
+type AnyState = { [key: string]: any }
+type StateOrDispatch<S = AnyState> = S | Dispatch
+
+type AnyProps = { [key: string]: any }
+
+export type MapToProps<P = AnyProps> = {
+  (stateOrDispatch: StateOrDispatch, ownProps?: P): FixTypeLater
+  dependsOnOwnProps?: boolean
 }
 
 export function wrapMapToPropsConstant(
   // * Note:
-  //  It seems that the dispatch identifier here
-  //  could be dispatch in some cases (ex: whenMapDispatchToPropsIsMissing)
-  // and state in some others (ex: whenMapStateToPropsIsMissing)
+  //  It seems that the dispatch argument
+  //  could be a dispatch function in some cases (ex: whenMapDispatchToPropsIsMissing)
+  //  and a state object in some others (ex: whenMapStateToPropsIsMissing)
   //
-  getConstant: (dispatch: Dispatch) => FixTypeLater
+  getConstant: (dispatch: Dispatch) => { dispatch?: Dispatch }
 ) {
   return function initConstantSelector(dispatch: Dispatch) {
     const constant = getConstant(dispatch)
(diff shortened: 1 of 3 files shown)


src/connect/wrapMapToProps.ts
```diff
@@ -1,7 +1,16 @@
+import { Dispatch } from 'redux'
+
 import verifyPlainObject from '../utils/verifyPlainObject'
 
-export function wrapMapToPropsConstant(getConstant) {
-  return function initConstantSelector(dispatch, options) {
+export type MapToProps = {
+  (stateOrDispatch: any, ownProps?: any): any
```
markerikson, line 6: :hand: Let's replace all uses of `any` with either `type $FixTypeLater = any`, or `unknown`. I want it to be clear these are placeholders and not final

markerikson:
Why did `package-lock.json` change in this PR, and so massively? That doesn't seem right.

src/connect/wrapMapToProps.ts
```diff
@@ -1,7 +1,16 @@
...
+  dependsOnOwnProps: boolean
+}
+
+export function wrapMapToPropsConstant(
+  getConstant: (dispatch: Dispatch, options: any) => void
```
markerikson, line 11: :hand: Hmm. Despite the variable name, I don't _think_ what's being passed in here is actually `dispatch` all the time.  This gets called twice in the `mapDispatch` file and once in `mapState`:

`mapState`: 

```js
export function whenMapStateToPropsIsMissing(mapStateToProps) {
  return !mapStateToProps ? wrapMapToPropsConstant(() => ({})) : undefined
}
```

`mapDispatch`: 

```js
export function whenMapDispatchToPropsIsMissing(mapDispatchToProps) {
  return !mapDispatchToProps
    ? wrapMapToPropsConstant((dispatch) => ({ dispatch }))
    : undefined
}

export function whenMapDispatchToPropsIsObject(mapDispatchToProps) {
  return mapDispatchToProps && typeof mapDispatchToProps === 'object'
    ? wrapMapToPropsConstant((dispatch) =>
        bindActionCreators(mapDispatchToProps, dispatch)
      )
    : undefined
}
```

Based on that, it looks like what gets passed in is actually `state` when we're dealing with `mapState`, and `dispatch` when we're dealing with `mapDispatch`.  Also, there _aren't_ any `options` being passed in or used anywhere.

On the other hand, for the `mapState` case, we ignore what's being passed as the first arg anyway.

We can maybe leave `dispatch: Dispatch` for now, but I think we can remove `options` entirely.

markerikson (changes requested):

timdorr:
> Why did `package-lock.json` change in this PR, and so massively? That doesn't seem right.

Looks like the lockfile on master was updated to v2, but not on the TS branch. I just fixed that up now.

src/connect/wrapMapToProps.ts
tony-go: Thanks a lot for the feedback @markerikson 🙌 

So, I finally use `unknown` and `FixTypeLater`, let me know if it make sense for you. 

I noticed that other functions have a `stateOrDispatch` argument. So it make a lot of sense.

tony-go: When I step back a bit I found that there are too much `FixTypeLater`, WDYT ? I can dig a bit more (using break points) to better understand what's happening there.

markerikson: If you'd like to spend some time further analyzing exactly what's being passed through, sure, go ahead :)

tony-go:
Not sure, but feel like I'm reaching something.

Let me know.

markerikson:
Ought to be good for now. Thanks for working on this!

PR #1746, opened 2021-06-29
Convert reactBatchedUpdates to TypeScript

AlexMunoz:
Hello!
here is my small contribution to https://github.com/reduxjs/react-redux/issues/1737. 
I added ts-ignore comments to ignore the missing types.


markerikson:
Per the `@types/react-dom` package, the types for `unstable_batchedUpdates` are:

```ts
export function unstable_batchedUpdates<A, B>(callback: (a: A, b: B) => any, a: A, b: B): void;
export function unstable_batchedUpdates<A>(callback: (a: A) => any, a: A): void;
export function unstable_batchedUpdates(callback: () => any): void;
```

Can we add those into a types file here somewhere, and at least use an `as` cast or something to force both of those files to use that type?

markerikson:
hmm. something's not right with that last push - now it's showing all the other changes from the `typescript` branch that are already _in_ that branch :)

AlexMunoz:
Hi @markerikson, I just fixed it!

I added the type definitions for unstable_batchedUpdates using a declaration file.

markerikson:
Cool. I'm not sure how having declarations of other modules will end up affecting the typings we output from the build process, but let's get this in now and worry about that later :)  Thanks!

PR #1747, opened 2021-06-30
feature/rename unsafe lifecycles

PepijnSenders:
We're still running react-redux@v4 to support some pages that use redux-form ( 😢, notoriously hard to upgrade ). Could we maybe create a release of v4 that has the unsafe lifecycles prefixed with `UNSAFE_`? Maybe a next minor 4.5?

In this PR I ran the codemod on `src/` let me know if we can do something like this, otherwise we'll patch locally.

### DIFF
--- a/src/components/Provider.js
+++ b/src/components/Provider.js
@@ -35,7 +35,7 @@ export default class Provider extends Component {
 }
 
 if (process.env.NODE_ENV !== 'production') {
-  Provider.prototype.componentWillReceiveProps = function (nextProps) {
+  Provider.prototype.UNSAFE_componentWillReceiveProps = function (nextProps) {
     const { store } = this
     const { store: nextStore } = nextProps
 
--- a/src/components/connect.js
+++ b/src/components/connect.js
@@ -363,5 +363,5 @@ export default function connect(mapStateToProps, mapDispatchToProps, mergeProps,
     }
 
     return hoistStatics(Connect, WrappedComponent)
-  };
+  }
 }
--- a/src/components/connect.js
+++ b/src/components/connect.js
@@ -212,7 +212,7 @@ export default function connect(mapStateToProps, mapDispatchToProps, mergeProps,
         this.trySubscribe()
       }
 
-      componentWillReceiveProps(nextProps) {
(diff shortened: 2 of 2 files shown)


markerikson:
Hey, thanks for the PR. 

Putting out a 4.x minor/patch would be potentially doable, especially given the small changes here and the fact that you took the time to make them.

That said, I'm _really_ curious: what's preventing you from at least upgrading to React-Redux v5?  That came out in late 2016, and did use legacy context.

Just off the top of my head, I would _think_ you should be able to upgrade from v4 to v5 without any issues, but I'm also not clear on exactly what work Redux-Form is doing internally - I think I remember seeing it reaching into `this.context.store` to grab the Redux store instance.

PepijnSenders:
@markerikson trust me I tried! There's some really hacky stuff happening inside that redux-form version that we use. They don't use connect but use the child context values from the library directly.

markerikson:
Yeah, that's actually what I was getting at. I would expect that it's mostly just accessing `this.context.store`, which I _think_ is the same in both 4.x and 5.x.

Just for kicks, can you point me to the exact version of Redux-Form you're using?

timdorr:
When redux-form added 5.x support, they only changed the peer dep requirement: https://github.com/redux-form/redux-form/pull/2631

While it will complain on install about the mismatch, you can safely install 5.x alongside redux-form versions that only explicitly support 4.x.

PepijnSenders:
@timdorr yea our version also allows 5.x, but it breaks because of the adding of the custom store key for us. We can probably fix it, but this code that uses it is very legacy and would rather silence the warnings than deep diving into what's going wrong with redux-form.

PepijnSenders:
@timdorr any chance we can release this?

3imed-jaberi (commented):
I don't think that we need to release this PR. we need to update `package.json` on this line [here](https://github.com/reduxjs/react-redux/blob/4.x/package.json#L99) by making sure that the v4.x is supported by all versions lower than `16.9.0`.

markerikson:
Closing this as we don't meaningfully support 4.x any more.

PR #1749, opened 2021-06-30
Refactor: move d.ts file

AlexMunoz:
Hello @markerikson 
Relating to your last comment  https://github.com/reduxjs/react-redux/pull/1746#issuecomment-871041574, i have made a small change to avoid any possible problems. The custom types for the external dependencies will be outside of src and they will be ignored from the build process.

I'm not so sure if it's worth having these custom types, maybe it would be easier to install the @types/* of these dependencies. What do you think?

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -13,6 +13,6 @@
     "outDir": "./lib",
     "forceConsistentCasingInFileNames": true
   },
-  "include": ["src/**/*"],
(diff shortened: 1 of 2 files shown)


timdorr:
Yeah, I'd rather have it here. Thanks!

markerikson:
I'm fine with just picking these individual types for now. We can look at pulling in the complete types files later if needed.

PR #1751, opened 2021-07-03
Convert package management to Yarn v2

markerikson:
This PR:

- Switches the React-Redux package management process to use Yarn v2
  - Added Yarn "binary" and config file, and added a lockfile
  - Updated gitignore
  - Marked the `website` folder as a workspace
  - Switched the build steps to use Yarn
  - Removed the NPM lockfiles


timdorr:
I'm not familiar, but why does `.yarn/releases/yarn-berry.cjs` exist? That looks to be yarn itself, which I would assume to be redundant to your own local version.

markerikson:
This is actually how Yarn v2 is designed to work: installations are per-project, and the Yarn "binary" is supposed to be committed to the repo.

Yarn v1 has to be installed globally, and then you have a config file in the repo that says "use v2".  When you run `yarn` in the repo folder, v1 starts, looks at the config file, sees "v2", and bootstraps the execution of v2 from there.

It's admittedly kinda goofy, but it works.

.yarnrc.yml
```diff
@@ -0,0 +1,9 @@
+nodeLinker: node-modules
+
+plugins:
+  - path: .yarn/plugins/@yarnpkg/plugin-workspace-tools.cjs
+    spec: '@yarnpkg/plugin-workspace-tools'
+  - path: .yarn/plugins/@yarnpkg/plugin-compat.cjs
+    spec: '@yarnpkg/plugin-compat'
```
timdorr, line 7: This shouldn't be needed if we're using node-modules above. The PNP linker is disabled, so this package shouldn't have any effect.

PR #1753, opened 2021-07-05
Add first version of types for useSelector hook

lindskogen:
Based on type definitions from https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-redux/index.d.ts#L617


markerikson:
Hah, this is hilarious! At literally the same time you filed this PR, I sat down and started trying to convert `useSelector` to TS myself, and I just got it working :)

Our changes are almost identical, for obvious reasons, but your version looks a bit more rigorously defined than what I'd done, so I'm happy to go with this instead.

lindskogen:
Oh sorry, I will ping you next time I start working on something 😅

PR #1754, opened 2021-07-06
Update Babel config for modern output

markerikson:
- Switched babelrc to use recommended settings for "modern JS", with `targets: esmodules` and `bugfixes: true`
- Fixed missing .tsx extensions in build scripts
- Silenced Rollup warning about missing flag

### DIFF
--- a/.babelrc.js
+++ b/.babelrc.js
@@ -1,17 +1,25 @@
 const { NODE_ENV, BABEL_ENV } = process.env
 const cjs = NODE_ENV === 'test' || BABEL_ENV === 'commonjs'
-const loose = true
 
 module.exports = {
   presets: [
-    ['@babel/env', { loose, modules: false }],
+    [
+      '@babel/preset-env',
+      {
+        targets: {
+          esmodules: true,
+        },
+        // Use the equivalent of `babel-preset-modules`
+        bugfixes: true,
+        modules: false,
+      },
+    ],
     '@babel/preset-typescript',
   ],
   plugins: [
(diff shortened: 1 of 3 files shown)


codeonquer:
Hi，I'm confused with the code in `.babelrc.js`:

```js
...
targets: {
  esmodules: true,
},
...
```

The code will keep `let`/`const` syntax in `dist/react-redux.js`/`dist/react-redux.min.js`/`es/*.js`/`lib/*.js`. If I want to use react-redux, I have to transform the code of react-redux by Babel again (because you already transform it by Babel)。If you want to keep `import/export` syntax, I think `modules: false` in `.babelrc.js` is enough.

I wonder What is the perpose of this. Hope I didn't offend you.

phryneas:
What is the problem with `let` and `const`? Those are part of ECMAScript2015, which is 7 years at this point. Even with other targets than `esmodules` we are targeting more modern browsers.

codeonquer:
Thanks for your answer, I get it.

Because of the users of our products, I have to support iOS9. When I use react-redux, I config the webpack like this as usual:

```js
{
  test: /\.(ts|js)x?$/,
  exclude: /node_modules/
}
```

This doesn't work on iOS9.

Then I find that redux's `targets` is `browsers: ['ie >= 11']`, so i am confused with the difference between redux and react-redux.

Thanks for your awesome work.

codeonquer:
Thanks for your detailed answers! I will transpile myself. ^-^

PR #1758, opened 2021-07-06
Convert connect components and logic to TS (with caveats)

markerikson:
This PR:

- Converts `connectAdvanced`, `connect`, and `selectorFactory` to TS
- Removes the deprecated argument warnings left over from the v6-v7 transition (ie, "withRef is not an option any more", etc)

These types are **absolutely not final!**  For one thing, I guarantee this isn't actually inferring the right props and carrying them through.  That's going to take a lot more work (and probably from someone _other_ than me).

Also, I had to slap a bunch of `unknown` and `// @ts-ignore` everywhere.  Some of those can maybe stay, because they're internal implementation details that aren't worth fussing over. Others probably ought to be fixed.

But, I've at least managed to hack together enough types to get the files compiling without errors and sorta-mostly look correct-ish.

Ref: #1737


src/components/connectAdvanced.tsx
```diff
@@ -1,44 +1,63 @@
...
+  state: [payload: unknown, counter: number],
+  action: { payload: unknown }
+) {
   const [, updateCount] = state
   return [action.payload, updateCount + 1]
 }
 
+type UseLayoutEffect = typeof useLayoutEffect
+type EffectFunc = (...args: any[]) => void | ReturnType<React.EffectCallback>
+type EffectDeps = Parameters<UseLayoutEffect>[1]
```
lindskogen, line 43: There is a `DependencyList` type exported from `react` if you'd like to use that instead!

markerikson:
Well, let's get this in, and we can review the types later.

PR #1759, opened 2021-07-07
Convert entry file to ts

myNameIsDu:
-  I added the entry file export type

- I reviewed @type/react-redux. Except for the type definitions of functional methods such as connect, the others all exist, but it may need to be checked again

- Slightly modified connect and connectAdvanced. Note that in @types/react-redux connectOptions are called options. ConnectAdvancedOptions is called ConnectOptions, I think this may not be great, so I changed it

-  I did not find the role of alternate-renderers.js, so I haven't changed it yet

Ref: #1737


markerikson:
Cool, this looks like it's pretty good for now.

FYI, `alternate-renderers` is the same as `index.{js,ts}`, but it _doesn't_ turn on React's batched updates for our subscriptions. That's because `unstable_batchedUpdates` only exists for ReactDOM and React Native for the most part - other renderers like `ink` don't have that.

We could potentially put all of our exports into something like `exports.ts`, and then have `index.ts` just do:

```ts
export * from './exports'

import { unstable_batchedUpdates as batch } from './utils/reactBatchedUpdates'

setBatch(batch)
```

and `alternate-renderers.ts` would be just be `export * from './exports'`

that would at least ensure things don't get out of sync.

PR #1765, opened 2021-07-09
Refactor tests to Typescript

### LINKED ISSUE #1737: [v8] Port React-Redux to TypeScript

StevenLangbroek:
I'm funemployed and saw this was still WIP, so thought I'd jump in :) (yes I'm bored).

### Todo

- [x] Add dependencies
- [ ] Convert non-react-native tests
- [ ] Configure Jest for react-native
- [ ] Convert react-native tests

Closes #1737


StevenLangbroek:
@markerikson I think I may have found some issues with overloading on `connect`:

image

Calling `connect` with just a state mapper is perfectly valid right? If so I think args after should all be made optional... Maybe I'm misunderstanding. Would love to discuss :)

StevenLangbroek:
If it shouldn't be allowed from a type-safety perspective but we want to test what happens anyway I can `@ts-expect-error` these.

markerikson:
I would assume at this point that any types failures are actually bugs in the current types in this branch :)

I'm gonna say mark types errors as `// @ts-ignore` for now. We know the tests _should_ pass at runtime, and we know our types are almost definitely buggy.  Let's focus this effort on getting the tests themselves passing, and look at the types separately.

StevenLangbroek:
@markerikson sure thing. I'll add as specific a ts-expect-error pragma as I can so we're not swallowing unknown mismatches

PR #1767, opened 2021-07-12
Refactor/test hooks

myNameIsDu:
- add ts-jest and modify jest config

- transform test/hooks js files to tsx

ref: https://github.com/reduxjs/react-redux/issues/1737


test/hooks/useSelector.spec.tsx
```diff
@@ -81,17 +101,17 @@ describe('React', () => {
...
 
         it('subscribes to the store synchronously', () => {
-          let rootSubscription
+          let rootSubscription: Subscription
 
           const Parent = () => {
             const { subscription } = useReduxContext()
             rootSubscription = subscription
-            const count = useSelector((s) => s.count)
+            const count = useSelector<DefaultRootState>((s) => s.count)
```
markerikson, line 109: :hand: Standard usage for `useSelector` is to either explicitly declare the type of `s`, or create a "pre-typed" version of the hook, per https://redux.js.org/usage/usage-with-typescript#define-typed-hooks .

Either of those is fine with me, but I'd appreciate changing over to those instead of using a generic in all of these.

test/hooks/useSelector.spec.tsx
```diff
@@ -380,20 +406,30 @@ describe('React', () => {
           const spy = jest.spyOn(console, 'error').mockImplementation(() => {})
 
           const Parent = () => {
-            const count = useSelector((s) => s.count)
+            const count = useSelector<DefaultRootState, number>((s) => s.count)
```
markerikson, line 409: :hand: Similarly, here the return type should be inferred correctly if the state is typed right

test/hooks/useSelector.spec.tsx
```diff
@@ -380,20 +406,30 @@ describe('React', () => {
...
-
-          const ConnectedWrapper = connect(({ count }) => ({ count }))(
-            ({ parentCount }) => {
-              return <Child parentCount={parentCount} />
-            }
-          )
+          interface ParentContentProps {
+            parentCount: number
+          }
+          const ConnectedWrapper = connect<
```
markerikson, line 415: :speech_balloon: here on the other hand, `connect` mostly requires the generics

markerikson (changes requested):

test/hooks/useSelector.spec.tsx
myNameIsDu: Yes, that's a great way to write it

markerikson (approved):

markerikson:
Looks good!

PR #1769, opened 2021-07-15
refactor: transform test/components/hooks to tsx

myNameIsDu:
transform test/components/hooks to tsx

ref: #1737


markerikson:
Thanks!

PR #1770, opened 2021-07-16
chore: modify jest config, ts-jest => 26.xx.xx, tsconfig'include add …

myNameIsDu:
I found some big problems:

- ts-jest does not check the converted type file, the previous configuration was problematic
  
- I have now fixed it, but due to the strict mode of tsconfig, there are some type errors that need to be reworked  :sob:

ref: #1737

### DIFF
--- a/package.json
+++ b/package.json
@@ -109,7 +109,7 @@
     "rimraf": "^3.0.2",
     "rollup": "^2.32.1",
     "rollup-plugin-terser": "^7.0.2",
-    "ts-jest": "^27.0.3",
+    "ts-jest": "26.5.6",
     "typescript": "^4.3.4"
   },
   "browserify": {
--- a/package.json
+++ b/package.json
@@ -110,7 +110,7 @@
     "rollup": "^2.32.1",
     "rollup-plugin-terser": "^7.0.2",
     "ts-jest": "26.5.6",
-    "typescript": "3.9"
+    "typescript": "^4.3.4"
   },
   "browserify": {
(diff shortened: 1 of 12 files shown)


package.json
```diff
@@ -109,7 +109,7 @@
     "rimraf": "^3.0.2",
     "rollup": "^2.32.1",
     "rollup-plugin-terser": "^7.0.2",
-    "ts-jest": "^27.0.3",
+    "ts-jest": "26.5.6",
```
timdorr, line 112: Why is this being downgraded?

myNameIsDu: When I was using ts-jest@27, I encountered some [problems]( https://github.com/kulshekhar/ts-jest/issues/2611#issuecomment-851014389)
Then I upgraded JEST to version 27 and encountered this [problem](https://githubmemory.com/repo/facebook/fbjs/issues/441) again
In the end, I can only use version 26 first

myNameIsDu: In addition,I will fix the problem left over from converting the TS file, But this may take some time, as strict type checking may require a lot of additional fixes:cold_sweat:

markerikson:
Tests are passing, changes look good. Anything else you want to do in this PR, or is it good to merge?

myNameIsDu:
> Tests are passing, changes look good. Anything else you want to do in this PR, or is it good to merge?

I think it is OK to merge, and I will initiate PR in future updates

markerikson:
Great, thanks again!

PR #1771, opened 2021-07-18
Remove connectAdvanced

### LINKED ISSUE #1236: Deprecating connectAdvanced

markerikson:
This PR:

- Consolidates all `connect` component implementation into a single file and function, and removes the `connectAdvanced` function and public export

Fixes #1236 , which announced the deprecation of `connectAdvanced` two years ago.

This is a draft for now, but I intend to merge this after #1737 gets merged into `master`.


markerikson:
Goodbye!

sktguha:
Good riddance 😂🙏

PR #1772, opened 2021-07-20
refactor: transform test/components/Provider => ts

myNameIsDu:
transform test/components/Provider => ts

ref: #1737


markerikson:
Thanks!

PR #1776, opened 2021-07-23
refactor: transform test/integration/server-rendering => ts

myNameIsDu:
transform test/integration/server-rendering => ts

ref: #1737

### DIFF
--- a/test/integration/server-rendering.spec.tsx
+++ b/test/integration/server-rendering.spec.tsx
@@ -8,28 +8,64 @@
 
 /*eslint-disable react/prop-types*/
 
-import React from 'react'
+import React, { FunctionComponent } from 'react'
 import { renderToString } from 'react-dom/server'
 import { createStore } from 'redux'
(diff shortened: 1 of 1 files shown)


test/integration/server-rendering.spec.tsx
```diff
@@ -8,28 +8,64 @@
...
+      state = { greeting: 'Hello' },
+      action: ActionType
+    ) {
       return action && action.payload ? action.payload : state
     }
+    interface GreetingProps {
+      greeting: string
+      greeted: string
+    }
+    const Greeting: FunctionComponent<GreetingProps> = ({
```
markerikson, line 35: :hand: I don't think `FunctionComponent` is actually needed here - just use `GreetingProps` after the destructuring. But we can get this in now and tweak stuff later.

(Then again, we could probably replace a lot of our classes with functions anyway)

markerikson:
Thanks!

PR #1781, opened 2021-07-29
Refactor/test connect

myNameIsDu:
Refactor test connect

link: #1737


markerikson:
Looks okay so far. Lemme know when you have the rest of the logic uncommented and converted!

myNameIsDu:
@markerikson   Because it's a big file, I think it's a good idea for you to check it earlier, and comment out the test cases that are not converted, because they don't pass the check and all of them are temporarily commented out until I convert them little by little, Finally, thank you for giving me valuable writing instructions

myNameIsDu:
> Looks okay so far. Lemme know when you have the rest of the logic uncommented and converted!

ok, thanks

myNameIsDu:
@markerikson I don't know why it uses npm instead and causes an error that triggers yarn, strange

markerikson:
@myNameIsDu looks like commit 8bbc934dcd6648948fd92365c5c21d61563c109d added a `package-lock.json` file.  Delete that and I suspect CSB should be happier.

myNameIsDu:
> @myNameIsDu looks like commit [8bbc934](https://github.com/reduxjs/react-redux/commit/8bbc934dcd6648948fd92365c5c21d61563c109d) added a `package-lock.json` file. Delete that and I suspect CSB should be happier.

Yes, then the default order for the CSB is to check the lock file of npm => yarn  :satisfied: It's an interesting thing, but I don't know much about it

myNameIsDu:
@markerikson  I found some old skip cases, should I delete them?

test/components/connect.spec.tsx
```diff
@@ -2020,6 +2382,7 @@ describe('React', () => {
 
         rtl.render(
           <ProviderMock store={store}>
+            {/*// @ts-ignore */}
```
markerikson, line 2385: :question: _just_ out of curiosity, what issues are we ignoring in this area of the code?

test/components/connect.spec.tsx
```diff
@@ -2241,40 +2665,39 @@ describe('React', () => {
...
-              <Wrapper />
-            </ProviderMock>
-          )
-        ).toThrow(
-          `To access the wrapped instance, you need to specify { withRef: true } in the options argument of the connect() call`
-        )
-        spy.mockRestore()
-      })
-
+      //   it.skip('should throw when trying to access the wrapped instance if withRef is not specified', () => {
```
markerikson, line 2668: :hand: Yeah, this is dead - `getWrappedInstance()` went away in version 6, and I guess we never removed this test.  Delete it.

markerikson (changes requested):
Looks great! My only comments are a question about what TS errors we're ignoring in a few places, and that the `getWrappedInstance` test can be deleted.

test/components/connect.spec.tsx
myNameIsDu: image
This descript is also all skip, I can also delete it directly?

markerikson: yeah, let's delete that one too

test/components/connect.spec.tsx
markerikson: Hmm. Yeah, it's entirely possible that's a bug in the types.

Let's leave the ignore in there for now, and maybe leave a `// TODO` comment saying to re-investigate it later?

myNameIsDu: That's a good idea. I think I can fix it after this PR #1771  is done

markerikson:
Awright, let's get this in.

Thank you so much for working on this!

PR #1786, opened 2021-08-02
refactor: transform react-native => ts & modify jestConfig: delete th…

myNameIsDu:
refactor: transform react-native => ts & modify jestConfig: delete the JS matching pattern

link #1737 
 
there is a problem:

image
In this case, I've deleted the previous argument

### DIFF
--- a/test/react-native/batch-integration.tsx
+++ b/test/react-native/batch-integration.tsx
@@ -12,8 +12,10 @@ import { useIsomorphicLayoutEffect } from '../../src/utils/useIsomorphicLayoutEf
 import * as rtl from '@testing-library/react-native'
 import '@testing-library/jest-native/extend-expect'
 
+import type { MiddlewareAPI, Dispatch as ReduxDispatch } from 'redux'
+
 describe('React Native', () => {
-  const propMapper = (prop) => {
+  const propMapper = (prop: any) => {
     switch (typeof prop) {
       case 'object':
       case 'boolean':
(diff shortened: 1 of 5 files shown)


test/react-native/batch-integration.tsx
```diff
@@ -467,33 +570,35 @@ describe('React Native', () => {
...
         const button = rendered.getByTestId(testID)
         rtl.fireEvent.press(button)
       }
 
-      const clickAndRender = (rendered, testID) => {
+      const clickAndRender = (rendered: RenderedType) => {
         // Note: Normally we'd wrap this all in act(), but that automatically
         // wraps your code in batchedUpdates(). The point of this bug is that it
         // specifically occurs when you are _not_ batching updates!
         clickButton(rendered, 'setTimeout')
```
markerikson, line 591: :hand: Yeah, I think there's actually a bug in this test here. It should be `clickButton(rendered, testID)` - I apparently missed passing that through when I was writing this :)

Can you restore the arguments for this section and fix this line?

markerikson (changes requested):

test/react-native/batch-integration.tsx
myNameIsDu: Okay, no problem

myNameIsDu: Okay, no problem

markerikson:
Looks good, thanks!

PR #1792, opened 2021-08-08
Migrate useSelector and connect to useMutableSource

wurstbonbon:
This very rough draft to get the ball rolling and probably too naive. I branched away from `remove-connectadvanced` to be closer to the final API. This is meant to be a starting point, feel free to ignore it if it doesn't provide any value. Many tests are failing and some of them hopefully give a hint to where things will become problematic.

React testing library [does not support React 18 yet](https://github.com/testing-library/react-testing-library/pull/937) and to be able to run the tests one needs to [download the package](https://ci.codesandbox.io/status/testing-library/react-testing-library/pr/937) for the PR and install it locally because yarn2 doesn't like URLs without file extension.


markerikson:
Neat, thanks! Good to have something to use as a starting point - appreciate you doing this!

It'll be a while before I have time to fully dig through this. My focus atm is on trying to write some additional documentation for the "Redux Essentials" tutorial covering how to use RTK Query.  After that I should have time to start thinking more seriously about React-Redux v8.

markerikson:
FYI, the immediate failure is due to no `/tmp/react.tgz` file existing (which makes sense given that it's running on a random CI VM).

I'd suggest temporarily committing the tarball directly to this branch and referencing that instead.

wurstbonbon:
@markerikson I just figured out the correct URL of package

markerikson:
Well, not sure why the tests didn't run, but at least the size checks passed, which means it built cleanly. Given that this is TS, that's at least a decent first step :)

PR #1798, opened 2021-08-22
Fix connect.md

38elements:
I read this line.  It seems that a wrapper component rely on its props and the selected Redux store’s state.
https://github.com/reduxjs/react-redux/blob/d4e4eba9ccbd488b103b3c5625a37e15b1427d11/src/components/connectAdvanced.tsx#L479

I think a wrapped component rely on its state.


timdorr:
Thanks, but this is fine either way. It's a "component that will be wrapped", and is a "wrapper component".

38elements:
Thank you for reply.

PR #1799, opened 2021-08-22
Update hooks.md with multiple component selector

hotyhuang:
Hi, I got a little confused when reading about this code... I believe the `makeSelectCompletedTodosCount` here is the selector factory, and `selectCompletedTodosCount` is a selector instance created by the factory.

LMK if my understanding is wrong.

### DIFF
--- a/docs/api/hooks.md
+++ b/docs/api/hooks.md
@@ -206,7 +206,7 @@ const makeSelectCompletedTodosCount = () =>
   )
 
 export const CompletedTodosCount = ({ completed }) => {
-  const selectCompletedTodosCount = useMemo(makeSelectCompletedTodosCount, [])
+  const selectCompletedTodosCount = useMemo(makeSelectCompletedTodosCount(), [])
 
   const matchingCount = useSelector((state) =>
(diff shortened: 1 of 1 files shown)


markerikson:
Appreciate the PR, but the current code is correcct as-is - it's just abbreviating the usage here.

What we have now is:

```js
const selectCompletedTodosCount = useMemo(makeSelectCompletedTodosCount, [])
```

`useMemo` accepts a callback function (kind of like another "factory" in a sense). It will call back and save the return value for later use.

Here, we're passing `makeSelectCompletedTodosCount` as the "factory" directly to `useMemo`.

This is exactly the same as if we wrote:

```js
useMemo(() => {
  return makeSelectCompletedTodosCount()
}, [])
```

We're just eliminating the outer arrow function that does nothing except call `makeSelectCompletedTodosCount`.

If we did this:

```js
useMemo(makeSelectCompletedTodosCount(), [])
```

we would actually be calling `makeSelectCompletedTodosCount` too early, and passing _the new selector function_ as the argument to `useMemo`, which wouldn't be correct.

PR #1800, opened 2021-08-22
"pure" component -> "pure" function in connect.md

38elements:
Relationship between wrapper component and wrapped component `connect.md`.

```js
const WrapperComponent = connect(mapStateToProps, actionCreators)(WrappedComponent)
```

I read this lines.

https://github.com/reduxjs/react-redux/blob/d4e4eba9ccbd488b103b3c5625a37e15b1427d11/src/components/connectAdvanced.tsx#L277-L480

> Assumes that the wrapped component is a “pure” component and does not rely on any input or state other than its props and the selected Redux store’s state.

* [PureComponent](https://reactjs.org/docs/react-api.html#reactpurecomponent) rely on props and state. This is inconsistent with "does not rely on any input or state". Therefore, I think [pure function](https://en.wikipedia.org/wiki/Pure_function) is correct, not pure component.

* A wrapper component does not have its state. A wrapped component may have its state and may rely on its state. This is inconsistent with "does not rely on any input or state".  Therefore, I think wrapper component is correct, not wrapped component.

### DIFF
--- a/docs/api/connect.md
+++ b/docs/api/connect.md
@@ -230,7 +230,7 @@ connect(mapStateToProps, mapDispatchToProps, null, { context: MyContext })(
 
 - default value: `true`
 
-Assumes that the wrapped component is a “pure” component and does not rely on any input or state other than its props and the selected Redux store’s state.
+Assumes that the wrapper component is a “pure” function and does not rely on any input or state other than its props and the selected Redux store’s state.
 
 When `options.pure` is true, `connect` performs several equality checks that are used to avoid unnecessary calls to `mapStateToProps`, `mapDispatchToProps`, `mergeProps`, and ultimately to `render`. These include `areStatesEqual`, `areOwnPropsEqual`, `areStatePropsEqual`, and `areMergedPropsEqual`. While the defaults are probably appropriate 99% of the time, you may wish to override them with custom implementations for performance or other reasons.
 


timdorr:
I'm sorry, but this is still correct as-is. The component can be a class component or function component, it doesn't matter what form it takes, just that is is "pure". And the wrapper/wrapped thing still works either way.

38elements:
@timdorr 
Thank you for reply. May I ask you a question?  When a wrapped component which is functional component with useState or class component changes its state, the component' s rendering result is changed by its state. Is this case "wrapped component rely on its state "?

timdorr:
It's the same definition as PureComponent, because that's all we're doing there: making the WrapperComponent for the wrapped component extend PureComponent instead of Component.

38elements:
I think PureComponent rely on its state.
https://reactjs.org/docs/react-api.html#reactpurecomponent

markerikson:
FWIW, we've never used `PureComponent` internally. In v5 and v6, it was class components with `shouldComponentUpdate`. As of v7, we use `React.memo()`, same as you would if you were wrapping any other function component yourself for perf:

https://github.com/reduxjs/react-redux/blob/v7.2.4/src/components/connectAdvanced.js#L473-L474

38elements:
> The word "pure" here specifically means "only depends on props and state", rather than potentially reading from some mutating external value directly.

Thanks.

I look like that wrapped(inner) component's state is ignored in existing description.

markerikson:
I'm... not sure what point you're trying to make here.

The current wording of the doc is current as-is.

`pure` mode assumes that the wrapped/inner component only needs to re-render when its incoming props change because it doesn't care about any other external mutative data sources, thus we optimize the behavior of the wrapper/outer component to make sure that happens.

The _state_ of the wrapped component is irrelevant to `connect`.

38elements:
I feel uncomfortable with the following part.
"wrapped component is a “pure” component and does not rely on any input or state"
wrapped/inner component rely on state.

markerikson:
You're really over-worrying about this :)

Realistically, there's no need to change anything here. `connect` is no longer the default API - hooks are.  Besides, `pure` is the default option here, and no one actually modifies this option these days anyway - the only reason it's there is because of weird usages in the early days of React and React-Redux.

38elements:
@timdorr @markerikson 
I am very sorry. I have understood your idea. I think your idea is following.
"Since it assumes that the wrapped component is a “pure” component and does not rely on any input or state other than its props and the selected Redux store’s state, wrapped component is memorized."

PR #1804, opened 2021-08-24
Reduce unnecessary calls to useSelector selector

sufian-slack:
This change is intended to fix https://github.com/reduxjs/react-redux/issues/1801 on the 8.x line

By checking if the store state has differed prior to recalculating a selector, we can avoid unnecessary selector recalculation in most cases for components on mount.


markerikson:
Looks great! Let's do this.  

(There's a very good chance this will get immediately replaced as I try migrating over to the new `useSyncExternalStore` API, but this is a good starting point, and the tests are still valuable)

PR #1808, opened 2021-09-03
Initial experimental React 18 compat prototyping

markerikson:
This PR:

- Updates our React devDeps to `18.experimental` versions, along with React Testing Library's PR build from https://github.com/testing-library/react-testing-library/pull/937
- Updates our TS config to use the types for those React builds
- Fixes almost all of the failing tests, which were primarily due to state updates being triggered outside of `act()`
- Actually swaps `useSelector` over to use `useSyncExternalStore` instead of our own subscription logic
- Also updates `connect` to use `useSyncExternalStore`, albeit in a kinda sketchy way
- Rips out the `pure` option for `connect`

@dai-shi has a proof-of-concept PR at https://github.com/pmndrs/zustand/pull/550 that will likely be useful as a starting point.

## Status

~~At this point I see 3 test failures.  Two look like they're related to tests for "avoid setting state on unmount", and the other looks like it's a batching-related issue.  Need to investigate those further.~~

~~Note that this so far involves _no_ library code changes whatsoever.  Once I get the current tests updated to be passing based on expected behavior with React 18 overall, the next step is to try switching our logic over to use `useSyncExternalState` per https://github.com/facebook/react/pull/22211 , and see what happens.~~

I've just commented out / tweaked those three test failures for now so I can move on. I'll review them more later.

### Update 1

**I just successfully got `useSelector` migrated over to use `useSyncExternalStore`!**  I'm seeing one legit test failure atm - `uSES` is not correctly preserving the previous selected result when you give it a new selector reference, because of how it memoizes things.

I wrote some notes in https://github.com/facebook/react/pull/22211 , and I'll paste those here:

> - React-Redux currently catches errors thrown from selectors inside of subscriptions, saves them in a ref, and forces a re-render. If the selector runs clean while rendering, no problem. If it throws again, we attach the original error's message to the new error just in case it was two different things that happened. However, `uSES` always swallows errors in `checkIfSnapshotChanged`, so we can no longer check the original error ourselves.  I was able to work around this by having `useSelector` wrap the original selector with one that does a `try/catch` and saves errors to a ref, but wow is that hacky.
> - Is the "not a valid React 18 version" check accidentally inverted? It currently says `if (React.startTransition !== undefined) {`, which seems wrong. I'm currently running `18.0.0-alpha-1314299c7-20210901`, which most definitely has a `startTransition` function defined, and I _am_ seeing that warning printed.
> - Our test for "checks for updates when subscribing and re-renders if different" is seeing a _third_ render occur, when it used to only see _two_, and I can't figure out what's causing that extra render to happen (especially since renders 2 and 3 are seeing the same selected state).  My vague hypothesis is that it may have something to do with the test dispatching an action in a `useLayoutEffect` in a child, while the subscription happens in a `useEffect` in the hook in the parent?
> - The way that `useSyncExternalStoreExtra`'s memoization check is implemented means that if you get a new selector function, there's no way for it to be equal to the prior result and reuse it, because all of the memoization variables are internal to the `useMemo` hook.  Our v7 `useSelector` does this behavior by keeping around the last selected value in a `useRef`.  We have a test for that in our codebase, and it's failing atm, and I'm pretty sure this is a "correct" failure based on what I'm seeing in the `uSES` implementation.  

### Update 2

**I've just gotten `connect` updated as well, and I've also ripped out the `pure` option**.  

I'm not convinced that the way I integrated `useSyncExternalStore` into `connect` is at all "correct". `connect` is still doing all the actual child props derivation logic. It's really just deferring to `uSES` for the "trigger a re-render" step.

All of the "impure behavior" tests were failing with this approach.  After thinking about it, this option seemed ripe for removal. It only existed because early React ecosystem usage sometimes still relied on external mutable sources that would be different every time the component re-rendered, so we _always_ had to re-run `mapState` and the props merging logic.  In today's world, that's no longer the case - external mutability is no longer a thing that shows up in codebases, and at this point if you're still doing that it's a bug.

I did run a SourceGraph search to try to find any places where people were actually using `{pure: false}`. Best I could come up with was https://sourcegraph.com/search?q=context:global+lang:JavaScript+connect%28.%2B%2C+.%2B%2C+.%2B%2C+%7B.*pure%5Cs%3F:%5Cs%3Ffalse.*%7D%29&patternType=regexp , which only had 59 hits in 1.2M repos. Of those 59 hits, half were FlowType typedefs.  So, this seems safe to kill off, and it simplifies several other bits of the remaining codebase anyway.

At this point, there's two tests failing, and ironically they're both bugs in `uSES` itself:

- It's not correctly memoizing and reusing the last returned value if it gets a new selector reference
- It's printing a "no `useLayoutEffect` in SSR" warning

Now that I think about it, I do have some tests for `useSelector` that I'd temporarily tweaked and gotten to "pass" by changing the assertions just to let me keep going. I need to review those further.


markerikson:
Hmm. Not sure why the tests aren't running. Created this as a "draft" initially and thought that might be why, but doesn't look like they're running after changing it to a normal PR either.

timdorr:
If you rebase off master, it should have a button for running tests now.

eps1lon:
> Is the "not a valid React 18 version" check accidentally inverted? It currently says if (React.startTransition !== undefined) {, which seems wrong. I'm currently running 18.0.0-alpha-1314299c7-20210901, which most definitely has a startTransition function defined, and I am seeing that warning printed.

I think this warning is intended since the shim of `useSyncExternalStore` does not support React 18. The warning should disappear in React 17 or once https://github.com/facebook/react/pull/22239 lands.

acdlite:
> It's not correctly memoizing and reusing the last returned value if it gets a new selector reference

This is intentional part of the design of useSyncExternalStore — we always call `getSnapshot` during render, because it might have been changed by a parent. That's how we avoid zombie props.

The "extra" version of the API supports skipping the additional call during render... but only if you pass the same selector. So it has to be stable/memoized. (The difference compared to `useMutableSource` is that it won't de-opt to synchronous mode if it renders during a transition, and it won't resubscribe.)

Alternatively, if you'd rather use the zombie props (to preserve Redux's existing behavior, I'm guessing?), you might as well skip the "extra" version of the API and track the latest selector on a ref like you currently do, since our selector implementation won't provide any additional benefit over the lower-level API.

acdlite:
Aside from preserving existing behavior, though, I'd say that for the majority of cases it's fine not to memoize the selector. Especially if they don't pass a custom `isEqual` argument. Most selectors aren't that computationally expensive.

For those cases where the selectors are expensive, there's a straightforward solution: wrap it in `useCallback`. You can tell users to think of it as an optimization, similar to other scenarios in React where you'd use `useMemo`, `memo`, or `PureComponent`. But you wouldn't want to do this prematurely everywhere, because usually the impact of running the selector again is negligible compared to the weight of another hook.

This is going to be our advice for `useSyncExternalStore` in general.

markerikson:
@acdlite I'm not sure I explained the "memoized value vs new selector ref" issue enough here.

When a new selector reference gets passed into `useSelector`, yes, we do need to call the new selector because it's entirely possible that it could return something completely different than the old one did.  (Granted, 99.9% of the time it's the exact some function code-wise, just inlined.)  However, it's also possible (and even likely) that the value returned from the new selector passes the equality check vs the last returned value.  In that case, we want to still reuse the old memoized value and return it, thus avoiding an extra unnecessary re-render.

What I'm seeing with the `uSESExtra` implementation atm is that because of how the memoization is implemented via a closure, it's impossible to get a new selector ref and still refer to the old result value to return it.

acdlite:
> However, it's also possible (and even likely) that the value returned from the new selector passes the equality check vs the last returned value. In that case, we want to still reuse the old memoized value and return it, thus avoiding an extra unnecessary re-render.

Ah I see, ok I think it makes sense to support that

acdlite:
What would happen if you stripped down the implementation as much as possible? What features is React missing in the current proposal? (Other than not bailing out on equal selections when the selector changes, as discussed above, which we'll add.)

In other words, in what ways would the following implementation be insufficient:

```js
function useSelector(selector, isEqual) {
  const store = useReduxStore();
  return useSyncExternalStoreExtra(
    store.subscribe,
    store.getState,
    selector,
    isEqual,
  );
}
```

Our goal with `useSyncExternalStore` is to upstream a lot of the complexity you're currently having to deal with in userspace, so whatever we're missing we'd love to hear about it.

I think the main extra things you currently have are:

- a subscription manager that's aware of the tree structure
- an error replay mechanism if a selector throws

My understanding is that both of these exist to account the zombie children scenario. However, the `useSyncExternalStore` is designed to avoid that, so I wonder which of your tests would break if you removed them and tried relying only on `useSES`?

acdlite:
> an error replay mechanism if a selector throws

I think we mentioned this previously, but React actually uses this exact same mechanism for all errors, which we leverage in `useSyncExternalStore`. So I'm hoping you can rely on that instead of implementing it again yourself.

acdlite:
> However, it's also possible (and even likely) that the value returned from the new selector passes the equality check vs the last returned value. In that case, we want to still reuse the old memoized value and return it, thus avoiding an extra unnecessary re-render.

So I remembered why I didn't implement it this way: React doesn't need the reference to be the same to skip re-rendering.

The way bailouts in React work is that, if a component re-renders but none of the state or props have changed, we assume the children are equal even if they fail an equality check.

So even if `useSyncExternalStoreExtra` returns a new reference, if no other state in the component has changed, then React will still bail out.
 
If you have a test that asserts the references are equal, then it will obviously fail, but if your test instead asserts that the children don't re-render, then it should pass.

If you find this still doesn't work as expected, could you point me to a failing test so I can take a look?

markerikson:
Sounds good.  And yeah, the currently failing test is at https://github.com/reduxjs/react-redux/pull/1808/files#diff-be963b0f28d22eab25f8f42209afbe9a4f5d16696f016853c310c7832303cdb5R621-R623 :

```js
        it('reuse latest selected state on selector re-run', () => {
          const alwaysEqual = () => true
          const Comp = () => {
            // triggers render on store change
            useNormalSelector((s) => s.count)
            const array = useSelector(() => [1, 2, 3], alwaysEqual)
            renderedItems.push(array)
            useLayoutEffect(() => {
              renderedItems.push(array)
            })
            return <div />
          }

          rtl.render(
            <ProviderMock store={normalStore}>
              <Comp />
            </ProviderMock>
          )

          normalStore.dispatch({ type: '' })
          act(() => {
            normalStore.dispatch({ type: '' })
          })

          expect(renderedItems.length).toBe(2)
          // TODO This is failing, and correctly so. `uSES` drops the memoized value if it gets a new selector.
          expect(renderedItems[0]).toBe(renderedItems[1])
        })
```

markerikson:
@acdlite I'll have to circle back to your other questions about "missing scenarios" later when I have more time to think about it.

You can see in this PR that I did some very stupid hacky things to get `connect` to continue working as-is - mostly the need to trigger those cascading subscriptions inside of the subscription listeners.  Not sure there's anything applicable for `uSES` out of that, though.

For the error handling, `useSelector` currently tries to keep the original error from the subscriber around for reference, so that if it happens again while rendering, we can say "error Y may be related to previous error X".  Tbh I'm not sure how fully useful that is, and falling back to `uSES`'s own error handling may be acceptable here.

acdlite:
Since this is not idiomatic Redux and you don't already support it, you could add that as a separate API?

acdlite:
Ok I fixed the memoization issue: https://github.com/facebook/react/pull/22307. 

Available in the experimental release channel, starting with `use-sync-external-store@0.0.0-experimental-fd5e01c2e-20210913`


Lemme know if it works as expected

markerikson:
The idea of an `areStatesEqual` comparison does have application beyond "we're being bad and mutating the state".  For example, if you know that you're only ever reading from `state.x`, you could supply `(prevState, newState) => prevState.x !== newState.x` and avoid _any_ execution of the selectors for this component until you know that `state.x` has changed.  So, sort of a pre-optimization that could skip the "run the selector" stage entirely.

`connect` has had that option since v5. I'm not sold on adding it, but since the question did just come up in that other thread, it was just on my mind as a thing that I don't know if we could do with `uSES` as currently implemented. Maybe we could, if we did some `getSnapshot` shenanigans, but unlikely it would be worth the effort (especially given that we've gotten along without that option thus far).

markerikson:
@acdlite Just updated all our React devDep packages locally to `0.0.0-experimental-fd5e01c2e-20210913`.  Looks like that does pass the "return the last memoized value" test now.

I only see one other current test failure, which is the "no warnings from `useLayoutEffect` in SSR" test :)

but I do have a couple other render-count related tests I need to go back and review a bit.

markerikson:
Huh.  @acdlite , I clearly have issues understanding how the whole `dispatcher` thing works :)

`unstable_useSyncExternalStore` does not appear to be exported by `react-dom@0.0.0-experimental-fd5e01ce-20210913`. However, it seems like it _is_ made available on the dispatcher, and if I hand-edit `use-sync-external-store.development.js` to refer to `builtInAPI = React.unstable_useSyncExternalStore`, and I add a logging line to the `uSES` function in `react.development.js`, I can confirm that the built-in API _is_ being called.

In fact, when I do that, I actually see _better_ test results in a way.

When I was doing the first work on this PR earlier, I saw three of our tests that do render counts in assertions start failing because of an extra render pass happening, and I couldn't figure out what was causing it.

If I do these hand-edits to force the use of the built-in `uSES`, I'm now seeing those tests revert back to the original (and expected) number of renders.  Here's the "failures" with the built-in version that show the reversion to the original behavior:

```
  ● React › hooks › useSelector › lifecycle interactions › notices store updates between render and store subscription effect

    expect(received).toEqual(expected) // deep equality

    - Expected  - 1
    + Received  + 0

      Array [
        0,
        1,
    -   1,
      ]

      238 |           // 2) Render due to dispatch, still sync in the initial render's commit phase
      239 |           // TODO 3) ??
    > 240 |           expect(renderedItems).toEqual([0, 1, 1])
          |                                 ^
      241 |         })
      242 |       })
      243 |

      at Object.<anonymous> (test/hooks/useSelector.spec.tsx:240:33)

  ● React › hooks › useSelector › performance optimizations and bail-outs › calls selector twice once on mount when state changes during render

    expect(received).toEqual(expected) // deep equality

    Expected: 3
    Received: 2

      427 |           // TODO As with "notice store updates" above, we're now getting _3_ renders here
      428 |           // expect(renderedItems.length).toEqual(2)
    > 429 |           expect(renderedItems.length).toEqual(3)
          |                                        ^
      430 |         })

```

So, it looks like there's a slight difference in behavior between the shim and the built-in implementation, which is not terribly surprising.

This is interesting. We _should_ want to have our tests run against the built-in version, but I can't just hand-edit the shim file to make it work the way I want to as a real solution.

Also not sure how to handle the difference in render counts here.  Will have to think about this a bit.

acdlite:
I think the issue is just that I `useSyncExternalStore` is still prefixed, even in the experimental channel, so the shim doesn't pick up the native version yet.

The reason that test fails is because the shim does not work with `createRoot`. It only works in legacy mode, where updates are always sync.

The idea is that the shim will always pick the native version in React 18, and always the shim in anything below that, because `useSyncExternalStore` doesn't exist in anything before 18.

The issue right now is that it _also_ doesn't exist in the 18 alphas yet. That's why you get the warning, as @eps1lon explained here: https://github.com/reduxjs/react-redux/pull/1808

(I didn't realize React testing library automatically uses `createRoot` when it's available. We should ask them (and by "them" I guess I mean @eps1lon) not do that, since there are enough subtle batching changes that it's likely to break some existing tests if they all get automatically switched over when you upgrade React. But this is a perfect example of why we prefix things — people add feature testing for unstable APIs and then when we ship the stable it has the potential break their users. See also: https://github.com/remix-run/react-router/issues/7308.

OTOH if React Testing Library users are already using `act` everywhere then maybe it'll just work? Doesn't seem like that's ​the case, though.)

Until the `useSyncExternalStore` API is finalized, what I'll do is update the shim to import the unstable version. (I should have done that from the beginning, didn't consider this in-between early testing phase.)

In the meantime, you can continue your testing by downgrading React and React DOM back to 17.

acdlite:
> (I should have done that from the beginning, didn't consider this in-between early testing phase.)

Well, I kind of did — that's what the warning was for :D

acdlite:
Ok the prefix issue is fixed now, starting with `0.0.0-experimental-806aaa2e2-20210913`

acdlite:
> I can only assume there's some shenanigans going on with import timing and Jest modules somehow? Perhaps it also has to do with React Hooks Testing Library not being updated yet.

It looks like you updated the `react` and `react-dom` packages, but not `react-test-renderer`. The API is imported from the `react` package (which is why `use-sync-external-store` attempts to use it) but the renderer doesn't implement it. If you change the versions to match (either both 17 or both 18) then it will work.

eps1lon:
> (* I didn't realize React testing library automatically uses createRoot when it's available. We should ask them (and by "them" I guess I mean @eps1lon) not do that, since there are enough subtle batching changes that it's likely to break some existing tests if they all get automatically switched over when you upgrade React. 

We specifically do this because you made it clear in your messaging that `createRoot` will be the default in React 18. It's also a bad impression to trigger warnings by default (`ReactDOM.render is deprecated`) in React 18.

I want to stress that warnings are very frustrating for library users in test suites since many suites are configured to fail if any warnings are triggered. So "it's just a warning" isn't helpful advise.

So it's a bit odd now to hear that we should not default to `createRoot` when React itself considers it the default going forward. I'll see if more repositories report integration issues. Though from my impression doing this work in Material-UI we had relatively few places where `act()` was missing. But that test suite is also maintained by myself so I may have a bit of a bias towards how well `act` is integrated into existing React test suites.

But the messaging is still a problem in my opinion. I don't have a strong opinion what should be the default going forward. But the ecosystem should agree on one, not many.

markerikson:
Will respond in more detail on the morning, but quick thoughts :

- yeah, I need to bump the renderer package
- we've been assuming that v8 would require React 18 as a min peer dep, just like v7 required 16.8 for hooks. So, I'd want to stick with 18 for testing. That said, if 16/17 compat is viable, then A) we'd need to figure out a good way to run tests against each version, and B) the tests would have to account for some differences in behavior like batching. Like, I just saw earlier that the shim produces an extra render in some cases than the built-in api does.

acdlite:
> createRoot will be the default in React 18

You can default to that for new apps that use React 18, the question is whether you should automatically switch over existing apps to `createRoot` as soon as they upgrade.

Any updates that aren't wrapped in `act` are immediately going to break. Though maybe I'm mistaken in my assertion that users of React Testing Library don't already use `act` everywhere?

My expectation is that most existing apps aren't that rigorous about using `act` everywhere, because the `act` warning doesn't fire in all cases. For example, there's no act warning for class-based updates today. Or root-level updates.

acdlite:
> we've been assuming that v8 would require React 18 as a min peer dep

Ok that's fine, I wasn't sure if you were going to require a bump or not. Either way the prefix thing is fixed now so if you update the version I think it'll work.

> Like, I just saw earlier that the shim produces an extra render in some cases than the built-in api does.

I would try it again, I tried it locally and the only time I got an extra render was when running the shim in `createRoot`. Which shouldn't happen now that I fixed the prefix thing.

acdlite:
@eps1lon Ah, does React Testing Library automatically wrap all updates in `act` as part of the framework? In that case, then defaulting to `createRoot` whenever it's available makes sense to me, I think.

eps1lon:
> > createRoot will be the default in React 18
> 
> You can default to that for new apps that use React 18, the question is whether you should automatically switch over existing tests to `createRoot` as soon as they upgrade.
> 
> Any updates that aren't wrapped in `act` are immediately going to break those tests. Though maybe I'm mistaken in my assertion that users of React Testing Library don't already use `act` everywhere?

Our APIs are all wrapped in `act`, yes. Examples that come to mind that probably break are `HTMLElement.focus()`.

Though I suspect that most test should pass. On the other hand, if we default to `ReactDOM.render` and people let their tests fail if they see warnings, then every test will fail. So the question is which scenario is more common which we're unlikely to answer. 

> My expectation is that most existing apps aren't that rigorous about using `act` everywhere, because the `act` warning doesn't fire in all cases.

That was my impression as well.

acdlite:
We should continue the testing discussion in a separate thread, but my thinking re: the legacy API warning was that you should either upgrade all your tests up front, or if that's not feasible, update your test infra to temporarily silence those warnings until you're able to gradually fix them.

Could maybe publish a codemod, too, though there are so many testing frameworks and patterns idk if that'll work

markerikson:
Hmm. I just pulled down build `0.0.0-experimental-50263d327-20210914` for `react`, `react-dom`, and `react-test-renderer`.  Good news is the `useSelector` tests all pass now (with the tweak of removing the "expect an extra render" assertions I'd had to change with the shim).

Bad news - now a couple other test files are suddenly failing, because they use `react-dom/server` and that's throwing a "Not yet implemented" error:

```
  ● React › server rendering › should be able to render connected component with props and state from store

    Not yet implemented

      712 |
      713 |       try {
    > 714 |         actualChildProps = useSyncExternalStore(
          |                            ^
      715 |           subscribeForReact,
      716 |           actualChildPropsSelector
      717 |         )

      at Object.useSyncExternalStore (node_modules/react-dom/cjs/react-dom-server-legacy.node.development.js:4874:9)
      at Object.useSyncExternalStore (node_modules/react/cjs/react.development.js:1629:21)
      at Connect(Greeting) (src/components/connect.tsx:714:28)
      at renderWithHooks (node_modules/react-dom/cjs/react-dom-server-legacy.node.development.js:5279:16)
```

acdlite:
Server error is expected for now, just haven't implemented that part yet

brainkim:
Have y’all tried this against @dai-shi’s https://github.com/dai-shi/will-this-react-global-state-work-in-concurrent-rendering yet? Has anyone confirmed that usage of the `useSyncExternalStore()` API actually mitigates tearing?

markerikson:
Nope, but amusingly someone else brought this up in a React 18 discussion just a couple hours ago :)

I'll add it to my list of things to try, but it would be great if someone else wanted to try that as well.

markerikson:
I just tried running @dai-shi 's tests against this v8 branch. As setup, I copied the existing `react-redux` test to `react-redux-v8`, also installed the CSB build from this PR as an alias of `react-redux-v8`, and updated the React and ReactDOM deps to the same versions I have in this PR.

The results were... interesting:

```
 FAIL  __tests__/all_spec.js (35.122 s)
  react-redux
    with useTransition
      √ test 1: updated properly with transition (2651 ms)
      × test 2: no tearing with transition (65 ms)
      √ test 3: ability to interrupt render (1 ms)
      × test 4: proper branching with transition (7230 ms)
    with intensive auto increment
      √ test 5: updated properly with auto increment (2220 ms)
      × test 6: no tearing with auto increment (1 ms)
  react-redux-v8
    with useTransition
      × test 1: updated properly with transition (5859 ms)
      √ test 2: no tearing with transition (68 ms)
      × test 3: ability to interrupt render (1 ms)
      × test 4: proper branching with transition (4506 ms)
    with intensive auto increment
      √ test 5: updated properly with auto increment (2221 ms)
      × test 6: no tearing with auto increment (2 ms)
```

I feel like this sorta makes sense, though?  We've established the whole point of `uSES` is to _not_ deal with transitions

But, I _am_ surprised that there still appears to be tearing of some kind.

dai-shi:
Failing test 1 is unexpected. There can be a problem somewhere, and it can be the tool.

The goal for r-r-v8 is to pass test 1, 2, and 3. 4 is not possible. (5 and 6 don't work well now).

> I feel like this sorta makes sense, though? We've established the whole point of `uSES` is to _not_ deal with transitions

I didn't think so, but now unsure with the recent discussion in the WG.

brainkim:
@dai-shi @markerikson Do y’all have a link to the react working group discussions? Happy to read new thoughts on this stuff.

acdlite:
will-this-react-global-state-work-in-concurrent-rendering is a community-maintained spec, so some of the assertions it makes are based on assumptions that have recently become outdated. A big one, as @markerikson pointed out, is whether external store mutations are expected to work with `startTransition`. That's not the case anymore — even with `useMutableSource`, we never guaranteed that would work in all cases, because full support for concurrency requires maintaining multiple concurrent copies of your state. The only way to do this is with the built-in APIs: `useState` and `useReducer`. You can read more background on this decision here: https://github.com/reactwg/react-18/discussions/86

The new strategy is for store updates to always be synchronous.

That doesn't mean concurrent rendering is completely disabled — you can still pass a prop that is derived from external state to `useDeferredValue`. That's how you might implement a typeahead, for example: a synchronous state value that controls the text input, and a deferred value that populates the typeahead suggestions.

```js
const syncValue = useSyncExternalStore(...);
const deferredValue = useDeferredValue(syncValue);

return ( 
  <>
    <Input value={syncValue} />
    <ExpensiveVisualization value={deferredValue} />
  </>
);
```

@dai-shi, perhaps we can work together to update the will-this-react-global-state-work-in-concurrent-rendering project to match the latest expectations?

***

A word of caution, though: test suites like will-this-react-global-state-work-in-concurrent-rendering can be really useful for detecting the presence of _certain_ tearing bugs — but it can't prove the _absence_ of _any_ tearing bugs. It's really just not possible to exhaustively prevent concurrency issues with runtime assertions; you'd need something like Rust's ownership system and compile-time borrow checker.

Simple example: one thing that's missing from the will-this-react-global-state-work-in-concurrent-rendering test suite are tests that use progressive hydration, which is another form of concurrent rendering that is different than just time slicing. Suspense is another example, of course. We have other features like pre-rendering offscreen content that introduce even more subtle scenarios. It's just not feasible for us to expect every library maintainer (whether open or closed source) to become a concurrency expert and account for every one of React's myriad upcoming features.

So instead the message is: just `useSyncExternalStore`. (Or useState, useReducer, or one of the other built-in APIs.) Let React worry about ensuring the correct behavior. And if you find any bugs, file an issue and we will fix them in React itself.

src/hooks/useSelector.ts
```diff
@@ -16,94 +19,29 @@ function useSelectorWithStoreAndSubscription<TStoreState, TSelectedState>(
...
-    () => createSubscription(store, contextSub),
-    [store, contextSub]
-  )
-
-  const latestSubscriptionCallbackError = useRef<Error>()
-  const latestSelector = useRef<TSelector<TStoreState, TSelectedState>>()
-  const latestStoreState = useRef<TStoreState>()
-  const latestSelectedState = useRef<TSelectedState>()
+  const subscribe = useMemo(() => {
+    const subscription = createSubscription(store, contextSub)
```
acdlite, line 23: Since the plan is to require React 18 in the next version, can we remove this subscription indirection and subscribe directly to the store? Now that 18 has batching by default.

markerikson: The one other hypothetical piece that would potentially lose is that in theory you can avoid the "zombie child" problem by slapping a `connect()`-ed component above a `useSelector()`-ed component, in which case the selector subscribes to the connect's `Subscription`.

In practice I don't think anyone seriously does that.

Given that `Subscription` is like half of the total bytes here for `useSelector` anyway (like 1K), that also seems reasonable to do.

acdlite: I think we addressed the zombie child problem is fixed by the new strategy where we always re-run the selector in the render phase (if it changed), right?

That was the intent when we implemented useSyncExternalStore, at least. If you can think of an example where it breaks let me know so we can fix it upstream.

dai-shi:
(sorry for the slight off topic.)

> it can't prove the absence of all tearing bugs.

Totally agreeable. It's just one limited test. Passing it doesn't guarantee anything. I got interested after I saw Flarnie Marchan's talk that tearing is not hypothetical but it happened in Facebook, and it's something users can observe.

> The new strategy is for store updates to always be synchronous.

Given that CM is gone, and uSES is coming, I wonder what's left to test. But,

> perhaps we can work together to update the will-this-react-global-state-work-in-concurrent-rendering project to match the latest expectations?

if there's room for trial, I'm happy to help.

> That doesn't mean concurrent rendering is completely disabled — you can still pass a prop that is derived from external state to useDeferredValue.

This sounds interesting, and I'm still trying to understand it. (https://github.com/reactwg/react-18/discussions/86#discussioncomment-1322765)

src/hooks/useSelector.ts
markerikson: Also a good point, and tbh I'm not sure if we actually _have_ tests for zombie child-related behavior in our codebase atm.  Worth trying to add something.

acdlite:
> Totally agreeable. It's just one limited test. Passing it doesn't guarantee anything. I got interested after I saw Flarnie Marchan's talk that tearing is not hypothetical but it happened in Facebook, and it's something users can observe.

Yeah absolutely, it's a good idea and an e2e test like you've done is the right way to go about it. Only clarifying so peopel are aware of the limits of the approach.

> Given that CM is gone, and uSES is coming, I wonder what's left to test.

It could still be useful to demonstrate the difference between built-in state APIs like `useState` versus libraries that use external systems. Or to demonstrate common patterns with popular frameworks, like the `useDeferredValue` pattern I posted above.

Usually when I'm evaluating a component or library to check for concurrent compatibility, I take an adversarial approach — I open the codebase and look for things that I know won't work. Almost always, the issue is something related to reading from an external data source, which is why we're so invested in providing a first-class solution to that. But we could do something similar for libraries that request it — they ask for an audit by submitting a pull request, and the reviewer identifies areas that might need to change, and creates a test that demonstrates the problem.

brainkim:
Do y’all have a way to squash `useSyncExternalStore()` shim `useLayoutEffect()` warnings for SSR yet? Is the intention not to use `useSyncExternalStore()` in SSR?

acdlite:
@brainkim We'll circumvent it somehow, I'm guessing with the same `canUseDOM` hack everyone else uses.

acdlite:
The plan for server rendering is to add a `getServerSnapshot` third argument that is called instead of the normal `getSnapshot` both on the server and during hydration: https://github.com/facebook/react/pull/22347

That ensures it's compatible with incremental hydration in React 18.

The userspace shim won't be able to call `getServerSnapshot` during hydration because older versions of React don't expose that information, but hydration is also synchronous so it's not net worse anyway. It's just as good/bad as it already is.

acdlite:
The server implementation landed. Available starting with `0.0.0-experimental-79b8fc667-20210920`

markerikson:
I've spent the last couple days resuscitating the React-Redux benchmarks repo at  .  I was just able to do a v8 alpha build from this branch, load it into the benchmarks suite, and compare vs 5.x, 6.x, and 7.x:

```
+-----------------------------------------
¦ Version       ¦ Avg FPS ¦ Render       ¦
¦               ¦         ¦ (Mount, Avg) ¦
+---------------+---------+--------------+
¦ 5.1.2         ¦ 42.01   ¦ 212.1, 1.0   ¦
+---------------+---------+--------------+
¦ 6.0.1         ¦ 9.77    ¦ 193.1, 9.2   ¦
+---------------+---------+--------------+
¦ 7.2.5         ¦ 43.94   ¦ 221.3, 0.6   ¦
+---------------+---------+--------------+
¦ 8.0.0-alpha.0 ¦ 43.76   ¦ 223.5, 0.7   ¦
+-----------------------------------------

+-----------------------------------------
¦ Version       ¦ Avg FPS ¦ Render       ¦
¦               ¦         ¦ (Mount, Avg) ¦
+---------------+---------+--------------+
¦ 5.1.2         ¦ 12.66   ¦ 140.6, 0.1   ¦
+---------------+---------+--------------+
¦ 6.0.1         ¦ 7.92    ¦ 133.4, 7.2   ¦
+---------------+---------+--------------+
¦ 7.2.5         ¦ 21.44   ¦ 157.9, 1.1   ¦
+---------------+---------+--------------+
¦ 8.0.0-alpha.0 ¦ 20.88   ¦ 172.6, 1.3   ¦
+-----------------------------------------


Results for benchmark forms:
+-----------------------------------------
¦ Version       ¦ Avg FPS ¦ Render       ¦
¦               ¦         ¦ (Mount, Avg) ¦
+---------------+---------+--------------+
¦ 5.1.2         ¦ 49.67   ¦ 1371.0, 0.4  ¦
+---------------+---------+--------------+
¦ 6.0.1         ¦ 49.42   ¦ 1452.5, 3.0  ¦
+---------------+---------+--------------+
¦ 7.2.5         ¦ 49.69   ¦ 1321.1, 0.4  ¦
+---------------+---------+--------------+
¦ 8.0.0-alpha.0 ¦ 47.06   ¦ 1331.3, 0.5  ¦
+-----------------------------------------


Results for benchmark stock-ticker:
+-----------------------------------------
¦ Version       ¦ Avg FPS ¦ Render       ¦
¦               ¦         ¦ (Mount, Avg) ¦
+---------------+---------+--------------+
¦ 5.1.2         ¦ 33.08   ¦ 424.5, 0.9   ¦
+---------------+---------+--------------+
¦ 6.0.1         ¦ 12.86   ¦ 331.3, 11.8  ¦
+---------------+---------+--------------+
¦ 7.2.5         ¦ 32.82   ¦ 469.7, 0.6   ¦
+---------------+---------+--------------+
¦ 8.0.0-alpha.0 ¦ 29.69   ¦ 463.0, 0.6   ¦
+-----------------------------------------


+-----------------------------------------
¦ Version       ¦ Avg FPS ¦ Render       ¦
¦               ¦         ¦ (Mount, Avg) ¦
+---------------+---------+--------------+
¦ 5.1.2         ¦ 40.12   ¦ 866.8, 0.3   ¦
+---------------+---------+--------------+
¦ 6.0.1         ¦ 28.95   ¦ 741.3, 27.2  ¦
+---------------+---------+--------------+
¦ 7.2.5         ¦ 43.40   ¦ 896.8, 0.3   ¦
+---------------+---------+--------------+
¦ 8.0.0-alpha.0 ¦ 44.37   ¦ 1020.8, 0.3  ¦
+-----------------------------------------


+-----------------------------------------
¦ Version       ¦ Avg FPS ¦ Render       ¦
¦               ¦         ¦ (Mount, Avg) ¦
+---------------+---------+--------------+
¦ 5.1.2         ¦ 23.57   ¦ 110.4, 2.0   ¦
+---------------+---------+--------------+
¦ 6.0.1         ¦ 13.81   ¦ 109.4, 10.8  ¦
+---------------+---------+--------------+
¦ 7.2.5         ¦ 23.90   ¦ 125.4, 1.1   ¦
+---------------+---------+--------------+
¦ 8.0.0-alpha.0 ¦ 23.32   ¦ 136.0, 1.1   ¦
+-----------------------------------------
```

Those numbers come with a bunch of caveats:

- FPS is a lousy metric, but it's _a_ number
- These are horribly synthetic benchmarks (as in, "Mark cranked up the `NUM_COMPONENTS` constant until the numbers dropped below 60FPS")
- This is only for `connect`. I haven't set up any scenarios with `useSelector` yet, although I should be able to do that now that I've revamped the build setup
- What really matters here is the relative FPS values, and the "mount/render" numbers
- This is just one run, and the numbers do fluctuate a bit from run to run

Overall, the key takeaway is that `connect` on this branch is basically equivalent in perf to `connect` in v7, which is exactly what I would hope and expect given the actual changes I made so far, and that there's no major regressions like we had in v6.

I'll look at trying to set up some `useSelector` benchmarks soon.

markerikson:
I spent the evening trying to get a hooks-based benchmarks scenario in place. I did. And then the first time I ran it with 8.0-alpha, the app never updated.

Know why?

BECAUSE ANDREW CHANGED THE ORDER OF THE ARGUMENTS TO `useSyncExternalStore` AND THE EQUALITY FUNCTION WAS GETTING TREATED AS `getSnapshot`! :)

The latest alpha had gotten installed in the benchmarks repo, but I hadn't tried to update React-Redux to match that yet. Oops.

So, for now I just stubbed out `getServerSnapshot` to reuse `store.getState`. **YES, I KNOW THIS IS "WRONG"**.  I just wanted to keep the client-side code running for now so I could benchmark things.

First meaningful hooks scenario results:

```
Results for benchmark deeptree-nested-hooks:
+-----------------------------------------
¦ Version       ¦ Avg FPS ¦ Render       ¦
¦               ¦         ¦ (Mount, Avg) ¦
+---------------+---------+--------------+
¦ 7.2.5         ¦ 44.33   ¦ 167.5, 0.5   ¦
+---------------+---------+--------------+
¦ 8.0.0-alpha.0 ¦ 45.75   ¦ 166.4, 0.6   ¦
+-----------------------------------------
```

Again, the numbers are very imperfect and this is just one benchmark, but traditionally this is the one that's shown the biggest variation between different builds.  So, very encouraging that 7.x and 8.x are basically identical here.

Also... **ALL THE TESTS NOW PASS!**  Like I said, I know the SSR behavior is wrong and I'll have to sort out what to do with that.  

Andrew has suggested having some kind of a `serverSnapshot` prop in `Provider` itself, which I guess would require passing that down via context and having a trivial `() => serverSnapshot` getter in use everywhere in the tree if that's available.  On my todo list, but I'll need some time to let that sink in.

markerikson:
I just ripped the `Subscription` usage out of `useSelector`.  All tests still pass, except the one we had that specifically checked for "you can stick a `connect()`ed component above and it will subscribe to that instead" (ie, the use case I mentioned above).

I also realized the TS types output was getting nested one level too deep (as `./es/src/index.d.ts` instead of `./es/index.d.ts`), which made the published package busted for TS usage.  Managed to fiddle with the `tsconfig` enough to make that seem to cooperate and the types work right now.

types/use-sync-external-store/index.d.ts
```diff
@@ -0,0 +1,21 @@
+/* eslint-disable no-unused-vars */
+
+type VoidFunc = () => void
+
+declare module 'use-sync-external-store' {
```
eps1lon, line 5: [`@types/use-sync-external-store`](https://www.npmjs.com/package/@types/use-sync-external-store) is now available.

markerikson:
I rebased this against `master`, and squashed a bunch of temp commits together to clean up the history.  I also fixed up some types issues.  Tests pass, types pass, the CSB build looks like it runs and compiles okay in an example app.  I think it's time to merge this PR and publish an initial v8 alpha so that people could start trying it out.

I still haven't tried to implement an actual API for server snapshots. I'll tackle the SSR thing in a follow-up.

I've updated the React peerDep to list `16 || 17 || 18` (although I don't know how to list "experimental" builds here). Per prior comments, I'm generally planning to make this 18-only by the time it's out.

Related to that, I'm seriously considering removing the `use-sync-external-store` shim dep, and inlining the logic from `useSyncExternalStoreExtra` back into `useSelector`.  Otherwise, the logic from the `useSyncExternalStore` shim function ends up in the final app bundle even when you're using the built-in `uSES` function, and that's just a waste of bytes.  If we're not going to support 16/17 via the shim in the final v8 release, then I don't see a point in having that dep at all.

Also, I had removed use of `Subscription` from `useSelector`, but that still ends up in the app bundle because `Provider` has to use it in case there's some `connect` usage in the app.  So, can't get away from that.

acdlite:
@markerikson We could publish a version of the "extra" module that relies directly on the native API, instead of using the shim indirection. That way if multiple libraries depend on it, users only download the code once.

markerikson:
@acdlite yup, that'd definitely be another possibility, and I'd be good with that.

markerikson:
Also FYI to anyone reading, these changes are now live in https://github.com/reduxjs/react-redux/releases/tag/v8.0.0-alpha.0 !

brainkim:
>  If we're not going to support 16/17 via the shim in the final v8 release, then I don't see a point in having that dep at all.

My understanding is that the shim will support post-hooks versions once the API is stable.

markerikson:
My point is that the "shim" part is only useful if we planned to have v8 be compatible with 16/17.  But, I was already planning to enforce a hard switch so that React-Redux v8 _requires_ React 18, in the same way that React-Redux v7 required 16.8+ for hooks support.

As currently written, the `use-sync-external-store` package will always end up including the shim code itself in the final app bundle, even if you're using a version of React that does have a real `uSES`.  If we really do hard-require React 18, then that's a useless 750 bytes for end users that wouldn't even run.

acdlite:
As of https://github.com/facebook/react/pull/22662, if you import `use-sync-external-store/with-selector` it will compose the built-in React API, not the shim. (The shim version is available as a separate import: `use-sync-external-store/shim/with-selector`.)

Kasdy:
> Also, I had removed use of `Subscription` from `useSelector`

Hello, shouldn't this bring back the old quadratic algorithm and slow unmount of many components (#1523)?
Since Redux unsubscribe is passed to `useSyncExternalStore` and it relies on indexOf\splice (like in version <7.2.0)
Thank you

markerikson:
@kasdy : not sure, but if there's a problem here it's probably going to be a React issue rather than a React-Redux issue specifically.

Could you try to set up a test case with both React-Redux v8 + React 18, but also "just" React 18 + a fake little store implementation + `useSyncExternalStore`, and see what the actual behavior looks like?

Kasdy:
@markerikson 
>Hmm. However... because we're now using store.subscribe directly, that does use indexOf/splice.

Yes! that's what I meant :)

I adapted the benchmark code in #1523 to use Hooks and tested:

React 17 + React-Redux v7.2.6
Screenshot 2022-02-05 01 24 30

React 18 + React-Redux v8.0.0
Screenshot 2022-02-05 01 36 43


```javascript
const store = createStore(
);

function App() {
    const state = useSelector(s => s);
    const dispatch = useDispatch();

    const [children, setChildren] = useState(0);

    const toggleChildren = () => setChildren(c => c ? 0 : 50000);

    return <div>
        <button onClick={toggleChildren}>Toggle Children</button>
        <button onClick={() => dispatch(increment())}>Increment</button>
        {[...Array(children).keys()].map(i => <Child key={i}/>)}
    </div>;

}

function Child() {
    const v = useSelector(s => s);
    return <div className="child">{v}</div>;
}
```

markerikson:
Gotcha.

Could you file a new issue so we can keep track of this separately?

Also: I would really like to have a unit test that captures this intended behavior so that we don't regress again, but off the top of my head I don't know how you'd actually validate this experimentally (especially since perf-sensitive behavior is not a good fit for unit tests that run in C).

Kasdy:
@markerikson 
done #1869

About the unit test, that would be really cool, but I can't think of a way to detect it... because it works even though it's slow.
measuring the test execution time does not look good to me either.
I discovered it by chance, because I was studying the new React18 hooks, and I was digging in React-Redux source.

>Also also: if you have time to file a PR that changes useSelector back to using Subscription again, that would be great!

I have to check the guidelines and everything... but I think I can do it :) it will be my first PR, it will be fun :D

Issue #1732, opened 2021-06-10
Investigation: try React-Redux v7 / v8 with React 18 alpha


markerikson:
React 18 is now available as preview alpha releases (per https://reactjs.org/blog/2021/06/08/the-plan-for-react-18.html ).

The React team has put together a Working Group discussion forum to describe the changes in React 18 in more detail: https://github.com/reactwg/react-18/discussions

Looking at that forum, there's several issues that may be relevant to React-Redux in some way:

- Overviews:
  - [Introducing React 18](https://github.com/reactwg/react-18/discussions/4)
  - [Upgrading to React 18 on the Client](https://github.com/reactwg/react-18/discussions/6)
  - [Installing React 18 Alpha](https://github.com/reactwg/react-18/discussions/9)
- Behavior changes:
  - [Automatic batching](https://github.com/reactwg/react-18/discussions/21)
  - [Adding Strict Effects to StrictMode](https://github.com/reactwg/react-18/discussions/19)
  - [How to support Strict Effects](https://github.com/reactwg/react-18/discussions/18)
  - [Testing with Strict Effects](https://github.com/reactwg/react-18/discussions/17)
  - [How to support Strict Effects](https://github.com/reactwg/react-18/discussions/18)

I'd like to have some folks start trying out React-Redux v7 with React 18 alpha, and report any interesting findings.  We could potentially use some of the example apps from the Redux tutorials, and any of the example CodeSandboxes linked in our docs.

You would need to:

- Switch the React version to one of the React 18 alphas, and update the app to use `React.createRoot()` per instructions
- Try out the app
- Hand-inspect the behavior to see if anything seems unexpected

We're interested in seeing examples and comparisons with both `connect` and `useSelector`.

It would also be worth cloning the React-Redux repo, and updating our React devDep to run our tests against the alphas.

The React team has reported that a brief test of a couple Redux-based apps combined with `createRoot + StrictMode` did not appear to be working correctly - in particular, no re-rendering happening despite actions being dispatched.

It's very possible that Strict Effects running effects callbacks more than once is causing problems with subscriptions.

We'd appreciate any feedback reports!  

**Please provide this info**:

- Details on the example app used
- Link to updated source code / CodeSandbox if possible
- Description of observed behavior, good or bad ("works okay", "clicking button X doesn't update the UI", etc)

gaearon:
[...]

PR #1817, opened 2021-09-14
Fixed compatibility with React 18 strict effects

Andarist:
Related to: https://github.com/reduxjs/react-redux/issues/1732#issuecomment-918697306

I could write a test for this but that would require me to target the [`feature/react-18-initial-compat`](https://github.com/reduxjs/react-redux/tree/feature/react-18-initial-compat) and I'm not sure if that's a good idea at this point.

Let me know if you want me to target some other branch or maybe if this is totally irrelevant for the time being. Note that I have not actually tested this with React 18 yet but, from a theoretical point of view, it should help with the investigated issue. I was just curious to see if there is any particular reason why the "symmetry" was not here so I've decided to run a test suite against a "symmetric" implementation to check out if there would be any failures. Since the test suite has passed - I've thought that it might be worth pushing this out.

### DIFF
--- a/src/components/Provider.tsx
+++ b/src/components/Provider.tsx
@@ -22,7 +22,6 @@ export interface ProviderProps<A extends Action = AnyAction> {
 function Provider({ store, context, children }: ProviderProps) {
   const contextValue = useMemo(() => {
     const subscription = createSubscription(store)
-    subscription.onStateChange = subscription.notifyNestedSubs
     return {
       store,
       subscription,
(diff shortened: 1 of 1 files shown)


markerikson:
Yeah, I tried out literally the same change myself locally :)  We ought to do this to the `7.x` branch as well just for completeness sake.

markerikson:
Narrator:  it turns out they should, in fact, have done this to the `7.x` branch for completeness sake.

PR #1835, opened 2021-11-02
[DRAFT] Add partial SSR support (Provider and useSelector)

markerikson:
This PR:

- Updates the context types to include a `getServerState` field
- Updates `Provider` to accept a `serverState` prop, and if included, pass down a `getServerState` callback that returns that value
- Updates `useSelector` to pass `getServerState` to `uSES`
- Identifies areas where `connect` is almost definitely abusing `uSES` :)

I'd like to get some eyes on what I've got so far and find out if I'm headed in the right direction overall.  The changes here are based on comments made by @acdlite in various React 18 Working Group discussions.

I'm also interested in example tests we can use to verify that any of these changes are working as intended - please offer suggestions or PRs!

`connect`'s logic is screwy because we need to know if the child props are being recalculated _first_, so that we either trigger a re-render or go ahead and cascade the subscription notifications down the tree.  As a result, my current code still has `connect` doing _all_ the actual meaningful comparison work, and then abusing `uSES` by passing down the "child props selector" function as the `getSnapshot` argument to `uSES`.  The actual call to `getState` still happens within `connect` itself.

This has at least passed all our tests so far, but now that I'm trying to figure out how to potentially take the `getServerState` callback and pass it to `uSES`, I'm pretty sure all of this is basically incorrect usage :)

This suggests there are three possibilities:

- Some hypothetical fix that involves just shuffling pieces around and somehow it'll end up being sufficiently okay to work
- Completely rewriting the guts of `connect` + the `mapState/mapDispatch` logic (which has been basically as-is and untouched since it was written in late 2016) in order to better accomodate the revised requirements
- Leave `connect` as-is and say "look, it'll _run_ okay, but if SSR hydration is your concern we just can't fully support that - it's too difficult.  So, if you're still using `connect` you'll have to accept whatever fallback behavior React does here".

Paging interested parties:

@acdlite @timdorr @eps1lon @phryneas @dai-shi @Andarist

### DIFF
--- a/src/components/Provider.tsx
+++ b/src/components/Provider.tsx
@@ -4,17 +4,23 @@ import { createSubscription } from '../utils/Subscription'
 import { useIsomorphicLayoutEffect } from '../utils/useIsomorphicLayoutEffect'
 import { Action, AnyAction, Store } from 'redux'
 
-export interface ProviderProps<A extends Action = AnyAction> {
+export interface ProviderProps<A extends Action = AnyAction, S = any> {
   /**
    * The single Redux store in your application.
    */
-  store: Store<any, A>
+  store: Store<S, A>
+
+  /**
+   * An optional server state snapshot. Will be used during initial hydration render if available, to ensure that the UI output is consistent with the HTML generated on the server.
+   */
+  serverState?: S
+
   /**
    * Optional context to be used internally in react-redux. Use React.createContext() to create a context to be used.
    * If this is used, you'll need to customize `connect` by supplying the same context provided to the Provider.
    * Initial value doesn't matter, as it is overwritten with the internal state of Provider.
    */
-  context?: Context<ReactReduxContextValue<any, A>>
+  context?: Context<ReactReduxContextValue<S, A>>
   children: ReactNode
 }
 
@@ -22,14 +28,16 @@ function Provider<A extends Action = AnyAction>({
   store,
   context,
   children,
+  serverState,
 }: ProviderProps<A>) {
   const contextValue = useMemo(() => {
     const subscription = createSubscription(store)
     return {
       store,
       subscription,
+      getServerState: serverState ? () => serverState : undefined,
     }
-  }, [store])
+  }, [store, serverState])
 
   const previousState = useMemo(() => store.getState(), [store])
 
--- a/src/components/connect.tsx
+++ b/src/components/connect.tsx
@@ -602,7 +602,7 @@ function connect<
         ? props.store!
         : contextValue!.store
 
-      const getServerSnapshot = didStoreComeFromContext
+      const getServerState = didStoreComeFromContext
         ? contextValue.getServerState
         : store.getState
 
@@ -738,8 +738,9 @@ function connect<
           // TODO This is incredibly hacky. We've already processed the store update and calculated new child props,
           // TODO and we're just passing that through so it triggers a re-render for us rather than relying on `uSES`.
           actualChildPropsSelector,
-          // TODO Need a real getServerSnapshot here
-          actualChildPropsSelector
+          getServerState
+            ? () => childPropsSelector(getServerState(), wrapperProps)
+            : actualChildPropsSelector
         )
       } catch (err) {
         if (latestSubscriptionCallbackError.current) {
--- a/src/components/connect.tsx
+++ b/src/components/connect.tsx
@@ -27,7 +27,10 @@ import defaultMapStateToPropsFactories from '../connect/mapStateToProps'
 import defaultMergePropsFactories from '../connect/mergeProps'
 
 import { createSubscription, Subscription } from '../utils/Subscription'
-import { useIsomorphicLayoutEffect } from '../utils/useIsomorphicLayoutEffect'
+import {
+  useIsomorphicLayoutEffect,
+  canUseDOM,
+} from '../utils/useIsomorphicLayoutEffect'
 import shallowEqual from '../utils/shallowEqual'
 
 import {
@@ -124,6 +127,7 @@ function subscribeUpdates(
       return
     }
 
+    // TODO We're currently calling getState ourselves here, rather than letting `uSES` do it
     const latestStoreState = store.getState()
 
     let newChildProps, error
@@ -157,6 +161,7 @@ function subscribeUpdates(
       childPropsFromStoreUpdate.current = newChildProps
       renderIsScheduled.current = true
 
+      // TODO This is hacky and not how `uSES` is meant to be used
       // Trigger the React `useSyncExternalStore` subscriber
       additionalSubscribeListener()
     }
@@ -597,6 +602,10 @@ function connect<
         ? props.store!
         : contextValue!.store
 
+      const getServerSnapshot = didStoreComeFromContext
+        ? contextValue.getServerState
+        : store.getState
+
       const childPropsSelector = useMemo(() => {
         // The child props selector needs the store reference as an input.
         // Re-create this selector whenever the store changes.
@@ -724,7 +733,10 @@ function connect<
 
       try {
         actualChildProps = useSyncExternalStore(
+          // TODO We're passing through a big wrapper that does a bunch of extra side effects besides subscribing
           subscribeForReact,
+          // TODO This is incredibly hacky. We've already processed the store update and calculated new child props,
+          // TODO and we're just passing that through so it triggers a re-render for us rather than relying on `uSES`.
           actualChildPropsSelector,
           // TODO Need a real getServerSnapshot here
           actualChildPropsSelector
--- a/src/hooks/useSelector.ts
+++ b/src/hooks/useSelector.ts
@@ -48,13 +48,13 @@ export function createSelectorHook(
       }
     }
 
-    const { store } = useReduxContext()!
+    const { store, getServerState } = useReduxContext()!
 
     const selectedState = useSyncExternalStoreWithSelector(
       store.subscribe,
       store.getState,
       // TODO Need a server-side snapshot here
-      store.getState,
+      getServerState || store.getState,
       selector,
       equalityFn
     )
--- a/src/utils/useIsomorphicLayoutEffect.ts
+++ b/src/utils/useIsomorphicLayoutEffect.ts
@@ -9,9 +9,11 @@ import { useEffect, useLayoutEffect } from 'react'
 // is created synchronously, otherwise a store update may occur before the
 // subscription is created and an inconsistent state may be observed
 
-export const useIsomorphicLayoutEffect =
+// Matches logic in React's `shared/ExecutionEnvironment` file
+export const canUseDOM = !!(
   typeof window !== 'undefined' &&
   typeof window.document !== 'undefined' &&
   typeof window.document.createElement !== 'undefined'
-    ? useLayoutEffect
-    : useEffect
+)
+
+export const useIsomorphicLayoutEffect = canUseDOM ? useLayoutEffect : useEffect
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "coverage": "codecov"
   },
   "peerDependencies": {
-    "react": "^18.0.0-beta"
+    "react": "^18.0.0-rc"
   },
   "peerDependenciesMeta": {
     "react-dom": {
@@ -52,15 +52,11 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.12.1",
-    "@testing-library/react-12": "npm:@testing-library/react@^12",
     "@types/hoist-non-react-statics": "^3.3.1",
     "@types/use-sync-external-store": "^0.0.3",
     "hoist-non-react-statics": "^3.3.2",
-    "react-17": "npm:react@^17",
-    "react-dom-17": "npm:react-dom@^17",
-    "react-is": "^18.0.0-beta-fdc1d617a-20211118",
-    "react-test-renderer-17": "npm:react-test-renderer@^17",
-    "use-sync-external-store": "1.0.0-beta-fdc1d617a-20211118"
+    "react-is": "^18.0.0-rc.0",
+    "use-sync-external-store": "^1.0.0-rc.0"
   },
   "devDependencies": {
     "@babel/cli": "^7.12.1",
@@ -81,6 +77,7 @@
     "@testing-library/jest-dom": "^5.11.5",
     "@testing-library/jest-native": "^3.4.3",
     "@testing-library/react": "13.0.0-alpha.4",
+    "@testing-library/react-12": "npm:@testing-library/react@^12",
     "@testing-library/react-hooks": "^3.4.2",
     "@testing-library/react-native": "^7.1.0",
     "@types/object-assign": "^4.0.30",
@@ -104,9 +101,12 @@
     "jest": "^26.6.1",
     "prettier": "^2.1.2",
     "react": "18.0.0-beta-fdc1d617a-20211118",
+    "react-17": "npm:react@^17",
     "react-dom": "18.0.0-beta-fdc1d617a-20211118",
+    "react-dom-17": "npm:react-dom@^17",
     "react-native": "^0.64.1",
     "react-test-renderer": "18.0.0-beta-fdc1d617a-20211118",
+    "react-test-renderer-17": "npm:react-test-renderer@^17",
     "redux": "^4.0.5",
     "rimraf": "^3.0.2",
     "rollup": "^2.32.1",
--- a/src/components/Context.ts
+++ b/src/components/Context.ts
@@ -8,6 +8,7 @@ export interface ReactReduxContextValue<
 > {
   store: Store<SS, A>
   subscription: Subscription
+  getServerState?: () => SS
 }
 
 export const ReactReduxContext =
--- a/yarn.lock
+++ b/yarn.lock
@@ -8837,7 +8837,7 @@ __metadata:
   languageName: node
   linkType: hard
 
-"react-is@npm:18.0.0-beta-fdc1d617a-20211118, react-is@npm:^18.0.0-beta-fdc1d617a-20211118":
+"react-is@npm:18.0.0-beta-fdc1d617a-20211118":
   version: 18.0.0-beta-fdc1d617a-20211118
   resolution: "react-is@npm:18.0.0-beta-fdc1d617a-20211118"
   checksum: 402e28c80019e9deaee919c4373606736ea36142d1d1e09e0cab89d52e8bd647f1176a383987cca5c4e83bfee33bba8091ab347363ea4bdd4f9c3a707812c8b9
@@ -8855,6 +8855,13 @@ __metadata:
   version: 16.13.1
   resolution: "react-is@npm:16.13.1"
   checksum: 11bcf1267a314a522615f626f3ce3727a3a24cdbf61c4d452add3550a7875326669631326cfb1ba3e92b6f72244c32ffecf93ad21c0cad8455d3e169d0e3f060
+  languageName: node
+  linkType: hard
+
+"react-is@npm:^18.0.0-rc.0":
+  version: 18.0.0-rc.0-next-f2a59df48-20211208
+  resolution: "react-is@npm:18.0.0-rc.0-next-f2a59df48-20211208"
+  checksum: e0f8b1c8ef759153072b7508f8832164651aa855c9c939d607090899d9d522c68ad12ec9630f34a56c46330542356207dfc09d744b8ffe2eccc45a83f296fcae
   languageName: node
   linkType: hard
 
@@ -8966,7 +8973,7 @@ __metadata:
     react-17: "npm:react@^17"
     react-dom: 18.0.0-beta-fdc1d617a-20211118
     react-dom-17: "npm:react-dom@^17"
-    react-is: ^18.0.0-beta-fdc1d617a-20211118
+    react-is: ^18.0.0-rc.0
     react-native: ^0.64.1
     react-test-renderer: 18.0.0-beta-fdc1d617a-20211118
     react-test-renderer-17: "npm:react-test-renderer@^17"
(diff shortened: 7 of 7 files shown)


src/components/connect.tsx
```diff
@@ -609,6 +614,10 @@ function connect<
         ? props.store!
         : contextValue!.store
 
+      const getServerSnapshot = didStoreComeFromContext
+        ? contextValue.getServerState
+        : store.getState
```
Andarist, line 619: Do you plan to support server snapshots for a `!didStoreComeFromContext` case? I'm not interested in it myself but it would be a surprise to me that, as a user, I couldn't use this functionality when using a `store` prop

src/components/connect.tsx
```diff
@@ -736,7 +745,10 @@ function connect<
 
       try {
```
Andarist, line 746: q: when this can throw with `useSyncExternalStore`? shouldn't `useSyncExternalStore` be resilient to tearing etc that could make this sort of stuff throw pre-uSES?

src/hooks/useSelector.ts
```diff
@@ -44,13 +44,13 @@ export function createSelectorHook(
...
 
-    const { store } = useReduxContext()!
+    const { store, getServerState } = useReduxContext()!
 
     const selectedState = useSyncExternalStoreWithSelector(
       store.subscribe,
       store.getState,
       // TODO Need a server-side snapshot here
-      store.getState,
+      getServerState || store.getState,
```
Andarist, line 53: I wonder - do you maybe have a good reference for the `getServerState` snapshot API~? I've tried searching through WG discussions but I came empty-handed. I recall reading about this but would love to refresh myself on why this is needed with `uSES` and how React avoid the need for it with its primitives

src/utils/useIsomorphicLayoutEffect.ts
```diff
@@ -9,9 +9,11 @@ import { useEffect, useLayoutEffect } from 'react'
 // is created synchronously, otherwise a store update may occur before the
 // subscription is created and an inconsistent state may be observed
 
-export const useIsomorphicLayoutEffect =
+// Matches logic in React's `shared/ExecutionEnvironment` file
```
Andarist, line 12: nit: using an actual URL makes it easier for people to jump quickly to referenced files (if needed):
```suggestion
// Matches logic in React's `shared/ExecutionEnvironment` file
// https://github.com/facebook/react/blob/5cccacd131242bdea2c2fe4b33fac50d2e3132b4/packages/shared/ExecutionEnvironment.js#L10-L14
```

src/components/connect.tsx
```diff
@@ -34,7 +34,10 @@ import defaultMapStateToPropsFactories from '../connect/mapStateToProps'
 import defaultMergePropsFactories from '../connect/mergeProps'
 
 import { createSubscription, Subscription } from '../utils/Subscription'
-import { useIsomorphicLayoutEffect } from '../utils/useIsomorphicLayoutEffect'
+import {
+  useIsomorphicLayoutEffect,
+  canUseDOM,
```
Andarist, line 39: If I see correctly this is a leftover - this actually ain't used in this file atm.

src/components/Provider.tsx
```diff
@@ -19,14 +22,15 @@ export interface ProviderProps<A extends Action = AnyAction> {
...
+function Provider({ store, context, children, serverState }: ProviderProps) {
+  const contextValue: ReactReduxContextValue = useMemo(() => {
     const subscription = createSubscription(store)
     return {
       store,
       subscription,
+      getServerState: serverState ? () => serverState : undefined,
     }
-  }, [store])
+  }, [store, serverState])
```
Andarist, line 33: If I understand correctly the `serverState` should be an immutable value, used only for hydration purposes. Might be worth skipping this here? Or putting it in a ref (that would be read by `getServerState`) to make this more explicit? Obviously, that would break the rules of hooks but IMHO it's sometimes worth it to represent some constraints etc

src/hooks/useSelector.ts
eps1lon: > I've tried searching through WG discussions but I came empty-handed.

https://github.com/facebook/react/issues/22361 might help. Note that my initial post made some false assumptions that were corrected.

> I recall reading about this but would love to refresh myself on why this is needed with uSES

By the time a particular component is hydrated, the store might've already been mutated. Simplest example I encountered was something like

```jsx
<MutatesStoreInAnEffect />
<ReadsStore />
<React.Suspense>
  <ReadsStore />
</React.Suspense>
```

Assuming no component suspends, the server would render the same value for the store in every `ReadsStore`. But if you hydrate this tree then the second `ReadsStore` might be hydrated after the effects of `MutatesStoreInAnEffect` are run (due to selective hydration). The first `ReadsStore` would hydrate fine because we still have the initial state.

So we need an API to tell React with what state the markup was rendered on the server.

> how React avoid the need for it with its primitives

Not sure I understand this statement correctly but React does need `getServerSnapshot` in every version. It needs it even more in React 18. In React 17 you'd be safe against hydration mismatch if your store is never mutated during render. In React 18 you can even get hydration mismatches if your store is mutated in effects.

Andarist: Thank you for the link and the explanation. This is what I was roughly expecting. I didn't clearly understand how the client value is handled though, and according to Andrew, the "mismatched" component is just re-rendered on top of the hydrated component that was hydrated based on the server snapshot.

> Not sure I understand this statement correctly but React does need getServerSnapshot in every version. It needs it even more in React 18. In React 17 you'd be safe against hydration mismatch if your store is never mutated during render. In React 18 you can even get hydration mismatches if your store is mutated in effects.

I'm wondering about the situation when no external store is involved. Just replace `MutatesStoreInAnEffect` with `UpdatesStateInAnEffect` and `ReadsStore` with `ReadsPropOrContextValueThatComesFromReactState` in your example. This is basically the very same situation as the described by your, but `useState` doesn't support any `getServerSnapshot` method:
https://github.com/facebook/react/blob/5cccacd131242bdea2c2fe4b33fac50d2e3132b4/packages/react-reconciler/src/ReactFiberHooks.new.js#L2592-L2594

So I wonder, how the implementation is different for both and how React can handle this better for a plain state value?

Andarist: One thing I've learned from my past attempts at debugging React code is that I'm not that good at reasoning about such sublime code 🤣 

Maybe @acdlite would be able to shed some light on this.

src/components/connect.tsx
markerikson: Sort of the other way around. I made this change speculating I might need to use this in `connect` somehow, but I haven't made any actual code changes to `connect` yet - that's why this is a WIP PR.

src/hooks/useSelector.ts
acdlite: The reason useState "just works" is because you can't update a component until after it has already hydrated

acdlite:
A suggestion regarding `connect`:

What's nice thing about the `connect` API is that because it's a component factory/higher-order component, you can fork the implementation inside the factory instead of within the component's render body. You're already taking advantage of this by, for example, conditionally wrapping the `ConnectFunction` component with `React.memo`. But it looks like most the forking happens within `ConnectFunction`, inside the body of the component. Because of the hook rules, this means that a single component has to pay the cost of every possible hook that you might need to satisfy all the different `connect` API contracts.

One thing you could do to manage the complexity of these different variants is to lift the forked behavior out of the component and into the factory, like you already do with `React.memo`. Something like this:

```js
// This code is way simplified, obviously, just trying to illustrate the basic idea
function connect(mapStateToProps, mapDispatchToProps, mergeProps) {
  if (mapStateToProps == null) {
    // No selector provided
    // These don't need useSyncExternalStore at all.
    if (typeof mapDispatchToProps === 'function') {
      // Custom action creators
      return connectWithNoSelectorAndCustomActionCreators(mapDispatchToProps);
    } else if (mapDispatchToProps != null) {
      // Basic action creators
      const actionCreators = mapDispatchToProps;
      return connectWithNoSelectorAndBasicActionCreators(actionCreators);
    } else {
      // No action creators
      return connectNoOp(mapDispatchToProps);
    }
  } else {
    // Selector provided
    if (mergeProps === null) {
      const selector = mapStateToProps;
      if (typeof mapDispatchToProps === 'function') {
        // Custom action creators
        return connectWithCustomActionCreators(mapDispatchToProps);
      } else if (mapDispatchToProps != null) {
        // Basic action creators
        const actionCreators = mapDispatchToProps;
        return connectWithBasicActionCreators(mapDispatchToProps);
      } else {
        // No action creators
        return connectWithNoActionCreators(mapStateToProps);
      }
    }
  }

  // ...and so on
}
```

Then each branch can make different implementation trade offs, instead of needing to write a single component that handles every scenario. I think this would help with managing maintenance complexity. It would also be better for performance because each connected component is only paying the cost of the hooks that it needs. So for example, a `connect` with no `mapStateToDispatch` doesn't need a `useSyncExternalStore`.

acdlite:
> That's kinda what I was saying earlier - sure, it's possible to completely refactor / rewrite all that existing logic - it's "just" code! :) Question is how much effort it would take and how much benefit there would be.

Maybe there's a compromise where we can make it work for the "simple" cases, then fall back to the current implementation for anything more complex. That might simplify the upgrade strategy/messaging for existing apps:

- All existing code just works
- If you use `useSelector`, it also works with Concurrent React features
- If you use `connect` without edge case APIs, it also works with Concurrent React features
- If you use `connect` with edge case APIs, you may or may not run into subtle concurrency issues. Consider migrating to `useSelector` if you care about Concurrent React compatibility.

markerikson:
From Dan, a couple React 18 SSR demos that may be good to play with:

- https://codesandbox.io/s/kind-sammet-j56ro?file=/src/Html.js
- https://github.com/vercel/next-rsc-demo

Ephem (commented):
Nice work on this. `connect` is tricky indeed, but I think most of that doesn't relate to `getServerSnapshot` specifically, but supporting concurrency in general. 😅

To me it looks like this approach should get Redux to the first step described under "Hydrating Data" in Sebastians semi-new [Library Upgrade Guide: <script> (e.g. SSR frameworks)
](https://github.com/reactwg/react-18/discussions/114), that is, everything works when all state is emitted and available on the client before React hydration starts (just like today).

I think modifying Redux state with Suspense combined with streaming is where it's going to get really tricky, really fast, but also not something I'm sure Redux _should_ support and either way a later problem. 😄

markerikson:
@Ephem Thanks for the feedback!

Could I ask you for a favor?  The biggest thing I need atm is an example project that does Redux-based SSR, and ideally in conjunction with React 18 SSR, so that I have something to play with.

Dan linked me a couple React 18 SSR projects, which I pasted into a comment just above.

Could you either point me to a couple of Redux-based SSR projects so that I can smash the two together, or if you actually have time try to set up a Redux+React 18 project so that I can use it as a testbed?

I want to try to do this set of changes the right way, but I really don't have any meaningful React/Redux SSR experience to draw from.  If I try to figure out how to set up all that stuff from scratch myself it's going to take a while before I even have something that I can test against :)

Ephem:
I can't promise anything, but I'll try to hook you up with something basic at least! I've just started looking into React Query's React 18 SSR-support as well, so this is heavily on my mind now and I'll see if I can come up with some tests etc as well, though personal time is never in great supply. 😅

About those two examples, I don't think Redux should actually support either right now. Looks simple enough to rip out Suspense and progressive hydration from the first one and set up the base case I think _should_ be supported thought!

------

Below this is just a look at the demos and why I don't think they should be supported now. Lets classify them:

https://codesandbox.io/s/kind-sammet-j56ro?file=/src/Html.js - Streaming rendering _with Suspense data fetching and progressive hydration_
https://github.com/vercel/next-rsc-demo - React Server Components

Neither Suspense data fetching nor RSCs are officially supported in the React 18 release and I think both are going to be hard to support fully in Redux. I'll try to write up why in more detail at some point, but it basically boils down to Redux being a single big store where any part of the app might behave differently based on any part of the store, and on hydration/getServerSnapshot _all data needs to already be there_. While it can support streaming, the app can not start hydrating before Redux has a **full** `serverState` so to speak, which means it gets tricky with Suspense where the store changes _during rendering_.

Or if you think about [this great image](https://youtu.be/pj5N-Khihgc?t=1045) from Shaundais fantastic talk, it's hard to imagine how Redux could currently hydrate these slices individually. RSCs are even trickier since they also happen on page navigation etc, so it's not really just hydration anymore.

src/components/connect.tsx
```diff
@@ -736,7 +745,10 @@ function connect<
 
       try {
         actualChildProps = useSyncExternalStore(
+          // TODO We're passing through a big wrapper that does a bunch of extra side effects besides subscribing
           subscribeForReact,
+          // TODO This is incredibly hacky. We've already processed the store update and calculated new child props,
+          // TODO and we're just passing that through so it triggers a re-render for us rather than relying on `uSES`.
           actualChildPropsSelector,
           // TODO Need a real getServerSnapshot here
           actualChildPropsSelector
```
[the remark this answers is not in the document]
markerikson: Yeah, I'm totally not going to bother letting that get passed to connected components as a prop. This is too complex already as it is :)  Might be able to add that warning though.

I think you may be on the right track with the `childPropsSelector` suggestion!

Ephem:
I wanted to get it done sooner than later (had a kid-free evening! :tada:), so I threw together something very quick and dirty from that first example: https://codesandbox.io/s/gifted-monad-npltr?file=/server/render.js

It's rough, but should showcase both how progressive hydration lets us click the button at the top before comments has hydrated, and that it will currently cause a hydration-mismatch for the comments section because a store value has updated. (Tearing itself is expected here, can't update a part of the app that hasn't hydrated yet)

I think this is the case Redux should aim to support for React 18. Data is "fetched" ahead of rendering on the server, and is available before hydration starts on the client, but progressive enhancement of Suspense-based code splitting should work without hydration mismatches, even when state updates in parents.

Hope it helps, didn't have time to comment it as much as I wanted, so do ask if things are unclear. 😄

markerikson:
Sweet, thank you! May not be able to look at this for a bit due to other priorities, but I _really_ appreciate your expertise on Redux+SSR in general and you putting this together specifically!

markerikson:
Just rebased this against the latest type changes in `master`.

@Ephem : could I ask you to try converting that sandbox into a minimal standalone test file?  I can play around with the sandbox as an actual project some (and I'll try to do that a bit now), but ultimately I'm going to want to have some kind of test that verifies this behavior works.  Don't need it to pass yet :)  just needs to do the equivalent store setup + rendering + hydration, and ideally _fail_ based on React warning about hydration mismatch or something with `connect`, so that I can confirm that whatever changes I might make here fix the behavior.

markerikson:
Well hey, early good news. I cloned the SSR example repo, added a `store.dispatch(incrementCount())` on startup before initial render, and confirmed that I get a hydration warning with both 7.2.6 and 8.0.0-beta.1.

I then used `yalc` to add a build from this branch, updated `<App>` to pass through `<Provider serverState={initialState}>`, and confirmed that there is _no_ hydration warning!

So that looks like `useSelector` works okay.  Now to play around with `connect` and see what I can come up with...

markerikson:
Aaaand.... 

applying @Ephem 's suggestion to `connect` seems to work!

```ts
        actualChildProps = useSyncExternalStore(
          subscribeForReact,
          actualChildPropsSelector,
          // Use the server state if it's available
          getServerState
            ? () => childPropsSelector(getServerState(), wrapperProps)
            : actualChildPropsSelector
```

I added a second "count" button to the demo app that uses `connect` instead for the store state, verified that I see a hydration warning with the build from the PR, then made that change locally and verified that I do _not_ see a hydration warning.

This really needs more extensive testing, but given that I also see I screwed up the beta by adding some things into `deps` instead of `devDeps`, I think I'm going to go ahead and merge these changes now, and put out `beta.2`.  We can add more tests in a follow-up PR.

markerikson:
told you I really needed a viable demo to compare against :)

src/components/Provider.tsx
```diff
@@ -4,32 +4,40 @@ import { createSubscription } from '../utils/Subscription'
...
 }
 
 function Provider<A extends Action = AnyAction>({
   store,
   context,
   children,
+  serverState,
 }: ProviderProps<A>) {
   const contextValue = useMemo(() => {
     const subscription = createSubscription(store)
```
Andarist, line 34: q: this component still relies on a custom Subscription and custom notification propagation etc. Doesn't `useSyncExternalStore` solve the ordering issue?

markerikson: Nope, not at all. `uSES` still creates store subscriptions same as always, and React still runs effects bottom-to-top on mount.  So, children still subscribe before parents, etc, and we still have to do work on our side to enforce the "top-down updates" requirement for `connect`.

The only thing in React that has ever helped with that was when we switched v6 to read state updates from context. It _worked_.... which is why it was so frustrating that it also turned out to be the worst case for performance updates.

Andarist: > we still have to do work on our side to enforce the "top-down updates" requirement for connect.

Isn't the problem relevant to `useSelector` too? Is React subscribing to the "external store" (Redux) with `uSES` still in the order that makes shenanigans required? Is the zombie children problem still relevant with `uSES`? Or is the problem somewhat fixed with automatic batching and top-down rerender that should happen with it? As automatic batching should end up rerendering things in top-down order, right?

markerikson: Yes, the problem _is_ entirely relevant to `useSelector`, and there's absolutely nothing we can do to fix it - because hooks can't render context providers, so we can't enforce top-down behavior.  That's why I documented all the "zombie child" concepts in the hooks API page in the first place:

https://react-redux.js.org/api/hooks#stale-props-and-zombie-children

And no, batching still has nothing to do with this problem, because it's _the subscription callbacks themselves_ that are the problem, and those run _outside_ of React when the store is updated.

The hacky-but-valid workaround that both `useSelector` and `uSES` take is catching errors in a selector when subscribers are notified, forcing a re-render anyway, re-running the selector _during_ the render, and seeing if it explodes again or succeeds.

Andarist: Ok, I understand how automatic batching is not the remedy for the problem.

> That workaround is good enough that it does actually mostly mitigate the issue. But the root cause of the problem still exists - components can end up subscribing to the store in an order where a child is notified before its parent, and if the child depends on using props in its selector, those props can be stale or the child may be reading state right before its parent is about to stop rendering the child.

I must say that I'm somewhat disappointed with the `uSES` API then. I was under the impression that this problem would actually be solved by this API. This problem is affecting all external stores and thus the implementation that would guard users against this problem is still very complicated - so marketing `uSES` as a tool solving external store problems is somewhat misleading. I think that what it solves was way easier to solve in the userland (maybe apart from allowing inline selectors, as that was super tricky) than the zombie children problem - solving this problem at the React level would be the bigger "win" for the library authors.

@acdlite is solving this problem out of the scope of `uSES`? How are other external stores supposed to handle those situations? Should we use the hacky solution outlined above by @markerikson?

markerikson: FWIW `uSES` already implements that "hacky solution" :) and specifically inspired by `useSelector` doing it first.

Really the problem is a mismatch in data shapes: a linear array of subscriber callbacks executed in order, vs a tree of components where A) mount effects are executed bottom-to-top, and B) components can be added and removed over time.

Ephem:
Nice work! 👏 Glad the sandbox helped and yeah, I plan to add that and more tests, though I have some other things to look at first. 😀

There might be some more nuance here, I'm still thinking and exploring, I'll get back with any findings.

src/components/Provider.tsx
Andarist: > If there were some way to actually correlate "adding a subscriber" with "actual position in the React component tree", then it would be theoretically possible to build a smarter notification system that actually notifies subscribers in tree parent->child order. But there's no way to know where a given component is in the tree.

I would put the asterisk here that it's no way to know that in the userland - that's why I think this is the problem worth solving in the React itself (in the `uSES`).

> FWIW uSES already implements that "hacky solution" :) and specifically inspired by useSelector doing it first.

Hm, so I'm a little bit confused right now. I wonder if I understand the whole zombie children problem right now - I think I understand the problem but I might not understand how the problem is solved in React Redux today.

So to get us on the same page. This is what I think right now:
1. zombie children problem is caused~ by subscriber callbacks being called "out of order" because a child's callback can be called **before** the parent's callback - this leads to errors being thrown when items in the store are deleted during the latest dispatch~
2. the error is captured by React Redux and a new rerender is being scheduled
3. that new rerender is already executed when props are already updated and thus the selector should no longer throw or a rerender for that child might not happen at all because that component could have been unmounted by the render that has thrown (unmounted by its parent rerendering with the selected state that already doesn't contain the deleted item)
4. so in a sense - the whole zombie children problem is already "solved" by capturing errors and rerendering, the point is that users of React Redux don't have to deal with this. They might occasionally observe weird-ish things when debugging, but that's acceptable.
5. however, there is still the cost of the child being rendered at all as that could have been prevented if subscriptions would merely mark components as "dirty" and the state would actually be selected in render - with such a solution the child's subscriber wouldn't select anything that could throw (if only this would be batched and performed in the top-down order). Or if the item wasn't completely removed and the error is caused by partial lack of information etc then we get 2 renders there instead of one.

Is this reasoning correct? I think that I see where the `use-sync-external-store/shim` implements the try/catch dance and that looks good. It means that libraries using that won't have to deal with this problem at all because that's covered there. I wonder though - if the implementation of the "native" `useSyncExternalStore` the same or is the issue mitigated there because things can actually be executed in the "correct" order in React 18?

markerikson: Yep, that's all correct.  (fwiw I don't think the "cost" thing in step 5 is really a meaningful issue.)

You can see that the actual `useSyncExternalStore` does the same 'rerender if error" behavior here:

https://github.com/facebook/react/blob/3dc41d8a2590768a6ac906cd1f4c11ca00417eee/packages/react-reconciler/src/ReactFiberHooks.new.js#L1482-L1491

Andarist: > Yep, that's all correct. (fwiw I don't think the "cost" thing in step 5 is really a meaningful issue.)

I agree that this could have been somewhat exaggerated on my side. If this is only affecting subscribers and not rerenders then I guess this is somewhat OK (I assume that the inner component might in fact not even rerender, even if it throws, if it gets "unmounted" by the parent). However, I still find it a little bit strange that React 18 doesn't try to mitigate this all together in a smarter way - but dunno, maybe the smarter way would have some perf implications for the whole thing.

PR #1838, opened 2021-11-03
Current links to connect-mapstate lead to "404 Not Found". They have been replaced with the correct one.

Ecool88:


timdorr:
Fixed a couple of those updated links, which need to go to the file itself. Thanks!

PR #1851, opened 2021-12-16
fix: Add types to ensure that there will be no compilation errors when struct: false

duan602728596:
I used the v8 version of react-redux in the project.

When typescript is configured with `strict: true`, The type of `let listeners = []` is recognized as `any[]`.

When typescript is configured with `strict: false`, The type of `let listeners = []` is recognized as `never[]`, 
This will cause `listeners.push(listener)` to prompt the error `TS2345: Argument of type 'Listener' is not assignable to parameter of type'never'.`.

When project is not configured with `strict: true`, this will cause a project compile-time error, so I added a type here to solve this problem.

### DIFF
--- a/src/utils/Subscription.ts
+++ b/src/utils/Subscription.ts
@@ -34,7 +34,7 @@ function createListenerCollection() {
     },
 
     get() {
-      let listeners = []
+      let listeners: Listener[] = []
       let listener = first
       while (listener) {
         listeners.push(listener)


timdorr:
Thanks!

markerikson:
Huh. Yeah, the added type for the array is useful here, but I'm also curious how you even ran into this. This is an internal structure that isn't exposed to consumers of the library. Did you have your app configured to try to compile dependencies out of `node_modules` or something?

duan602728596:
> Huh. Yeah, the added type for the array is useful here, but I'm also curious how you even ran into this. This is an internal structure that isn't exposed to consumers of the library. Did you have your app configured to try to compile dependencies out of `node_modules` or something?

Maybe I didn't ignore `node_modules`, so I ran into this problem.
This is a demo that I can reproduce: https://github.com/duan602728596/react-redux-build-project.
In this demo, the files in `node_modules/react-redux/src` were checked during compilation and an error was given.
This is the case with my project.

markerikson:
Yeah, I would normally advise _against_ trying to compile or type-check any code in `node_modules` :)

duan602728596:
> Yeah, I would normally advise _against_ trying to compile or type-check any code in `node_modules` :)

Yes, I will pay attention to these in the future.

Issue #464, opened 2016-08-18
Binding Store getState and subscribe calls properly in @next version


vhmth:
I noticed, per following the conversation [here](https://github.com/reactjs/react-redux/pull/416), that a rewrite of `connect` would fix the bottom-up state updates I was seeing in my nested container components.

I went ahead and tried to make that work with the [react-chrome-redux](https://github.com/tshaddix/react-chrome-redux) lib which creates a simple Store object from the content scripts of Chrome extensions (using an es6 class).

I noticed that the `this` context within the [react-chrome-redux Store](https://github.com/tshaddix/react-chrome-redux/blob/master/src/store/Store.js#L57) was `undefined` and noticed that the new rewrite of `connect` was not respecting the store's own context when calling [getState](https://github.com/reactjs/react-redux/blob/next/src/components/connectAdvanced.js#L162) and [subscribe](https://github.com/reactjs/react-redux/blob/next/src/utils/Subscription.js#L8).

Since the previous version of `connect` used to respect this context in [subscribe](https://github.com/reactjs/react-redux/blob/master/src/components/connect.js#L199) and [getState](https://github.com/reactjs/react-redux/blob/master/src/components/connect.js#L244), I believe we should respect it in the new version of `connect` as well (unless there's a good reason not to - in which case it should probably be stated as a potential breaking change for any libs that roll their own Stores that rely on having a local context).

PR #1857, opened 2021-12-23
Avoid allocating a redundant function in the `Provider`

Andarist:


src/components/Provider.tsx
```diff
@@ -39,7 +39,7 @@ function Provider<A extends Action = AnyAction>({
     }
   }, [store, serverState])
 
-  const previousState = useMemo(() => store.getState(), [store])
+  const previousState = useMemo(store.getState, [store])
```
Andarist, line 42: It seems that it breaks custom stores that rely on `this`. A test covering this use case was originally introduced [here](https://github.com/reduxjs/react-redux/pull/465) and fixed [this issue](https://github.com/reduxjs/react-redux/issues/464)

Given how rare of a use case this is and the allocation cost that this requires for all users I would say that it would be better to drop support for this use case and require users who need this to bind their stores appropriately. It's an easy thing to do and feels to me like something that can be easily handled in the userland.

OTOH `Provider` itself is, at least currently, usually not rendered often as it lives "on the top" of the app. So kinda :shrug:

markerikson: Typically `Provider `is only ever going to render once in an app anyway. Given that, I'm not sure this change really provides any benefit :)

markerikson:
I'm gonna pass on this one. I can see the argument for dropping a meaningless use case, but also I don't think there's enough actual benefit here to justify the change.

Andarist:
Fine by me 👍

PR #1858, opened 2021-12-23
Make VScode to always use the local TS version

Andarist:
This is something that is very annoying with VScode - by default, it's always using the TS version that is bundled in it, instead of using what the project is actually using. This can lead to different things being observed in the IDE and when running `tsc` or similar so I highly recommend adding this file to your project.


markerikson:
Oh neat, sure!

PR #1859, opened 2021-12-23
Cleanup more code after `pure` removal

Andarist:
Those code lines were redundant since this option is not supported anymore:
https://github.com/reduxjs/react-redux/blob/93d5ee5e1fd94b76040645336d483e9f9574de20/src/components/connect.tsx#L477-L483

### DIFF
--- a/src/components/connect.tsx
+++ b/src/components/connect.tsx
@@ -528,7 +528,6 @@ function connect<
     const displayName = `Connect(${wrappedComponentName})`
 
     const selectorFactoryOptions: SelectorFactoryOptions<any, any, any, any> = {
-      pure,
       shouldHandleStateChanges,
       displayName,
       wrappedComponentName,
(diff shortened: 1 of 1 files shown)


markerikson:
This is actually intentional.  In v6 we added some similar checks for options that had just been removed, specifically to help users who may have been using those options know that behavior was changing.  I did the same thing here.

Andarist:
@markerikson this PR doesn't remove that check - it removes "unreachable" code that depends on this already removed option. The user can't use this option anyway because using it throws an error.

markerikson:
Hah, my bad, sorry :)

src/components/connect.tsx
```diff
@@ -543,11 +542,6 @@ function connect<
       areMergedPropsEqual,
     }
 
-    // If we aren't running in "pure" mode, we don't want to memoize values.
-    // To avoid conditionally calling hooks, we fall back to a tiny wrapper
-    // that just executes the given callback immediately.
-    const usePureOnlyMemo = pure ? useMemo : (callback: () => any) => callback()
```
Andarist, line 549: Note that this actually changes behavior but ive assumed that to the more preferred one. The alternative is to inline the memoized value - as thay was effectively happening before my changes

Andarist:
@markerikson friendly 🏓

PR #1865, opened 2022-01-26
chore: add comments

HiWayne:


markerikson:
This PR appears to be based off an old version of React-Redux. Additionally, there's some Git conflict markers, and the comments don't appear to be in English.  I'm going to have to close this.

PR #1866, opened 2022-01-28
7.x fixbug useSelector

CosSalt:
bugfix，when value is NaN, useSelector will rerender


timdorr:
Thanks, but this isn't needed. You're already able to pass in your own `equalityFn` to `useSelector` to customize it beyond a simple referential equality check.

PR #1867, opened 2022-01-28
useSelector rerender

CosSalt:
bugfix，when value is NaN, useSelector will rerender


timdorr:
https://github.com/reduxjs/react-redux/pull/1866#issuecomment-1024775898

PR #1870, opened 2022-02-05
Switch back to Subscription in useSelector to fix unsubscribe perf

### LINKED ISSUE #1869: [v8] useSelector without custom subscription manager is slower than v7 (performance regression)

- React: 18.0.0-rc0
- ReactDOM: 18.0.0-rc0 
- Redux: 4.1.2
- React Redux: 8.0.0-beta2

In React-Redux v8 the custom `Subscription` manager was removed from `useSelector`, 
and `store.subscribe` from Redux is used directly in `useSyncExternalStore` call.

This will bring back a performance problem (#1450), because Redux unsubscription relies on indexOf\splice.

React 18 + React-Redux v8.0.0
Screenshot 2022-02-05 01 36 43

React 17 + React-Redux v7.2.6
Screenshot 2022-02-05 01 24 30

Test code:
```javascript
const store = createStore(
);

function App() {
    const state = useSelector(s => s);
    const dispatch = useDispatch();

    const [children, setChildren] = useState(0);

    const toggleChildren = () => setChildren(c => c ? 0 : 50000);

    return <div>
        <button onClick={toggleChildren}>Toggle Children</button>
        <button onClick={() => dispatch(increment())}>Increment</button>
        {[...Array(children).keys()].map(i => <Child key={i}/>)}
    </div>;

}

function Child() {
    const v = useSelector(s => s);
    return <div className="child">{v}</div>;
}
```

The `Subscription` manager used in v7 has better performance than Redux unsubscription used in v8 with many listeners.
It's still in the bundle, because `<Provider>` uses it, so it could be used again in `useSelector` too.

_No response_

- [X] Yes

markerikson:
The Subscription impl originally used an array of listeners just
like the actual Redux store, but in #1523 we changed it to be a
linked list to fix perf issues. The use of `indexOf+splice` caused
a quadratic cost to unmounting many components.

In v8, I originally dropped use of `Subscription` to potentially
save on bundle size. However, `Provider` still uses `Subscription`,
so that will always be in the bundle.

For this issue, a user pointed out that directly subscribing to the
store brought back the original quadratic unsubscription behavior,
which I've confirmed locally.

Swapping `store.subscribe` for `subscription.addNestedSub` fixes
that issue.

I've added a unit test that mounts and unmounts a massive tree and
measures the elapsed time. There's a distinct difference between
the "correct" and quadratic behavior. Weirdly, there's also a big
diff in "correct" time between React 18, and 17 + the "compat" entry
point, and I have no idea why.  It's not as bad as the quadratic
time, but it's still pretty expensive.

I've written the test to set an acceptable max unmount time based on
which React version we're running against, with some buffer added.

Fixes #1869 .

### DIFF
--- a/test/hooks/useSelector.spec.tsx
+++ b/test/hooks/useSelector.spec.tsx
@@ -1,14 +1,23 @@
 /*eslint-disable react/prop-types*/
 
-import React, { useCallback, useReducer, useLayoutEffect } from 'react'
+import React, {
+  useCallback,
+  useReducer,
+  useLayoutEffect,
+  useState,
+  useContext,
+} from 'react'
 import { createStore } from 'redux'
 import * as rtl from '@testing-library/react'
 import {
   Provider as ProviderMock,
   useSelector,
+  useDispatch,
   shallowEqual,
   connect,
   createSelectorHook,
+  ReactReduxContext,
+  Subscription,
 } from '../../src/index'
 import type {
   TypedUseSelectorHook,
@@ -29,7 +38,6 @@ describe('React', () => {
       let renderedItems: any[] = []
       type RootState = ReturnType<typeof normalStore.getState>
       let useNormalSelector: TypedUseSelectorHook<RootState> = useSelector
-      type VoidFunc = () => void
 
       beforeEach(() => {
         normalStore = createStore(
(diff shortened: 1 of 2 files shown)


test/hooks/useSelector.spec.tsx
```diff
@@ -660,6 +640,69 @@ describe('React', () => {
...
+          // Seeing an unexpected variation in elapsed time between React 18 and React 17 + the compat entry point.
+          // With 18, I see around 75ms with correct implementation on my machine, with 100K items.
+          // With 17 + compat, the same correct impl takes about 4200-5000ms.
+          // With the quadratic behavior, this is at least 13000ms (or worse!) under 18, and 22000ms+ with 17.
+          // The 13000ms time for 18 stays the same if I use the shim, so it must be a 17 vs 18 difference somehow,
+          // although I can't imagine why, and if I remove the `useSelector` calls both tests drop to ~50ms.
+          // So, we'll modify our expectations here depending on whether this is an 18 or 17 compat test,
+          // and give some buffer time to allow for variations in test machines.
+          const expectedMaxUnmountTime = IS_REACT_18 ? 500 : 7000
+          expect(elapsedTime).toBeLessThan(expectedMaxUnmountTime)
```
timdorr, line 704: I do worry about the stability of this. This seems like a future case of "works on my machine!" being a false-positive result. 

I don't think so based on how things are set up, but would it be possible to compare the React 17 vs 18 times to have an acceptable ratio instead? That would maintain stability between different CPU speeds.

markerikson: I don't particularly like it either :)  But the "incorrect" times I was seeing were 13.5s for 18 and 23s for 17, and given the number of children being unmounted I see that as a big enough gap that I don't think we're likely to have a false positive due to machine/environment variance.

I do kinda want to investigate _why_ it's slower with React 17 a bit further. But the immediate goal for this test is just to specifically document that "we expect unsubscribing lots of components to not be quadratic" so that we don't forget in the future, and catch that if we were to ever somehow lose that behavior.  I think the current test at least accomplishes that.

markerikson: Worst case, the test randomly starts failing due to CPU/env stuff, and we change how it works then.

markerikson:
Merging now, will put out another v8 beta in a bit.

PR #1884, opened 2022-04-02
Swap `useSyncExternalStore` shim behavior and update React deps

markerikson:
This PR:

- Updates all React-related dependencies to React 18 final
- Updates React Testing Library
- Swaps the shim behavior so that the primary entry point now uses the `useSyncExternalStore` shim behavior as the default
- Changes the old `"/compat"` entry point so that it is now `"/next"` and imports `uSES` directly from React without using the shim
- Updates the Jest test setup to test all three scenarios (main entry + React 18, main entry + React 17, next entry + React 18)
- Adds TS 4.6 to the TS test matrix


markerikson:
Just did a quick bundle size comparison of React-Redux 7.2.8, 8.0-beta.3 (with no `uSES` shim), and a build from this branch (with `uSES` shim).  The setup is a Vite app with React 18 and RTK 1.8.1, and a store with a basic counter slice and `<Counter>` component.  

As best as I can tell from `source-map-explorer`, the vendor chunk sizes (including _all_ of React, ReactDOM, RTK, and Immer) are:

- 7.2.8: 167.33K
- 8.0-b3: 167.21K
- 8.0-b4: 167.99K

In other words:

- 8.0-b3 is 120b smaller than 7.x, minified
- 8.0-b4 is 600b _larger_ than 7.x, minified

But, that 600 bytes gets us compat with React 16.8/17, and makes it possible for people to easily upgrade to React-Redux v8 even if they aren't using React 18 yet.  It also ensures compat with React Native, which won't support React 18 until the "New Architecture" is fully released:

https://reactnative.dev/blog/2022/03/30/version-068#opting-in-to-the-new-architecture

So, that slight increase in bundle size is well worth it.

Also, users who _are_ on React 18 can alias the `"/next"` entry point and drop the shim from their bundle.

PR #1893, opened 2022-04-11
fix(deps): add optional peer deps into `peerDependencies`

kyletsang:
In the [NPM docs](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#peerdependenciesmeta), optional peer deps should be specified in `peerDependencies` because this is where the versioning info comes from.  This is the example from the NPM docs:

```
{
  "name": "tea-latte",
  "version": "1.3.5",
  "peerDependencies": {
    "tea": "2.x",
    "soy-milk": "1.2"
  },
  "peerDependenciesMeta": {
    "soy-milk": {
      "optional": true
    }
  }
}
```

One thing I wasn't sure about was the react native compatibility, so I placed the version that's currently in devDeps as a placeholder.  I'll fix this once I get confirmation.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,10 @@
     "coverage": "codecov"
   },
   "peerDependencies": {
-    "react": "^18.0.0"
+    "@types/react": "^18.0.0",
+    "react": "^18.0.0",
+    "react-dom": "^18.0.0",
+    "react-native": "^0.64.1"
   },
   "peerDependenciesMeta": {
     "@types/react": {
--- a/yarn.lock
+++ b/yarn.lock
@@ -8931,7 +8931,10 @@ __metadata:
     typescript: ^4.3.4
     use-sync-external-store: ^1.0.0
   peerDependencies:
+    "@types/react": ^18.0.0
     react: ^18.0.0
+    react-dom: ^18.0.0
+    react-native: ^0.64.1
   peerDependenciesMeta:
     "@types/react":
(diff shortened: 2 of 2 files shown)


markerikson:
I really don't think we want to do this, especially given how NPM currently insists on installing those automatically and throws errors if there's a mismatch.

@timdorr , @Methuselah96 : thoughts here?

markerikson:
We've specifically rejected this change several times in the past:

- #1390
- #1425 

(and see related links)

But I do see some mentions of NPM 6 in those threads, so not entirely sure where things stand on that now.

kyletsang:
Does the peerDepsMeta field actually work with the package version info missing from the peerDeps field?  My understanding was these were to go hand in hand.

I believe the NPM 6 functionality was to warn in the console.  NPM 7+ will try to resolve the dependencies and reject if peer deps aren't installed, but specifying "optional" in the peerDepsMeta field should bypass that.

Methuselah96:
I think this is correct. I believe all the previous issues/PRs were created before NPM supported optional peer dependencies. My understanding is that npm 7+ will [not install optional peer dependencies](https://github.com/npm/rfcs/discussions/289#discussioncomment-131068) automatically, however I don't actively use npm, so it might be worth double-checking that this works as expected on npm 6 and npm 7+.

markerikson:
Thank you!

PR #1902, opened 2022-04-16
Update docs for v8 final

markerikson:
- Removed `connectAdvanced` and `pure` descriptions
- Updated all `render()` examples to show `createRoot`
- Added note that `batch` isn't needed with React 18
- Added TS props for `Provider` and documented `serverState`
- Updated TS usage descriptions
- Added warnings about `@types/react@18`


markerikson:
Let's do this!

silviogutierrez:
@markerikson : I guess I'm not fully understanding this line:

> In this example, the client has received HTML rendered by the server, as well as a serialized Redux state attached to `window`. The serialized state is used to both pre-fill the store's contents, _and_ passed as the `serverState` prop to `<Provider>`

Why must we initialize the store in `configureStore` with the initial server state, *and* then also pass it into the provider?

I ask because 99% of the Redux Tool Kit docs show `configureStore` as a static call with a singular `export const store = configureStore(...)`. And yet, if we want SSR, all these examples need to be something like:

```typescript

export const initStore = (state?: RootState) => {
    return configureStore({reducer: {}, preloadedState: state});
}

// Further complicated
export type RootState = ReturnType<ReturnType<typeof initStore>["getState"]>;
```

And more. To me, it seems far better to just allow store to remain a static export and take `serverState` passed into `Provider` as the initial state when hydrating.

I hope that makes sense. Passing the state twice is a bit of a hassle, but not insurmountable. But switching the store export from static to a function makes a lot of the docs harder to pass when using SSR.

Thanks a ton for your Redux and toolkit work! It's gotten far easier to use.

phryneas:
@silviogutierrez in a SSR scenario, you have one process that potentially renders 100 different pages - and during rendering them, it might interact with the store, dispatching actions and awaiting side effects from `getServerSideProps` etc.

With a "static export", all of these could never start with a clean store, but with whatever store the last SSR-rendered page had left over - and that mixed up state would in the end moved over to the clients. You can't do that. => in a SSR environment you need that `createStore` function so that every rendered page can have their own store, even if multiple pages are rendered in parallel from the same process.

markerikson:
Yeah, two answers here:

- The SSR state should be passed to `<Provider>` so that a call to the new React `hydrateRoot()` API has a single consistent state value to use through the entire initial hydration render pass, without any chance of a store dispatch changing that value halfway through (which would cause React to error due to a hydration mismatch)
- Yes, for SSR scenarios, you _do_ need to have some kind of a `makeStore` function, and I've written those myself.  But, as Lenz said, we can't show every possible setup scenario in our main examples.  I do want to add an example of this sort of thing to our docs at some point.

silviogutierrez:
@markerikson @phryneas 

Thanks for the explanation. That makes sense. This whole time, I thought `<Provider serverState={foo} />` was code you executed on... the server. Not on the client.

Below are some thoughts that I realize are way beyond the scope of this, and unlikely to happen. Nonetheless, it may help to write it out.

I really think trying to hold off on a dynamic store initialization has a lot of benefit. It keeps SSR in-line with 99% of the RTK and redux docs, vastly simplifies typing, and reduces import / file clutter. For example, you immediately run into the circular `RootState` problem when you create a helper `initStore` function that accepts optional initial state. You have to do the `combineReducers` chicken dance to get proper typing.

And I think the way to do this is what I accidentally tried to do before in this thread — state in the provider.

Yes, each request needs a store on the server, but the foundation of redux is immutability. So you shouldn't be mutating *anything* in the store based on requests. Only the initial state should vary.

So if `<Provider>` had a way to pass in initialState on the server, we would cover 99% of use cases. Of course, this would need `store` to have some way `replaceState` and return another `store` instance with the replaced state. No such thing exists and I can't even monkey patch it because `createStore` in redux has no visibility into `state` by design.

So pseudo code for this magical provider:

```typescript
const Provider = (props) => {
    const store = props.initialState ? props.store.replaceState(props.initialState) : props.store;

   // ... rest of Provider code.
};
```

Happy to elaborate on this if there's interest.

markerikson:
@silviogutierrez tbh I'm not sure what you're actually trying to suggest here :)  But also this issue is the wrong place to talk about possible feature changes.  Could you post this as a "Discussions" thread instead?

Issue #1302, opened 2019-06-05
mergeProps example will never shallowEqual


davidjbradshaw:
**Do you want to request a _feature_ or report a _bug_?**

Documentation bug

**What is the current behavior?**

The example of `mergeProps` in the documentation creates a new function each time it is called, so the `shallowEqual` test will always fail leading to the component always rerendering.

This is the current example:
```js
import * as actionCreators from './actionCreators'

function mapStateToProps(state) {
  return { todos: state.todos }
}

function mergeProps(stateProps, dispatchProps, ownProps) {
  return Object.assign({}, ownProps, {
    todos: stateProps.todos[ownProps.userId],
    addTodo: (text) => dispatchProps.addTodo(ownProps.userId, text)
  })
}

export default connect(mapStateToProps, actionCreators, mergeProps)(TodoApp)
```

**What is the expected behavior?**

I would expect the component to update only if `ownProps.userId` changes.

**Suggested fix**

If you're happy with this I can make a PR, but not sure if your happy to use `Ramda` in an example.
```js
import { identity, memoizeWith } from 'ramda'
import * as actionCreators from './actionCreators'

const memoize = memoizeWith(identity)
const addTodo = memoize(
  (userId, dispatchProps) => (text) => dispatchProps.addTodo(userId, text)
)

function mapStateToProps(state) {
  return { todos: state.todos }
}

function mergeProps(stateProps, dispatchProps, ownProps) {
  return Object.assign({}, ownProps, {
    todos: stateProps.todos[ownProps.userId],
    addTodo: addTodo(ownProps.userId, dispatchProps)
  })
}

export default connect(mapStateToProps, actionCreators, mergeProps)(TodoApp)
```

**Which versions of React, ReactDOM/React Native, Redux, and React Redux are you using?  Which browser and OS are affected by this issue? Did this work in previous versions of React Redux?**
7

markerikson:
I think that's actually kind of the point - creating new values in `mergeProps` _will_ cause re-renders.  That's one of the reasons why we generally discourage use of `mergeProps`.

If anything, I'd rather add a note that this is _not_ a recommended pattern, and you need to be cautious when using it.

davidjbradshaw:
But doesn't `mergeProps` get called every time the `store.subscribe` function in Redux gets called, in the same way that `mapState` does? 

[...]

PR #1905, opened 2022-04-18
add note for mergeProps

phapdinh:
this adds a note under the returns of mergeProps that any new values will cause a re-render and it is best to memoize in order to avoid that. Part of issue #1302


markerikson:
Thanks!

PR #1910, opened 2022-04-19
fix(ts): incorrect import of `Subscription` causes `noImplicitAny` error

### LINKED ISSUE #1909: TypeScript error when upgrading to react-redux 8.0.0

- React: 16.14.0
- ReactDOM/React Native: 16.14.0
- Redux: 4.2.0
- React Redux: 8.0.0

When compiling a TypeScript project that has the `noImplicitAny` config set to `true`, then react-redux `.ts` files are typechecked and result in a compilation error 

```
node_modules/react-redux/src/utils/Subscription.ts:95:8 - error TS7011: Function expression, which lacks return-type annotation, implicitly has an 'any[]' return type.

95   get: () => [],
          ~~~~~~~~


Found 1 error in node_modules/react-redux/src/utils/Subscription.ts:95
```

I believe the `ts` files should be stripped from the final bundle that is uploaded to npm in order to avoid such ts config mismatch.

There shouldn't be any compilation error. 

N/A

- [X] Yes

vicrep:
Closes #1909 , see [this comment](https://github.com/reduxjs/react-redux/issues/1909#issuecomment-1102907378) for context.

### DIFF
--- a/src/exports.ts
+++ b/src/exports.ts
@@ -27,7 +27,7 @@ import { useSelector, createSelectorHook } from './hooks/useSelector'
 import { useStore, createStoreHook } from './hooks/useStore'
 
 import shallowEqual from './utils/shallowEqual'
-import type { Subscription } from '../src/utils/Subscription'
+import type { Subscription } from './utils/Subscription'
 
 export * from './types'
 export type {


phryneas:
That's a good catch! @markerikson I think that might warrant a 8.0.1? 🎉

phryneas (approved):

timdorr:
Thanks!

PR #1922, opened 2022-05-20
Show warning instead of throwing error that pure option has been removed

### LINKED ISSUE #1914: Impossible to create a lib compatible with version 7 and 8

- React: 18
- ReactDOM: 18
- Redux: 4.2
- React Redux: 7 or 8

When we want to create a lib with a peer dependency to react-redux we face a problem.
People that use react 18 will provide react-redux 8. People that work with react 17 will provide react-redux 7.

The lib is doing basic things, but it uses the connect with pure option to true.
In v8, pure option is no more necessary but it throws an error to say the option is not necessary.
This exception makes the full app crash.
It is not possible to remove the option to let v7 correctly works.

We have this case in an unmaintained lib react-beautifull-dnd that blocks the migration of apps to react 18.

A console warning can be displayed but not a blocking error, mostly if this parameter is not used at all.

All

- [X] Yes

ApacheEx:
This PR unblocks users to upgrade react-redux to the latest version with having other dependent libraries like react-beautiful-dnd which are in maintenance mode.

Resolves #1914

p.s it also fixes one warning which was hidden as restore of console.error mock was not done.

### DIFF
--- a/src/components/connect.tsx
+++ b/src/components/connect.tsx
@@ -28,6 +28,7 @@ import { mergePropsFactory } from '../connect/mergeProps'
 import { createSubscription, Subscription } from '../utils/Subscription'
 import { useIsomorphicLayoutEffect } from '../utils/useIsomorphicLayoutEffect'
 import shallowEqual from '../utils/shallowEqual'
+import warning from '../utils/warning'
 
 import {
   ReactReduxContext,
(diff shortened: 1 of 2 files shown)


markerikson:
Thank you! I'll see if I can push out 8.0.2 tonight or tomorrow.

paztis:
Thanks !!!

Issue #1940, opened 2022-08-11
Connect HOC misses store changes when "offscreen"


bvaughn:
### What version of React, ReactDOM/React Native, Redux, and React Redux are you using?

- React: 0.0.0-experimental-e7d0053e6-20220325
- ReactDOM: 0.0.0-experimental-e7d0053e6-20220325
- Redux: 4.2.0
- React Redux: 8.0.2

Here is a Replay recording of this bug with some annotations from me:
http://app.replay.io/recording/replay-failing-reactdevtools-e2e-repro--3351989d-95d4-4035-80af-07d813a623aa

Here is a Loom (it's long) where I use Replay to investigate the problem:
https://www.loom.com/share/3d60c14466184b03acc9af7b80e9eff8

Also note that this issue is specific to an experimental React API, `Offscreen`. The React team has [published some early documentation](https://github.com/reactwg/react-18/discussions/19) about this API but it has not been released in a stable form yet.

### What is the current behavior?

A _connected_ component misses updates that are fired while it is inside of a hidden `Offscreen` component.

I believe this is because the component unsubscribes when it is _unmounting_ and does not check for missed updates when it is later remounted. (The selectors passed tot he map-state-to-props function were not executed.)

### What is the expected behavior?

The _expected_ behavior is perhaps up for debate. Should the component process Store updates (at a lower priority) while hidden or should it wait to render them only if it's later re-shown? In a sense, this question isn't worth debating _yet_ because I don't think React currently provides an API (even an experimental one) to _stay subscribed_ while in a hidden state.

That being said, I think the currently expected behavior is that when then `Offscreen` tree is later _re-shown_, React Redux `connect` should check the Store for the latest values and re-render if they have changed since the component was last rendered.

### Which browser and OS are affected by this issue?

Presumably all of them

### Did this work in previous versions of React Redux?

- [ ] Yes

Andarist:
> In a sense, this question isn't worth debating yet because I don't think React currently provides an API (even an experimental one) to stay subscribed while in a hidden state.

Do you think such an API is a good idea? I got an impression that the whole strict effects stuff was introduced specifically to avoid subscriptions in hidden components etc. 

---

As to the shared Replay - I couldn't resist taking a look at it 😅 I'm not sure if I understand the issue correctly in full so take my findings with the grain of salt.

[...]

PR #1941, opened 2022-08-14
Remove redundant `isMounted` ref from `connect`

Andarist:
*Most likely* this should fix https://github.com/reduxjs/react-redux/issues/1940 . It would be cool if @markerikson or @bvaughn could test this out using the parent of this commit in Replay: https://github.com/replayio/devtools/commit/477dd472eebf5210213500e798f6c94a7fad8c86

### DIFF
--- a/src/components/connect.tsx
+++ b/src/components/connect.tsx
@@ -103,7 +103,6 @@ function subscribeUpdates(
   lastWrapperProps: React.MutableRefObject<unknown>,
   lastChildProps: React.MutableRefObject<unknown>,
   renderIsScheduled: React.MutableRefObject<boolean>,
-  isMounted: React.MutableRefObject<boolean>,
   childPropsFromStoreUpdate: React.MutableRefObject<unknown>,
   notifyNestedSubs: () => void,
   // forceComponentUpdateDispatch: React.Dispatch<any>,
@@ -118,7 +117,7 @@ function subscribeUpdates(
 
   // We'll run this callback every time a store subscription update propagates to this component
   const checkForUpdates = () => {
-    if (didUnsubscribe || !isMounted.current) {
+    if (didUnsubscribe) {
       // Don't run stale listeners.
       // Redux doesn't guarantee unsubscriptions happen until next dispatch.
       return
@@ -622,17 +621,8 @@ function connect<
       const lastWrapperProps = useRef(wrapperProps)
       const childPropsFromStoreUpdate = useRef<unknown>()
       const renderIsScheduled = useRef(false)
-      const isProcessingDispatch = useRef(false)
-      const isMounted = useRef(false)
 
       const latestSubscriptionCallbackError = useRef<Error>()
-
-      useIsomorphicLayoutEffect(() => {
-        isMounted.current = true
-        return () => {
-          isMounted.current = false
-        }
-      }, [])
 
       const actualChildPropsSelector = useMemo(() => {
         const selector = () => {
@@ -677,7 +667,6 @@ function connect<
             lastWrapperProps,
             lastChildProps,
             renderIsScheduled,
-            isMounted,
             childPropsFromStoreUpdate,
             notifyNestedSubs,
             reactListener


src/components/connect.tsx
```diff
@@ -622,18 +621,9 @@ function connect<
       const lastWrapperProps = useRef(wrapperProps)
       const childPropsFromStoreUpdate = useRef<unknown>()
       const renderIsScheduled = useRef(false)
-      const isProcessingDispatch = useRef(false)
```
Andarist, line 625: this one is a bonus removal - this ref stayed completely unused

markerikson:
AHHHH that makes sense!

src/components/connect.tsx
markerikson: current-me has no clue what past-me was thinking for this variable :)

Andarist:
I guess the fix is not fully correct since tests fail 🤣 but nevertheless I think that it would fix the mentioned issue. It would be interesting to understand how the subscription nesting works and how it depends on this `isMounted` stuff to figure out a solution that would cover both requirements.

Andarist:
So basically with this change, we have 2 affected tests:
- [should not notify nested components after they are unmounted](https://github.com/reduxjs/react-redux/blob/720f0ba79236cdc3e1115f4ef9a7760a21784b48/test/components/connect.spec.tsx#L2062)
- [should enforce top-down updates to ensure a deleted child's mapState doesn't throw errors](https://github.com/reduxjs/react-redux/blob/720f0ba79236cdc3e1115f4ef9a7760a21784b48/test/components/connect.spec.tsx#L3355)

The first one fails because `mapStateToProps` gets called more times - it's not ideal but *potentially* wouldn't necessarily have to be treated as a big problem. However, the second one is more concerning and after some light analysis, I think that actually both failures are related to the same thing. I could be mistaken though.

So what's the problem? The problem here is that `mapStateToProps` for a given component is called before that component has a chance to get unsubscribed. This again is related to the usage of a layout effect to manage `isMounted`. With that, the component was treated as unmounted "sooner" and the notification from the parent (originating in its layout effect where `captureWrapperProps` leads to `notifyNestedSubs`) was simply being skipped. The `useSyncExternalStore` is a passive effect though so it gets unsubscribed later and thus we no longer skip this notification.

Interestingly, I went back in time to check out something and in 7.x this `isMounted` wasn't in the code at all. That's because the subscription to the store was not passive. It was also managed from a layout effect:
https://github.com/reduxjs/react-redux/blob/fee81904ba4414d75872da1e84b24e54214eb371/src/components/connectAdvanced.js#L417-L419

I've noticed one additional thing in the current implementation:
https://github.com/reduxjs/react-redux/blob/720f0ba79236cdc3e1115f4ef9a7760a21784b48/src/components/connect.tsx#L161-L163

Perhaps if this could be refactored to get more "inline" with the intended usage the problem would go away. I assume that in such a case React's implementation wouldn't rerender this component at all as it was already unmounted in its internal data structures and `mapStateToProps` wouldn't get called at all. This would change the timing of `mapStateToProps` calls though and perhaps it would have some performance implications as React would get notified about potential changes more often. However, I would also assume that it could mostly bailout from unwanted rerenders thanks to the `/with-selector`. You would know better though as you probably have already tried the less hacky usage of that hook and it didn't play out well for you 😅

markerikson:
Yeah, the reason for all that "non-standard" usage of `uSES` is that `connect` cares about more than just "the store state" or "the result of `mapState`" having changed.  It has to combine together the results of `mapState` + `mapDispatch`, _and_ optionally use the results of `mergeProps` if that's been supplied, and then use whatever _that_ combined object is to decide if the child component should re-render.

I couldn't come up with a good way to break apart that logic, so I ended up with an implementation where `connect` is still doing the bulk of the comparison work.

src/components/connect.tsx
erikras: Such is the live of a developer. What's comforting is that Future You thinks Present You is a moron, too. 🤣 

Of course this is a good thing. We want Future Us to be smarter than Present Us. 📈

markerikson:
After further investigation, I've concluded this is actually a bug in the `03-25` experimental React build Replay is using - it works as expected in a current (`08-17`) React experimental build.

See #1940 for discussion.

Thanks for the attempted fix!

Issue #1944, opened 2022-08-22
ConnectedProps: dispatch prop expected even if there is no mapDispatchToProps 


TheHaff:
### What version of React, ReactDOM/React Native, Redux, and React Redux are you using?

- React: 
- ReactDOM/React Native: 
- Redux:
- React Redux:

### What is the current behavior?

i use ConnectedProps to infer my component props instead of manual typings, but i noticed when reusing the result in a different component where i manually piped down the same props that it expects dispatch.

### What is the expected behavior?

Should'nt need dispatch in these cases

### Did this work in previous versions of React Redux?

- [ ] Yes

markerikson:
What do you mean by "expects `dispatch`"?  What error message are you seeing?

Can you put together a sandbox or repo that shows the actual problem?

The PR you filed doesn't look correct.  If you don't provide a `mapDispatch` argument, React-Redux _will_ pass in `props.dispatch` automatically.

PR #1945, opened 2022-08-22
Issue 1944 dispatch is required even without mapDispatchToProps 

TheHaff:
Issue: [1944](https://github.com/reduxjs/react-redux/issues/1944) Removed dispatch props from the connect(mapStateToProps) case.


markerikson:
Appreciate the PR, but this definitely does _not_ look correct.

When you don't provide a `mapDispatch` argument, React-Redux _will_ automatically pass in `props.dispatch` to the component.

markerikson:
They _are_ going to get `props.dispatch` from `connect` - that part is correct.  So, if you want to use those components in some other non-connected situation, then you you should write types that actually capture the real props that they're getting.  If necessary, do `type PropsWithDispatch = Omit<PropsFromRedux, 'dispatch'>` or something.

TheHaff:
Fair enough, thanks for having a look :)

PR #1947, opened 2022-08-25
selectorFactory: pass ownProps to areStatesEqual

jspurlin:
This PR adds the ability for consumers to have access to `ownProps` inside of `areStatesEqual`.

Currently there is no way to 

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -130,7 +130,7 @@ const mapStateToProps = (...args) => {
 
 When `options.pure` is true, `connect` performs several equality checks that are used to avoid unnecessary calls to `mapStateToProps`, `mapDispatchToProps`, `mergeProps`, and ultimately to `render`. These include `areStatesEqual`, `areOwnPropsEqual`, `areStatePropsEqual`, and `areMergedPropsEqual`. While the defaults are probably appropriate 99% of the time, you may wish to override them with custom implementations for performance or other reasons. Here are several examples:
 
-* You may wish to override `areStatesEqual` if your `mapStateToProps` function is computationally expensive and is also only concerned with a small slice of your state. For example: `areStatesEqual: (next, prev) => prev.entities.todos === next.entities.todos`; this would effectively ignore state changes for everything but that slice of state.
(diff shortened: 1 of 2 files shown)


jspurlin:
Note, I'm also willing to make the corresponding changes in the v7 and master branches

phryneas:
Our general stance on `connect` is that it is a legacy api that's essentially mostly there to support class components, which themselves are pretty much a legacy api in the React ecosystem - and that as such we won't be adding any changes or new features to `connect` since we generally want to encourage people to slowly migrate away from it instead of sending the opposite message.. 

Back in 2020, when #781 was last discussed, that wasn't as clear as today yet.

jspurlin:
So the approach before was "We don't know if we want to add the functionality, push it off" to "Hmm, we've gotten multiple requests for this, it seems reasonable to add especially considering the straightforwardness of the change" to "We pushed it off long enough that now it's considered legacy and we don't want to allow the change"? That doesn't sound great.

While I agree that migrating to the hooks approach is the better long-term strategy, making that transition might require a substantial amount of work and enabling the functionality this PR brings helps bridge the gap.

timdorr:
One point of order: This is targetting v5, which was last released almost 3 years ago. I'd rather not touch something that old. This should probably target v7 instead.

I would say that we should add this in. I agree that it may not be best for greenfield code to use `connect`. But if you're working in an existing project structure that is designed around building connected class components (and with a team that is trained as such), you're probably still going to use them for new code you write. So, adding this in won't be a concession in getting folks to use hooks. They aren't switching anytime soon.

jspurlin:
The assumption that 

> If you need this feature now, the code using connect has not been written yet.

may not be correct (and is not correct in my case). We have many connected components and when investigating application performance, it was determined that `mapStateToProps` (and the resulting code that comes along running `mapStateToProps`) is a performance bottleneck. There are straightforward ways to mitigate the need for running `mapStateToProps` (and the rest of the update flow) from `areStatesEqual` if we have access to `ownProps` (like `ownProps.id`, for example)

jspurlin:
Is there any opposition to adding this to v5 for backwards compat reasons?

phryneas:
> Is there any opposition to adding this to v5 for backwards compat reasons?

Please at least update your stack if you want new features. :slightly_frowning_face:

markerikson:
I'm willing to look at this for v7 and v8.

But we're no longer actively maintaining v5 and v6 at this point, and if for some reason your app is still on v5, I would _strongly_ urge you to upgrade to v7 immediately.  The public API is the same (with just a minor change to how the `refs` option works), and it has better perf.

jspurlin:
I've opened https://github.com/reduxjs/react-redux/pull/1951 against master and https://github.com/reduxjs/react-redux/pull/1952 against v7

timdorr:
OK, closing this one out.

PR #1951, opened 2022-08-29
connect: pass ownProps to areStatesEqual

jspurlin:
This PR adds the ability for consumers to have access to ownProps inside of areStatesEqual.

Currently there is no way to

### DIFF
--- a/src/components/connect.tsx
+++ b/src/components/connect.tsx
@@ -701,7 +701,7 @@ function connect<
         notifyNestedSubs,
       ])
 
-      let actualChildProps: uSES
+      let actualChildProps: Record<string, unknown>
 
       try {
         actualChildProps = useSyncExternalStore(
--- a/src/components/connect.tsx
+++ b/src/components/connect.tsx
@@ -696,7 +696,7 @@ function connect<
         notifyNestedSubs,
       ])
 
-      let actualChildProps: unknown
+      let actualChildProps: uSES
 
       try {
         actualChildProps = useSyncExternalStore(
--- a/src/components/connect.tsx
+++ b/src/components/connect.tsx
@@ -231,7 +231,12 @@ export interface ConnectOptions<
 > {
   forwardRef?: boolean
   context?: typeof ReactReduxContext
-  areStatesEqual?: (nextState: State, prevState: State) => boolean
+  areStatesEqual?: (
+    nextState: State,
+    prevState: State,
+    nextOwnProps: TOwnProps,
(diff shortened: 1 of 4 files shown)


jspurlin:
Ping, @markerikson & @timdorr can you take a look at this PR?

timdorr (approved):
This one looks good as well. 

(BTW, if you rebase on master, the test fail with TS@next will go away)

src/components/connect.tsx
```diff
@@ -701,7 +701,7 @@ function connect<
         notifyNestedSubs,
       ])
 
-      let actualChildProps: unknown
+      let actualChildProps: uSES
```
markerikson, line 704: :hand: This code is internal and the type here won't matter much, but this _should_ be unknown.  Note that we're _calling_ `useSyncExternalStore` here and saving the result into `actualChildProps` - it's not a pointer to the hook itself.

jspurlin: I rebased on master and pulled master. Not sure why it's showing as changing in my PR. It looks like that change was made here  https://github.com/reduxjs/react-redux/commit/fb8cfc0d76559d9ed028f26413e8845014ab51ee

markerikson: huh. @timdorr WHAT R U DOING?!?!?!? :)

timdorr: This is to fix typescript@next. You can't spread an unknown (which is done below this for the props on the component), which is no longer allowed. The value can't ever be anything other than a `uSES` because the only other option is for the useSyncExternalState to throw an error which is caught and then re-thrown. It either works and is a `uSES` or bails out of the function entirely.

markerikson: In that case, I think we ought to go with something like `Record<string, unknown>`.

The `uSES` type represents the actual `useSyncExternalStore` function, so it isn't the right type for the `actualChildProps` value.

jspurlin: I agree that `Record<string, unknown>` is a more accurate type. I've made update in this PR

timdorr: Ah, gotcha. That is more correct.

jspurlin:
@timdorr and @markerikson, is this (and #1952) good to be merged?

kysedate:
@timdorr, @markerikson, can this change and #1952 be merged?

markerikson:
Sorry, React-Redux has been pretty low on the time/mental priority list lately, but taking a look at things tonight.

PR #1952, opened 2022-08-29
connect: pass ownProps to areStatesEqual

jspurlin:
This PR adds the ability for consumers to have access to ownProps inside of areStatesEqual.

Currently there is no way to


jspurlin:
Ping, @markerikson & @timdorr can you take a look at this PR?

timdorr (approved):
I'm good with this as-is. Any concerns from other folks?

Issue #1955, opened 2022-09-20
Possible regression with ConnectedProps and React.ComponentProps<typeof ...> in 8.x


cascornelissen:
### What version of React, ReactDOM/React Native, Redux, and React Redux are you using?

- React: 16.14.0
- ReactDOM: 16.14.0
- Redux: 4.2.0
- React Redux: 8.0.2
- TypeScript: 4.8.3

### What is the current behavior?

The following simplified snippet throws a `TS2312` error after upgrading to react-redux v8.

```tsx
import React from 'react';
import { connect, ConnectedProps } from 'react-redux';

interface Props extends React.ComponentProps<'div'>, ConnectedProps<typeof connector> {}

const ExampleComponent: React.FunctionComponent<Props> = ({ data, ...props }) => {
    return <div {...props} />
}

const mapStateToProps = (state: { data: unknown[] }) => ({
    data: state.data
});

const connector = connect(mapStateToProps);
const ConnectedExampleComponent = connector(ExampleComponent);

// This next line is where the error happens
interface Props2 extends React.ComponentProps<typeof ConnectedExampleComponent> {}
```

```
TS2312: An interface can only extend an object type or intersection of object types with statically known members.
```

There seem to be no apparent changes to the definition of `ConnectedProps` between `@types/react-redux` and `react-redux` so I'm not sure what could be causing this.

The changelog does mention changes to the types for `connect` where it no longer has the `DefaultRootState` included. But even changing the definition of `connector` in the example above to include a generic with `any`, `{}` or the actual global state interface does not resolve the issue.

### What is the expected behavior?

It's still possible to use `React.ComponentProps<typeof ...>` like when using `react-redux@7.2.8` with `@types/react-redux@7.1.24`

### Which browser and OS are affected by this issue?

N/A

### Did this work in previous versions of React Redux?

- [X] Yes

markerikson:
Huh. I certainly haven't seen _that_ message before.

Uh.... summoning @phryneas and @Andarist HALP WAT THIS MEAN

(fwiw this is going to be pretty low priority for us to look at, given that we generally recommend using the hooks API instead of `connect` today, and we're focusing on RTK feature work atm.)

phryneas:
I could for a second reproduce this with the React 16 types, but then it was gone and just working. I guess this is some edge case with old React types?

Generally: you are using React 16 - is there any practical reason why you are updating React-Redux first? Generally I would probably update the rest of the stack first/at the same time.

[...]

Issue #1187, opened 2019-02-14
Conflict with passing a "context" prop


jmakGH:
Upon upgrading React and React Redux to the latest versions, an app I am working on started crashing on render.

After some debugging I traced the issue to a connected component being passed a "context" prop that was being used to control some of the app's features and completely unrelated to React's Context API.

I couldn't find any notes about that being a protected prop anywhere in the docs or online and thought it might be a good idea to post this here in case anyone else happened to stumble across this issue. 

The biggest takeaway was that the only error message came from React, which made debugging the issue kind of a hassle -- especially because everything was working prior to upgrading to the latest version (which makes sense since the version of React we were using was before their Context update).

Perhaps a note under the Troubleshooting section of the website could document this issue as a potential pitfall.

[Issue is recreated at this CodeSandbox. The app crashes whenever a context prop is passed to ConnectedChild.](https://codesandbox.io/s/ww86vw77qw)

**Which versions of React, ReactDOM/React Native, Redux, and React Redux are you using?  Which browser and OS are affected by this issue? Did this work in previous versions of React Redux?**
React 16.8.1
ReactDOM 16.8.1
React Redux 6.0.0

This issue wasn't happening with React v15 and React Redux v4

markerikson:
It is mentioned in the docs a couple times, but not particularly highlighted:

https://react-redux.js.org/api/connect#context-object

https://react-redux.js.org/using-react-redux/accessing-store#providing-custom-context

Up through 5.x, `store` was effectively a "reserved" prop, allowing users to pass a specific store instance directly to a connected component.

In 6.0, `store` is no longer meaningful as a prop, but we throw errors to let users know they probably made a mistake when upgrading and should change their code (because it won't do what they expected it to do).  We also added `context` as a "reserved prop", because we allow users to pass a custom context instance to a connected component.

Given that we're considering [having our next version be a major version change anyway](https://github.com/reduxjs/react-redux/issues/1177), perhaps we could consider renaming that prop to be something like `customContext` instead of just `context`.

jmakGH:
My mistake, I had overlooked that in the docs. I'm perhaps at fault for using your comment on #1104 as a reference guide for the migration without completely reading the updated docs.

I think you guys should be fine leaving it as `context` seeing as how I had a difficult time finding any discussion from others facing this issue. I'd assume most people now generally reserve `context` for React in general to avoid conflicts in semantics/code. We just had just been using it as a prop for a long time before they updated their Context API.

[...]

PR #1956, opened 2022-09-21
Fixed type issue with `ComponentProps` from older `@types/react`

Andarist:
fixes https://github.com/reduxjs/react-redux/issues/1955

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -1,4 +1,9 @@
-import { ClassAttributes, ComponentClass, ComponentType } from 'react'
+import {
+  ClassAttributes,
+  ComponentClass,
+  ComponentType,
+  FunctionComponent,
+} from 'react'
 
 import { Action, AnyAction, Dispatch } from 'redux'
 
@@ -82,7 +87,7 @@ export type GetLibraryManagedProps<C> = JSX.LibraryManagedAttributes<
 export type ConnectedComponent<
   C extends ComponentType<any>,
   P
-> = ComponentType<P> &
+> = FunctionComponent<P> &
   NonReactStatics<C> & {
     WrappedComponent: C
   }


src/types.ts
```diff
@@ -82,7 +87,7 @@ export type GetLibraryManagedProps<C> = JSX.LibraryManagedAttributes<
 export type ConnectedComponent<
   C extends ComponentType<any>,
   P
-> = ComponentType<P> &
+> = FunctionComponent<P> &
```
Andarist, line 90: `ComponentType` is a union - your `connect` returns a function component so there was no need to use the `ComponentType` here. Somehow because of that union, the inferred props through `ComponentProps` were also computed as a union, and that caused a problem. This only occurred with older `@types/react` **because** their `FunctionComponent` included `PropsWithChildren`.

I've wanted to write a type test for this - but to pull it off you would have to extend your test suite with a `@types/react` matrix.

markerikson:
Oh cool! I'll try to get this in tonight.

timdorr:
I'm going to merge this in now, as it looks good. We can do a release later today.

cascornelissen:
@timdorr, not trying to push anyone but I was wondering whether this release was forgotten yesterday? 😬

markerikson:
There's one other TS issue I'd like to fix before putting out a patch release (related to recent discussion in #1187 )

emab:
Seeing some TS errors after updating to this version. It seems to occur when importing an untyped component that spreads `state` in its `mapStateToProps` function and then trying to pass any props to that component.

I've created a sandbox here highlighting the issue:

https://codesandbox.io/s/flamboyant-dream-9gm53p?file=/src/App.tsx

Note [in this file](https://codesandbox.io/s/flamboyant-dream-9gm53p?file=/src/ConnectedComponentWithPropsUntyped.js) the two `mapStateToProps` functions - the one that spreads state causes the type issue in `App.tsx` whilst the other doesn't.

Andarist:
For what it's worth - I don't see the error going away after reverting the change from this PR.

I've also checked the inferred types for both of your untyped~ `mapStateToProps`.

```ts
// value declaration:
// const mapStateToProps = (state) => { return { ...state }; };
// its inferred signature:
// const mapStateToProps: (state: any) => any

React.FunctionComponent<{
    context?: React.Context<ReactReduxContextValue<any, AnyAction>> | undefined;
    store?: Store<any, AnyAction> | undefined;
} | {
    store?: Store<any, AnyAction> | undefined;
}> & hoistNonReactStatics.NonReactStatics<(props: any) => JSX.Element, {}> & {
    WrappedComponent: (props: any) => JSX.Element;
}

// value declaration:
// const mapStateToProps = (state) => { return { stateValue: state.stateValue }; };
// its inferred signature:
// const mapStateToProps: (state: any) => { stateValue: any; }

React.FunctionComponent<{
    [x: string]: any;
    [x: number]: any;
    [x: symbol]: any;
    context?: React.Context<ReactReduxContextValue<any, AnyAction>> | undefined;
    store?: Store<any, AnyAction> | undefined;
} | {
    [x: string]: any;
    [x: number]: any;
    [x: symbol]: any;
    store?: Store<any, AnyAction> | undefined;
}> & hoistNonReactStatics.NonReactStatics<(props: any) => JSX.Element, {}> & {
    WrappedComponent: (props: any) => JSX.Element;
}
```

PR #1958, opened 2022-09-23
Omit built-in context prop if user component props include context

### LINKED ISSUE #1957: v8 TS types don't allow passing a component prop named `context`
Original discussion in #1187 starting at https://github.com/reduxjs/react-redux/issues/1187#issuecomment-1252746667 , posted by @strongui  .

Pasting some relevant comment extracts:

> we have a huge app, we haven't had time to update until recently. 
> 
> ```
> const connector = connect(mapStateToProps, mapDispatchToProps);
> 
> const connectorCore = connect(null, mapDispatchToPropsCoreMethods);
> 
> const BaseComp = (props) => {
>   return <div></div>
> }
> 
> // This COMP will get all it's data from the redux store
> export const FullComponent = connector(BaseComp);
> // This COMP will demand data from the parent
> export const NonDataComponent = connectorCore(BaseComp);
> ```
> 
> Basically this. Same component, but one gets data form redux, the other will get data from parent.

Sandbox: https://codesandbox.io/s/tender-mestorf-x272kj?file=/src/index.tsx 

> now you see the TS error:
> 
> `Type '"LIST"' is not assignable to type '("LIST" | "CARD") & Context<ReactReduxContextValue<any, AnyAction>>'.ts(2322)`

From me:

> Hmm. Interesting.  If I fix the use of `connector(Button)`, and then swap back to React-Redux v7, the wrapped `Button` component says that `context` is just that enum, not a `ReactReduxContext` instance.
> 
> I wonder what about our types changed there?  It most likely had something to do with the TS conversion process.
> 
> The internal logic for "what do we _do_ with `props.context` internally once we see it" hasn't changed, and if I'm reading this right, it ought to actually pass that through properly if it's _not_ a context instance:
> 
> ```ts
>       const ContextToUse: ReactReduxContextInstance = useMemo(() => {
>         // Users may optionally pass in a custom context instance to use instead of our ReactReduxContext.
>         // Memoize the check that determines which context instance we should use.
>         return propsContext &&
>           propsContext.Consumer &&
>           // @ts-ignore
>           isContextConsumer(<propsContext.Consumer />)
>           ? propsContext
>           : Context
>       }, [propsContext, Context])
> 
>       // Retrieve the store and ancestor subscription via context, if available
>       const contextValue = useContext(ContextToUse)
> ```
> 
> It's probably this part:
> 
> ```ts
> export interface ConnectProps {
>   /** A custom Context instance that the component can use to access the store from an alternate Provider using that same Context instance */
>   context?: ReactReduxContextInstance
>   /** A Redux store instance to be used for subscriptions instead of the store from a Provider */
>   store?: Store
> }
> 
> interface InternalConnectProps extends ConnectProps {
>   reactReduxForwardedRef?: React.ForwardedRef<unknown>
> }
> 
>     function ConnectFunction<TOwnProps>(
>       props: InternalConnectProps & TOwnProps
>     )  {
>        // omit
>     }
> ```
> 
> Given that, I think this would _run_, but it won't compile, which is an issue.

markerikson:
This PR:

- Reworks the type inference for `connect` so that if the user's component expects `props.context` already, we use that type instead of the possible alternate `context: ReactReduxContext` type
- Uses a `Mapped` type to simplify the hover display types for connected components.

Connected components allow users to pass in a `ReactReduxContext` instance as a prop named `context`. We also do internal checks to see if that really _is_ a context instance, in case the user passed a real value as `props.context`.

However, the TS types did not reflect this - `props.context` was always a `ReactReduxContext` type, and if users did have a component that accepted a prop named `context`, this would cause a type clash.

In this case, the types didn't reflect the reality of runtime behavior.

By extracting "the props of the component" and omitting the built-in `context` field type if appropriate, this now works.

Fixes #1957 .

Additionally: due to the ridiculous types-level shenanigans inside of `connect`, the hover display of connected components is incredibly obtuse and hard to read.  

Over the last year I've experimented with several different util types that "expand/flatten" a TS typedef to make the actual fields more readable (and by "experimented" I mean "I stole them from other repos and pasted them in to see what would happen").  One of the simpler ones is a `Mapped` type:

```ts
type Identity<T> = T
export type Mapped<T> = Identity<{ [k in keyof T]: T[k] }>
```

(I don't pretend to understand all the magic there, but it works!)

By inserting that at the right spot in the connected inference sequence, we get a much better hover display for connected components.

Some examples from our type tests:

```ts
// Before:
const ConnectedButton: ConnectedComponent<(props: IButtonProps) => JSX.Element, Omit<IButtonOwnProps & {
    name: string;
} & DispatchProp<AnyAction>, "dispatch" | "name"> & {
    ...;
}>

// After
const ConnectedButton: ConnectedComponent<(props: IButtonProps) => JSX.Element, {
    label: string;
    context: 'LIST' | 'CARD';
    store?: Store<any, AnyAction> | undefined;
}>
```

```ts
// Before:
const Test: ConnectedComponent<typeof TestComponent, Omit<React.ClassAttributes<TestComponent> & OwnProps & StateProps & DispatchProps, "bar" | "onClick"> & {
    ...;
}>

// After
const Test: ConnectedComponent<typeof TestComponent, {
    foo: string;
    ref?: React.LegacyRef<TestComponent> | undefined;
    key?: React.Key | null | undefined;
    context?: React.Context<ReactReduxContextValue<any, AnyAction>> | undefined;
    store?: Store<...> | undefined;
}>
```

Most users aren't going to care about the `ref/key/context/store` part... but it _is_ much easier to read _and_ more accurate.  So we're throwing that in here as a bonus.

### DIFF
--- a/src/components/connect.tsx
+++ b/src/components/connect.tsx
@@ -1,17 +1,17 @@
 /* eslint-disable valid-jsdoc, @typescript-eslint/no-unused-vars */
 import hoistStatics from 'hoist-non-react-statics'
-import React, { useContext, useMemo, useRef } from 'react'
+import React, { ComponentType, useContext, useMemo, useRef } from 'react'
 import { isValidElementType, isContextConsumer } from 'react-is'
 
 import type { Store } from 'redux'
 
 import type {
-  AdvancedComponentDecorator,
   ConnectedComponent,
   InferableComponentEnhancer,
   InferableComponentEnhancerWithProps,
   ResolveThunks,
   DispatchProp,
+  ConnectPropsMaybeWithoutContext,
 } from '../types'
 
 import defaultSelectorFactory, {
@@ -465,18 +465,18 @@ function connect<
 
   const Context = context
 
-  type WrappedComponentProps = TOwnProps & ConnectProps
-
   const initMapStateToProps = mapStateToPropsFactory(mapStateToProps)
   const initMapDispatchToProps = mapDispatchToPropsFactory(mapDispatchToProps)
   const initMergeProps = mergePropsFactory(mergeProps)
 
   const shouldHandleStateChanges = Boolean(mapStateToProps)
 
-  const wrapWithConnect: AdvancedComponentDecorator<
-    TOwnProps,
-    WrappedComponentProps
-  > = (WrappedComponent) => {
+  const wrapWithConnect = <TProps,>(
+    WrappedComponent: ComponentType<TProps>
+  ) => {
+    type WrappedComponentProps = TProps &
+      ConnectPropsMaybeWithoutContext<TProps>
+
     if (
       process.env.NODE_ENV !== 'production' &&
       !isValidElementType(WrappedComponent)
--- a/src/types.ts
+++ b/src/types.ts
@@ -25,10 +25,6 @@ export interface DispatchProp<A extends Action = AnyAction> {
   dispatch: Dispatch<A>
 }
 
-export type AdvancedComponentDecorator<TProps, TOwnProps> = (
-  component: ComponentType<TProps>
-) => ComponentType<TOwnProps>
-
 /**
  * A property P will be present if:
  * - it is present in DecorationTargetProps
@@ -92,10 +88,17 @@ export type ConnectedComponent<
     WrappedComponent: C
   }
 
+export type ConnectPropsMaybeWithoutContext<TActualOwnProps> =
+  TActualOwnProps extends { context: any }
+    ? Omit<ConnectProps, 'context'>
+    : ConnectProps
+
 // Injects props and removes them from the prop requirements.
 // Will not pass through the injected props if they are passed in during
 // render. Also adds new prop requirements from TNeedsProps.
-// Uses distributive omit to preserve discriminated unions part of original prop type
+// Uses distributive omit to preserve discriminated unions part of original prop type.
+// Note> Most of the time TNeedsProps is empty, because the overloads in `Connect`
+// just pass in `{}`.  The real props we need come from the component.
 export type InferableComponentEnhancerWithProps<TInjectedProps, TNeedsProps> = <
   C extends ComponentType<Matching<TInjectedProps, GetProps<C>>>
 >(
@@ -107,7 +110,7 @@ export type InferableComponentEnhancerWithProps<TInjectedProps, TNeedsProps> = <
     keyof Shared<TInjectedProps, GetLibraryManagedProps<C>>
   > &
     TNeedsProps &
-    ConnectProps
+    ConnectPropsMaybeWithoutContext<TNeedsProps & GetProps<C>>
 >
 
 // Injects props and removes them from the prop requirements.
--- a/src/types.ts
+++ b/src/types.ts
@@ -93,6 +93,9 @@ export type ConnectPropsMaybeWithoutContext<TActualOwnProps> =
     ? Omit<ConnectProps, 'context'>
     : ConnectProps
 
+type Identity<T> = T
+export type Mapped<T> = Identity<{ [k in keyof T]: T[k] }>
+
 // Injects props and removes them from the prop requirements.
 // Will not pass through the injected props if they are passed in during
 // render. Also adds new prop requirements from TNeedsProps.
@@ -105,12 +108,14 @@ export type InferableComponentEnhancerWithProps<TInjectedProps, TNeedsProps> = <
   component: C
 ) => ConnectedComponent<
   C,
-  DistributiveOmit<
-    GetLibraryManagedProps<C>,
-    keyof Shared<TInjectedProps, GetLibraryManagedProps<C>>
-  > &
-    TNeedsProps &
-    ConnectPropsMaybeWithoutContext<TNeedsProps & GetProps<C>>
+  Mapped<
+    DistributiveOmit<
+      GetLibraryManagedProps<C>,
+      keyof Shared<TInjectedProps, GetLibraryManagedProps<C>>
+    > &
+      TNeedsProps &
+      ConnectPropsMaybeWithoutContext<TNeedsProps & GetProps<C>>
+  >
 >
 
 // Injects props and removes them from the prop requirements.
--- a/test/typetests/connect-options-and-issues.tsx
+++ b/test/typetests/connect-options-and-issues.tsx
@@ -32,6 +32,7 @@ import {
   createStoreHook,
   TypedUseSelectorHook,
 } from '../../src/index'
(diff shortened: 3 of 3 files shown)


markerikson:
HALP the TypeScript just blew up on the floor:

```ts
Version 4.2.4
Error: src/components/connect.tsx(789,5): error TS2322: Type 'FunctionComponent<WrappedComponentProps> & NonReactStatics<ComponentType<TProps>, {}> & { ...; }' is not assignable to type 'ComponentType<TActualOwnProps>'.
  Type 'FunctionComponent<WrappedComponentProps> & NonReactStatics<ComponentType<TProps>, {}> & { ...; }' is not assignable to type 'FunctionComponent<TActualOwnProps>'.
    Types of property 'propTypes' are incompatible.
      Type 'WeakValidationMap<WrappedComponentProps> | undefined' is not assignable to type 'WeakValidationMap<TActualOwnProps> | undefined'.
        Type 'WeakValidationMap<WrappedComponentProps>' is not assignable to type 'WeakValidationMap<TActualOwnProps>'.
          Type '(null extends WrappedComponentProps[K] ? Validator<WrappedComponentProps[K] | (WrappedComponentProps[K] & null) | undefined> : undefined extends WrappedComponentProps[K] ? Validator<...> : Validator<...>) | undefined' is not assignable to type '(null extends TActualOwnProps[K] ? Validator<TActualOwnProps[K] | (TActualOwnProps[K] & null) | undefined> : undefined extends TActualOwnProps[K] ? Validator<...> : Validator<...>) | undefined'.
            Type 'null extends WrappedComponentProps[K] ? Validator<WrappedComponentProps[K] | (WrappedComponentProps[K] & null) | undefined> : undefined extends WrappedComponentProps[K] ? Validator<...> : Validator<...>' is not assignable to type '(null extends TActualOwnProps[K] ? Validator<TActualOwnProps[K] | (TActualOwnProps[K] & null) | undefined> : undefined extends TActualOwnProps[K] ? Validator<...> : Validator<...>) | undefined'.
              Type 'Validator<WrappedComponentProps[K] | (WrappedComponentProps[K] & null) | undefined> | (undefined extends WrappedComponentProps[K] ? Validator<...> : Validator<...>)' is not assignable to type '(null extends TActualOwnProps[K] ? Validator<TActualOwnProps[K] | (TActualOwnProps[K] & null) | undefined> : undefined extends TActualOwnProps[K] ? Validator<...> : Validator<...>) | undefined'.
                Type 'Validator<WrappedComponentProps[K] | (WrappedComponentProps[K] & null) | undefined>' is not assignable to type 'null extends TActualOwnProps[K] ? Validator<TActualOwnProps[K] | (TActualOwnProps[K] & null) | undefined> : undefined extends TActualOwnProps[K] ? Validator<...> : Validator<...>'.
                  Type 'null extends WrappedComponentProps[K] ? Validator<WrappedComponentProps[K] | (WrappedComponentProps[K] & null) | undefined> : undefined extends WrappedComponentProps[K] ? Validator<...> : Validator<...>' is not assignable to type 'null extends TActualOwnProps[K] ? Validator<TActualOwnProps[K] | (TActualOwnProps[K] & null) | undefined> : undefined extends TActualOwnProps[K] ? Validator<...> : Validator<...>'.
                    Type 'Validator<WrappedComponentProps[K] | (WrappedComponentProps[K] & null) | undefined> | (undefined extends WrappedComponentProps[K] ? Validator<...> : Validator<...>)' is not assignable to type 'null extends TActualOwnProps[K] ? Validator<TActualOwnProps[K] | (TActualOwnProps[K] & null) | undefined> : undefined extends TActualOwnProps[K] ? Validator<...> : Validator<...>'.
                      Type 'Validator<WrappedComponentProps[K] | (WrappedComponentProps[K] & null) | undefined>' is not assignable to type 'null extends TActualOwnProps[K] ? Validator<TActualOwnProps[K] | (TActualOwnProps[K] & null) | undefined> : undefined extends TActualOwnProps[K] ? Validator<...> : Validator<...>'.
```
                      
oops

markerikson:
Board is green, let's get this in.

markerikson:
*sigh*

I was supposed to merge this an hour ago, _before_ publishing 8.0.3

src/components/connect.tsx
```diff
@@ -465,18 +465,18 @@ function connect<
...
   const initMapDispatchToProps = mapDispatchToPropsFactory(mapDispatchToProps)
   const initMergeProps = mergePropsFactory(mergeProps)
 
   const shouldHandleStateChanges = Boolean(mapStateToProps)
 
-  const wrapWithConnect: AdvancedComponentDecorator<
-    TOwnProps,
-    WrappedComponentProps
-  > = (WrappedComponent) => {
+  const wrapWithConnect = <TProps,>(
```
phryneas, line 474: is the comma a typo?

strongui:
@markerikson First of all thank you for doing all of this, I am honestly blown away how responsive you guys have been given the size of this library. 

I have updated my sandbox to use this version of the PR: https://codesandbox.io/s/tender-mestorf-x272kj?file=/src/client/Button.tsx

The error is gone! As long as the component has it's own context: **required** prop, there will be no error.

If the prop is optional IE: `context?: CARD | LIST`, we have the same TS error as before. Is that intentional you think or simply a limitation of TS in this case, since like you said some of these typing are pretty intense! In any case, thank you again, I can make this work!

Also 

`type Identity<T> = T
export type Mapped<T> = Identity<{ [k in keyof T]: T[k] }>`

This is pure magical! I plan on using this in our app, this is a complete game changer for hover display!

markerikson:
@strongui Nice, glad that works! 

I'd say that the change is semi-intentional / semi-limited.  If you look at the implementation, it boils down to a check of `TOwnProps extends { context : any } ?`.  In other words, "does the type of your component's props include a field named context in it?".  

I'm honestly not sure if you can detect _optional_ fields that way. It's possible you can, this was just the first solution I could think of.

Note that this is now out publicly in 8.0.4:

https://github.com/reduxjs/react-redux/releases/tag/v8.0.4

src/components/connect.tsx
markerikson: No, it was actually necessary!

When I removed it, TS thought that `<TProps>` was the start of a JSX element, because this _is_ a `.tsx` file.  I had to leave it in to convince TS "nope, this is a generic type declaration" (presumably similar to how in Python `(x)` is just parentheses, and `(x,)` is a tuple).

phryneas:
`TOwnProps extends { context?: any } ?` should do the trick if I'm not mistaken :)

markerikson:
Hmm. Yeah, that does actually appear to work.

PR #1960, opened 2022-10-06
`Provider`: pass state (`S`) generic through to `ProviderProps`

OliverJAsh:
This is necessary to catch mistakes such as this, where the type of `serverState` does not match the type of the state in the provided `store`. (I just shipped this bug to unsplash.com!)

```tsx
import { Provider } from 'react-redux';
import { Store } from 'redux';

declare const store: Store<{ foo: string }>;
<Provider store={store} serverState={"oops"}>foo</Provider>
```

### DIFF
--- a/src/components/Provider.tsx
+++ b/src/components/Provider.tsx
@@ -4,7 +4,7 @@ import { createSubscription } from '../utils/Subscription'
 import { useIsomorphicLayoutEffect } from '../utils/useIsomorphicLayoutEffect'
 import { Action, AnyAction, Store } from 'redux'
 
-export interface ProviderProps<A extends Action = AnyAction, S = any> {
+export interface ProviderProps<A extends Action = AnyAction, S = unknown> {
   /**
    * The single Redux store in your application.
    */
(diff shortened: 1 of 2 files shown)


src/components/Provider.tsx
```diff
@@ -24,12 +24,12 @@ export interface ProviderProps<A extends Action = AnyAction, S = any> {
   children: ReactNode
 }
 
-function Provider<A extends Action = AnyAction>({
+function Provider<A extends Action = AnyAction, S = any>({
```
markerikson, line 27: :question: Does it make sense for this to be `unknown` instead?

OliverJAsh: Definitely, but `ProviderProps` uses `any` as the default so I had to do the same here. If you wish I can change both of those defaults to `unknown`? That's a larger change and I wasn't sure if you would want that.

OliverJAsh: 9e09869307f829273d8a38a452d814a546b692b0

Issue #1644, opened 2020-10-17
InferableComponentEnhancerWithProps does not correctly infer class-based (React.Component flavor) component properties


yurigenin:
**What is the current behavior?**

When using a component class extending **React.Component** to connect to Redux store with **connect()** HOC the returned component is of an **InferableComponentEnhancerWithProps** type. But this type does not correctly derive the properties of the component because internally it relies on the component to extend **React.ComponentType**. But **React.Component** does not extend it, thus, the property inferring fails.

The definition of the **InferableComponentEnhancerWithProps** is here:

```
export type InferableComponentEnhancerWithProps<TInjectedProps, TNeedsProps> =
    <C extends ComponentType<Matching<TInjectedProps, GetProps<C>>>>(
        component: C
    ) => ConnectedComponent<C, Omit<GetProps<C>, keyof Shared<TInjectedProps, GetProps<C>>> & TNeedsProps>;
```
It uses **GetProps** type to match injected properties with **propTypes** properties:

```
export type GetProps<C> = C extends ComponentType<infer P> ? P : never;
```

It will always return **never** since **React.ComponentType** is defined as:

```
type ComponentType<P = {}> = ComponentClass<P> | FunctionComponent<P>;
```

But any component that is defined as:

```
class MyComponent extends React.Component<MyProps> {}
```

does not extend from either **ComponentClass** or **FunctionComponent**.

It results in all injected properties not getting omitted as intended because **GetProps** type is **never**:

```
ConnectedComponent<C, Omit<GetProps<C>, keyof Shared<TInjectedProps, GetProps<C>>> & TNeedsProps>;
```

**What is the expected behavior?**

All injected properties that match **propTypes** should be omitted.

Which versions of React, ReactDOM/React Native, Redux, and React Redux are you using?  Which browser and OS are affected by this issue? Did this work in previous versions of React Redux?

7.x

markerikson:
We don't maintain the React-Redux TS typings here - they're over in DefinitelyTyped.

markerikson:
Does that still happen with the latest version of `@types/react`?  We're not generally targeting React 16.x at this point, and `@types/react` _mostly_ stays in sync with the current React version.

noahnu:
It works with @types/redux and react-redux v7. Haven't tried using @types/redux with v8 since v8 ships with types. I'll give this a try this week, thanks for the quick reply.

[...]

PR #1991, opened 2023-03-22
fix(types): infer props from class components in React 16

noahnu:
Addresses #1644 which adds support for inferring props from class components (specifically PureComponents) in React v16. In React v16, not all components extend "ComponentType".

On the other hand, JSXElementConstructor handles class components and function components. For getting props, fallback to React's built-in props utility. I initially tried replacing GetProps with React.ComponentProps, but ran into some issues with refs. This is likely a "safer" change that won't be breaking in any way.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -1,8 +1,10 @@
 import {
-  ClassAttributes,
-  ComponentClass,
+  JSXElementConstructor,
   ComponentType,
   FunctionComponent,
+  ComponentClass,
+  ClassAttributes,
+  ComponentPropsWithRef,
 } from 'react'
 
 import { Action, AnyAction, Dispatch } from 'redux'
@@ -71,6 +73,8 @@ export type GetProps<C> = C extends ComponentType<infer P>
   ? C extends ComponentClass<P>
     ? ClassAttributes<InstanceType<C>> & P
     : P
+  : C extends keyof JSX.IntrinsicElements | JSXElementConstructor<any>
+  ? ComponentPropsWithRef<C>
   : never
 
 // Applies LibraryManagedAttributes (proper handling of defaultProps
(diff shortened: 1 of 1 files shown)


src/types.ts
```diff
@@ -105,7 +109,7 @@ export type Mapped<T> = Identity<{ [k in keyof T]: T[k] }>
 // Note> Most of the time TNeedsProps is empty, because the overloads in `Connect`
 // just pass in `{}`.  The real props we need come from the component.
 export type InferableComponentEnhancerWithProps<TInjectedProps, TNeedsProps> = <
-  C extends ComponentType<Matching<TInjectedProps, GetProps<C>>>
+  C extends JSXElementConstructor<Matching<TInjectedProps, GetProps<C>>>
```
timdorr, line 112: I'm really not sure I understand why this is OK to swap out for `ComponentType`. Things like [`React.memo` don't use this type](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/v16/index.d.ts#L860), even in the v16 types. It also seems like a fundamentally different type, describing just the construction of an element, not a complete description of a component for other use cases. For instance, things like `displayName` will be missing from this type. I could be misinterpreting, but I don't think this is the correct type to use here.

noahnu: `InferableComponentEnhancerWithProps` is used indirectly to infer props, so we don't need the additional static properties. At least not for the usage with `connect`.

If you look at [React.ComponentProps](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/d6c3b65cd36ea33204894086cd791ea3adb74f4b/types/react/v16/index.d.ts#L829), you'll see it assumes the type extends either an intrinsic element (which never makes sense to connect anyway) or a JSXElementConstructor.

The issue with [ComponentType](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/d6c3b65cd36ea33204894086cd791ea3adb74f4b/types/react/v16/index.d.ts#L78) is that it is a union of the class type [`ComponentClass`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/d6c3b65cd36ea33204894086cd791ea3adb74f4b/types/react/v16/index.d.ts#L583) which [`Component`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/d6c3b65cd36ea33204894086cd791ea3adb74f4b/types/react/v16/index.d.ts#L431) (or PureComponent) does not extend.

Ultimately the problem I was encountering is that for a simple PureComponent, `GetProps` was always returning `never`, because the PureComponent class does not extend ComponentType.

timdorr: Could you give a code example that shows how the current types break for v16 and would work for v17+? That may be more helpful for me to understand.

noahnu: Yes, I can add some test coverage to this PR as well. Will do first thing Monday.

PR #1994, opened 2023-04-08
rollup-plugin-terser pkg deprecated. Switched to @rollup/plugin-terser

GabrielMurry:
**Issue:**
The 'rollup-plugin-terser' package has been deprecated and no longer maintained.

**Discovered:**
Currently implementing rollup in my own package and looked at react-redux's rollup.config as an example. Noticed the deprecated package.

**Solved:**
Forked repo, uninstalled 'rollup-plugin-terser', then installed '@rollup/plugin-terser', which is the new and maintained rollup terser package. Imported 'terser' from '@rollup/plugin-terser' in rollup.config.js file. yarn.lock automatically updated with new package.

Hope this helps!

### DIFF
--- a/package.json
+++ b/package.json
@@ -88,6 +88,7 @@
     "@rollup/plugin-commonjs": "^15.1.0",
     "@rollup/plugin-node-resolve": "^9.0.0",
     "@rollup/plugin-replace": "^2.3.3",
+    "@rollup/plugin-terser": "^0.4.0",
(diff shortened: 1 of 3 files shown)


timdorr:
Thanks, but @markerikson is working on tooling changes for all our repos that drops Rollup as a tool for tsup (which uses esbuild under the hood). So, this will be overwritten later on anyways.

PR #1999, opened 2023-05-14
Added generic for passing type when using createSelectorHook context

lamo2k123:
Added the ability to specify the type for the store when creating a hook when using the context

### DIFF
--- a/src/hooks/useSelector.ts
+++ b/src/hooks/useSelector.ts
@@ -1,4 +1,5 @@
-import { useContext, useDebugValue } from 'react'
+import { useContext, useDebugValue, Context } from 'react'
+import { AnyAction } from 'redux'
 
 import { useReduxContext as useDefaultReduxContext } from './useReduxContext'
 import { ReactReduxContext } from '../components/Context'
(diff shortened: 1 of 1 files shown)


lamo2k123:
> Can you give some background on when this would be needed?

We needed to integrate several large applications using redux. We used the context for their providers and ran into a little problem in the selector hook that is bound to the general redux context.

Example:
```
import React from 'react';
import { Provider } from 'react-redux';
import { createRoot } from 'react-dom/client';

import { contextApp1, storeApp1 } from '#app-1/store';
import { contextApp2, storeApp2 } from '#app-2/store';

const $el = document.getElementById(import.meta.env.APP_HTML_ROOT_ELEMENT_ID);

if($el) {
    const root = createRoot($el);

    root.render(
        <React.StrictMode>
            <Provider
                store={storeApp1}
                context={contextApp1}
            >
                {/* APP 1 */}
            </Provider>
            <Provider
                store={storeApp2}
                context={contextApp2}
            >
                {/* APP 2 */}
            </Provider>
        </React.StrictMode>
    );
}
```

App store:
```
import { configureStore } from '@reduxjs/toolkit';
import { createDispatchHook, createSelectorHook, ReactReduxContextValue } from 'react-redux';
import { createContext } from 'react';

export const store = configureStore({
    reducer : {
        // ...
    }
    // ...
});

export type TStore = ReturnType<typeof store.getState>;

export const context = createContext<ReactReduxContextValue>(null as ReactReduxContextValue<TStore>);
export const useDispatch = createDispatchHook<TStore>(context);
export const useSelector = createSelectorHook(context); // This resolve incorrect types store useSelector((state) => state[])
```

markerikson:
Right, passing a custom context is what `createSelectorHook` is for.  I'm asking why the generic tweak is needed here, specifically, as opposed to using the approach we show in our docs:

https://redux.js.org/tutorials/typescript-quick-start#define-typed-hooks

This is what you do normally:

```ts
export const useAppSelector: TypedUseSelectorHook<RootState> = useSelector
```

And the same thing should work here:

```ts
export const useAppSelector: TypedUseSelectorHook<RootState> = createSelectorHook(SomeCustomContext)
```

lamo2k123:
> 

Oh, I was wrong. Thank you

PR #2022, opened 2023-06-03
Add warning for selectors that return the entire state

EskiMojo14:
builds on #2000


docs/api/hooks.md
```diff
@@ -240,6 +245,82 @@ export const App = () => {
...
+```ts
+// this selector will return a new object reference whenever called
+// meaning the component will rerender whenever *any* action is dispatched
+const { count, user } = useSelector((state) => ({
+  count: state.count,
+  user: state.user,
+}))
+```
+
+If a selector result is suitably stable, or memoised, it will not return a different result and thus not cause a warning to be logged.
```
msutkowski, line 265:
```suggestion
If a selector result is stable or memoized it will not return a different result or cause a warning to be logged.
```

🇺🇸

docs/api/hooks.md
```diff
@@ -240,6 +245,82 @@ export const App = () => {
...
+
+By default, this will only happen when the selector is first called. You can configure the check via context, or per `useSelector` call - either to run the check always, or never.
```
msutkowski, line 267:
```suggestion
By default, this will only happen when the selector is first called. You can configure the check in the Provider or at each `useSelector` call.
```

docs/api/hooks.md
```diff
@@ -240,6 +245,82 @@ export const App = () => {
...
+}
+```
+
+:::info
+This check is disabled for production environments.
+:::
+
+#### No-op selector check
+
+In development, a check is conducted on the result returned by the selector. It warns in console if the result is the same as the parameter passed in, i.e. the root state.
```
msutkowski, line 290:
```suggestion
In development, a check is conducted on the result returned by the selector. It warns in the console if the result is the same as the parameter passed in, i.e. the root state.
```

docs/api/hooks.md
```diff
@@ -240,6 +245,82 @@ export const App = () => {
...
+
+A `useSelector` call returning the entire root state is almost always a mistake, as it means the component will rerender whenever _anything_ in state changes. Selectors should select as specifically as possible.
```
msutkowski, line 292: Wonder if we should change `state changes` -> `in the store changes`?

msutkowski, line 292:
```suggestion
A `useSelector` call returning the entire root state is almost always a mistake, as it means the component will rerender whenever _anything_ in state changes. Selectors should be as granular as possible.
```

docs/api/hooks.md
```diff
@@ -240,6 +245,82 @@ export const App = () => {
...
+
+```ts no-transpile
+// this selector returns the entire state, meaning that the component will rerender unnecessarily
```
msutkowski, line 295:
```suggestion
// BAD: this selector returns the entire state, meaning that the component will rerender unnecessarily
```

docs/api/hooks.md
```diff
@@ -240,6 +245,82 @@ export const App = () => {
...
+const { count, user } = useSelector((state) => state)
+
+// instead, select only the state you need, calling useSelector as many times as needed
```
msutkowski, line 298:
```suggestion
// GOOD: instead, select only the state you need, calling useSelector as many times as needed
```

docs/api/hooks.md
```diff
@@ -240,6 +245,82 @@ export const App = () => {
...
+const count = useSelector((state) => state.count)
+const user = useSelector((state) => state.user)
+```
+
+By default, this will only happen when the selector is first called. You can configure the check via context, or per `useSelector` call - either to run the check always, or never.
```
msutkowski, line 303:
```suggestion
By default, this will only happen when the selector is first called. You can configure the check in the Provider or at each `useSelector` call.
```

src/components/Context.ts
```diff
@@ -1,6 +1,7 @@
 import { createContext } from 'react'
 import type { Action, AnyAction, Store } from 'redux'
 import type { Subscription } from '../utils/Subscription'
+import { CheckFrequency } from '../hooks/useSelector'
```
msutkowski, line 4:
```suggestion
import type { CheckFrequency } from '../hooks/useSelector'
```

src/components/Provider.tsx
```diff
@@ -3,6 +3,7 @@ import { ReactReduxContext, ReactReduxContextValue } from './Context'
 import { createSubscription } from '../utils/Subscription'
 import { useIsomorphicLayoutEffect } from '../utils/useIsomorphicLayoutEffect'
 import { Action, AnyAction, Store } from 'redux'
+import { CheckFrequency } from '../hooks/useSelector'
```
msutkowski, line 6:
```suggestion
import type { CheckFrequency } from '../hooks/useSelector'
```

msutkowski: We could probably be more consistent about `type` imports? 🤷

test/hooks/useSelector.spec.tsx
```diff
@@ -7,10 +7,11 @@ import React, {
   useState,
   useContext,
 } from 'react'
-import { createStore } from 'redux'
+import { Action, createStore } from 'redux'
```
msutkowski, line 10:
```suggestion
import { type Action, createStore } from 'redux'
```

test/hooks/useSelector.spec.tsx
```diff
@@ -7,10 +7,11 @@ import React, {
...
 import * as rtl from '@testing-library/react'
 import {
-  Provider as ProviderMock,
+  Provider,
+  ProviderProps,
```
msutkowski, line 14:
```suggestion
  type ProviderProps,
```

test/hooks/useSelector.spec.tsx
```diff
@@ -25,6 +26,22 @@ import type {
 } from '../../src/index'
 import type { FunctionComponent, DispatchWithoutAction, ReactNode } from 'react'
 import type { Store, AnyAction } from 'redux'
+import { UseSelectorOptions } from '../../src/hooks/useSelector'
```
msutkowski, line 29:
```suggestion
import type { UseSelectorOptions } from '../../src/hooks/useSelector'
```

msutkowski (commented):
Primarily just personal grammar preferences and 🪡 's. Feel free to ignore everything 👯 , otherwise looks prettttty good!

PR #2030, opened 2023-06-13
Add an E2E-ish test that verifies behavior when imported into RSCs

markerikson:
This PR:

- Adds a basic Next app with RSCs set up
- Imports `useSelector` into an RSC file, which _should_ currently cause an error when built
- Adds CI steps to build that project

Want this in place before I see if #2025 improves things


markerikson:
Yay, it fails to build!  Which is exactly what I want right now:

image

PR #2035, opened 2023-06-16
fix: Update hooks.md to fix custom equality function example 

VolkRiot:
Custom equality function documentation mixes up the new and previous value variables in the example code.

See logs in this code sandbox example: https://codesandbox.io/s/redux-useselector-usedispatch-sample-forked-zhfnx5?file=/src/Counter.js

You can also see the screenshot below following the example naming in the documentation. The logged result is after incrementing the count.

Screen Shot 2023-06-15 at 7 03 46 PM

### DIFF
--- a/docs/api/hooks.md
+++ b/docs/api/hooks.md
@@ -125,7 +125,7 @@ const selectedData = useSelector(selectorReturningObject, {
 import { useSelector } from 'react-redux'
 
 // equality function
-const customEqual = (oldValue, newValue) => oldValue === newValue
+const customEqual = (newValue, oldValue) => newValue === oldValue
 
 // later
 const selectedData = useSelector(selectorReturningObject, customEqual)


markerikson:
Huh. That's... extremely odd.

We rely on the `use-sync-external-store/withSelector` package for this, and it sure _looks_ like it's `(prev, current)` the couple places I see it called:

- https://github.com/facebook/react/blob/v18.2.0/packages/use-sync-external-store/src/useSyncExternalStoreWithSelector.js#L59-L85

VolkRiot:
> Huh. That's... extremely odd.
> 
> We rely on the `use-sync-external-store/withSelector` package for this, and it sure _looks_ like it's `(prev, current)` the couple places I see it called:
> 
> * https://github.com/facebook/react/blob/v18.2.0/packages/use-sync-external-store/src/useSyncExternalStoreWithSelector.js#L59-L85

Interesting, are they not possibly making the same error?

VolkRiot:
@markerikson Further playing around with code sandbox it looks like these values read correctly in ^8.0.0.
Apologies, I guess I just need to update my react-redux dependency

markerikson:
FWIW, it looks like we did indeed do `(new, old)` in v7!

https://github.com/reduxjs/react-redux/blob/7.x/src/hooks/useSelector.js#L40

PR #2038, opened 2023-06-19
Update packaging, build config, and entry points for v9

markerikson:
This PR:

- Switches the build setup to `tsup`
- Generates the usual set of ESM/CJS build artifacts I've set up for our other packages over the last few months, as well as a `react-redux/alternate-renderers` entry point to match our existing setup
- Updates all the `package.exports` fields to match
- Updates Jest to 29.  (I tried to switch to Vitest, but couldn't get it to work with the RN test)
  - Rewrites the RN `useSelector` test to use real `async/await` instead of fake timers
- Changes the peer deps to require React 18 and Redux 5.0-beta, and removes RTK (which I'll have to follow up on separately for 8.1.1 shortly)
- Updates the TS types to account for the `UnknownAction` and `Action<string>` requirements from the core
- Changes the primary entry point to assume React 18 and `useSyncExternalStore`, eliminating the need for the shim version of the `use-sync-external-store` package
  - Tweaks the export of `batch` so that TS doesn't need to refer to `react-dom`
- Deletes the `/next` entry point


EskiMojo14:
lgtm 😄 

fwiw Action<string> and Action are equivalent, it has a default

Issue #2020, opened 2023-06-02
Redux Toolkit Query 2.0 Beta Uncaught Error: could not find react-redux context value


jonaldinger:
### What version of React, ReactDOM/React Native, Redux, and React Redux are you using?

- React: 18.2.0
- ReactDOM/React Native: 18.2.0
- Redux: 5.0.0-beta.0
- Redux Toolkit: 2.0.0-beta.0
- React Redux: 8.0.7

### What is the current behavior?

When upgrading from react-redux 8.0.5 to 8.0.7, along side of Redux Toolkit 2.0.0-beta.0 + Redux 5.0.0-beta.0, my application code throws an error when using an RTK Query hook:

```
Uncaught Error: could not find react-redux context value; please ensure the component is wrapped in a <Provider>
    at useReduxContext (useReduxContext.js:24:11)
    at useStore2 (useStore.js:17:9)
    at useDispatch2 (useDispatch.js:14:19)
    at useQuerySubscription (buildHooks.ts:689:24)
    at useQuery (buildHooks.ts:964:42)
    at RequireAuthentication (RequireAuthentication.tsx:34:7)
```

The application code is wrapped in a Provider and has been functioning normally for months with the Redux Toolkit alphas, and is also functioning normally with the just-released Redux Toolkit beta. It is only when I upgrade to react-redux 8.0.7 that the error is thrown. When I downgrade to react-redux 8.0.5, without modifying any other dependencies, the error is not thrown.

Interestingly, `useSelector()`, which I would assume is accessing the same exact context from the provider, works fine. But `useGet*Query()` from RTK Query throws. Final note, I am using Yarn PnP and Yarn Workspaces. 

### What is the expected behavior?

No error should be thrown, as the component is wrapped in a provider.

### Which browser and OS are affected by this issue?

Chrome 113 on macOS Ventura

### Did this work in previous versions of React Redux?

- [X] Yes

EskiMojo14:
My assumption is that you're using a custom context, and providing the hooks to the react API module.

One of the breaking changes made in RTK's alphas/betas (I'm unsure which one) is that the signature of reactHooksModule changed: https://github.com/reduxjs/redux-toolkit/pull/3400

```ts
reactHooksModule({
    hooks: {
      useDispatch: createDispatchHook(MyContext),
      useSelector: createSelectorHook(MyContext),
      useStore: createStoreHook(MyContext),
    },
  })
```

This is likely the main thing that's causing your issue.

[...]

Issue #2029, opened 2023-06-11
Investigate per-React-version context instance caching


markerikson:
Per Lenz, we could try something like this:

https://github.com/apollographql/apollo-client/blob/2446acef85d5ca8af378aa11d8002b84f0e4aa38/src/react/context/ApolloContext.ts#L32-L54

phryneas:
But using `React.version` as a key - we noticed that using `React` as a key actually doesn't work, as that is a different object in each importing file.

phryneas:
@TobbeLino could you create a reproduction for this?

phryneas:
At this point I'd say we wait for a reproduction from you, because this seems like something `preact` definitely should not be doing - they should at least pretend to be `17.0.2-preact-10.15.1`, in the spirit of e.g. React version [`18.3.0-canary-e3fb7c1de-20230621`](https://www.npmjs.com/package/react/v/18.3.0-canary-e3fb7c1de-20230621) and on first look, this might be something they should be fixing on their end rather than us trying to work around it.

In the meantime, you can create a custom context in your Preact applications as a workaround, as described here: https://react-redux.js.org/api/hooks#custom-context

TobbeLino:
@phryneas Yeah, I know, they do this:
https://github.com/preactjs/preact/blob/523bbe93ccda2c5806284469e0e46f47e84e63b6/compat/src/index.js#L38
:) But I suppose some libraries will break if they use a custom version string...

Anyway, thanks a lot for the hint/workaround with custom context - it worked!
I'll try to investigate further and see if I can come up with something reproducible for you as well.

phryneas:
Well, React already ships with these "weird" custom version strings, so I don't think anything would break if preact would adopt that same style... :thinking: 

Great that the workaround is working for you!

TobbeLino:
@phryneas I managed to create a repo demonstrating these issues. It's a bit messy - sorry, but you should be able to reliably reproduce what I observed: https://github.com/TobbeLino/react-redux-issue

The repo can be used to test running two independent modules at the same time, both Preact (that causes a crash when "module 2" is loaded), and React 17.0.2 (which does not crash, but prints in the console: "Warning: Detected multiple renderers concurrently rendering the same context provider. This is currently unsupported.")

[...]

PR #2039, opened 2023-06-20
create singleton context by React version

phryneas:
Going forward, this should solve the "singleton context problem" as encountered e.g. in https://github.com/reduxjs/react-redux/issues/2020#issuecomment-1597881859

### DIFF
--- a/src/components/Context.ts
+++ b/src/components/Context.ts
@@ -1,4 +1,4 @@
-import { createContext } from 'react'
+import { createContext, version as ReactVersion } from 'react'
 import type { Context } from 'react'
 import type { Action, AnyAction, Store } from 'redux'
 import type { Subscription } from '../utils/Subscription'
@@ -15,13 +15,17 @@ export interface ReactReduxContextValue<
   noopCheck: CheckFrequency
 }
 
-let realContext: Context<ReactReduxContextValue> | null = null
+const ContextKey = Symbol.for(`react-redux-context-${ReactVersion}`)
+const gT = globalThis as { [ContextKey]?: Context<ReactReduxContextValue> }
+
 function getContext() {
+  let realContext = gT[ContextKey]
   if (!realContext) {
     realContext = createContext<ReactReduxContextValue>(null as any)
     if (process.env.NODE_ENV !== 'production') {
       realContext.displayName = 'ReactRedux'
     }
+    gT[ContextKey] = realContext
   }
   return realContext
 }


markerikson:
Wow. Lenz, you are a genius. And crazy :)

TobbeLino:
Doesn't this fix totally breaks running more than one react application at the same time on a single webpage if they happen to run the same version of react?

We are developing applications for customers to run on their existing web pages. Our application uses react, and many customers also have other react-applications running on their pages. Our app needs to be able to co-exist, and we have no control over what the customers are already running.

Imho, caching things in the global browser scope is not a good idea, becuase totally different instances of React (from different applications) should not use the same context.

phryneas:
@TobbeLino but they totally are, and that's actually the best case.

Imagine two independent React trees that both use `react-redux`. In a perfect world, the `react-redux` package would only be shipped to the browser once, and both of those applications would import that same package. In the past, the `ReduxContext` was a package-wide singleton, so in this scenario, those two independent applications were already using the same "Context object".  
Things started breaking when people accidentally had one React version and multiple different `react-redux` versions in their bundle. That's what this PR tries to avoid.

This `ReduxContext` object is not an object that actually contains the context value (well, it does while rendering a fiber, but that's an implementation detail). Multiple different applications have often been using that same `ReduxContext` object and it has never been a problem. 
The context *value* is still scoped to your component tree and only available to the components below the `ReduxContext.Provider` within an application, even if you had a dozen applications literally using the same `ReduxContext` object.

The only thing we do here is distinguish between multiple different `ReduxContext` versions for different React versions, because we want to make sure everything works if React starts changing context internals from one version to another.

TobbeLino:
Hmmm, ok, I admit this is a bit over my head :) 
I'll continue discussion in https://github.com/reduxjs/react-redux/issues/2029

src/components/Context.ts
```diff
@@ -15,13 +15,17 @@ export interface ReactReduxContextValue<
   noopCheck: CheckFrequency
 }
 
-let realContext: Context<ReactReduxContextValue> | null = null
+const ContextKey = Symbol.for(`react-redux-context-${ReactVersion}`)
+const gT = globalThis as { [ContextKey]?: Context<ReactReduxContextValue> }
```
akgayush, line 19: `globalThis` is not present in Android webview version < 71. I faced issues in Android OS version <= 8.1.0 due to this.

PR #2045, opened 2023-06-24
map context instances using `createContext` as key

phryneas:
@TobbeLino could you please try the CodeSandbox package build for this PR and give us feedback if it solves the problem?

```
npm i https://pkg.csb.dev/reduxjs/react-redux/commit/cd8a93bd/react-redux
```

### DIFF
--- a/src/components/Context.ts
+++ b/src/components/Context.ts
@@ -1,4 +1,4 @@
-import { createContext, version as ReactVersion } from 'react'
+import { createContext } from 'react'
 import type { Context } from 'react'
 import type { Action, AnyAction, Store } from 'redux'
 import type { Subscription } from '../utils/Subscription'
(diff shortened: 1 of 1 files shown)


TobbeLino:
@phryneas Great! Yes, my test cases (and also our preact apps) run fine now with this PR!

src/components/Context.ts
```diff
@@ -15,17 +15,20 @@ export interface ReactReduxContextValue<
   noopCheck: CheckFrequency
 }
 
-const ContextKey = Symbol.for(`react-redux-context-${ReactVersion}`)
-const gT = globalThis as { [ContextKey]?: Context<ReactReduxContextValue> }
+const ContextKey = Symbol.for(`react-redux-context`)
+const gT = globalThis as {
```
markerikson, line 19: Per https://github.com/reduxjs/react-redux/pull/2039#pullrequestreview-1524058384 , can we do some more checks to see if `globalThis` even exists?

phryneas: Given that Android 8.1 mentioned in that comment had it's official support from Google end on October 4, 2021, I would be very happy if we could avoid that - or at least only add it temporarily and remove it in a 9.0 release.

phryneas:
Let's combine this with  #2050

PR #2050, opened 2023-07-04
RSC-specific workarounds

phryneas:
* use `{}` as ReactReduxContext in environments where no context exists
* switch React imports to namespace imports to prevent hook import detection

### DIFF
--- a/src/components/Context.ts
+++ b/src/components/Context.ts
@@ -1,4 +1,4 @@
-import { createContext } from 'react'
+import * as React from 'react'
 import type { Context } from 'react'
 import type { Action, AnyAction, Store } from 'redux'
 import type { Subscription } from '../utils/Subscription'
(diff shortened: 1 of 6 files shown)


markerikson:
Is that import switch going to cause any kind of bundle size issue?

phryneas:
> (Although... I guess React doesn't get tree shaken anyway...)

Pretty much this. All it does is slightly increase the size of the cjs build because it now has to add a namespace import helper function, but no bigger bundle size changes beyond that.

src/components/Context.ts
```diff
@@ -15,34 +15,30 @@ export interface ReactReduxContextValue<
...
+const ContextKey = Symbol.for(`react-redux-context`)
+const gT = globalThis as {
+  [ContextKey]?: Map<
+    typeof React['createContext'],
+    Context<ReactReduxContextValue>
+  >
+}
+
+function getContext(): Context<ReactReduxContextValue> {
+  if (!React.createContext) return {} as any
```
phryneas, line 27: In an environment where `createContext` does not exist we can pretty much just return anything - there is no `useContext` that could consume it anyways.

EskiMojo14: and the provider needs to be in a client component, so presumably that won't be an issue either

EskiMojo14 (approved):

phryneas:
I just cherrypicked this commit https://github.com/reduxjs/react-redux/pull/2045/commits/b37e57e9eb8981b03b92a09c58bd15664cb64a12 in here since I had forgotten to merge that over earlier. @EskiMojo14 could you take another look please?

EskiMojo14:
yep, still lgtm

PR #2064, opened 2023-08-23
Add stack to dev mode checks

EskiMojo14:


EskiMojo14:
reused an old CSB to test: [link](https://codesandbox.io/s/billowing-lake-nytcmq?file=/src/App.tsx)

it's beautiful
image

EskiMojo14:
stack is a [non-standard property](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/stack), so i don't think we can come up with an approach to parse it reliably unfortunately

PR #2067, opened 2023-09-13
Use default Node version in CI

nickmccurdy:
Deprecated Node versions are used in CI, potentially causing security and reliability issues. Instead, it's better to use GitHub's default Node version, which also doesn't require additional downloads or installations.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -14,7 +14,6 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-node@v3
         with:
-          node-version: '18.x'
           registry-url: 'https://registry.npmjs.org'
           cache: 'yarn'
       - run: yarn install --frozen-lockfile
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,7 +19,6 @@ jobs:
       - name: Set up Node
         uses: actions/setup-node@v3
         with:
-          node-version: 16.x
           cache: 'yarn'
 
       - name: Install dependencies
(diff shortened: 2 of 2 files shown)


nickmccurdy:
The size action is warning because of https://github.com/preactjs/compressed-size-action/issues/93 but it seems like it's safe to ignore as it already runs on Node 16:

> The following actions uses node12 which is deprecated and will be forced to run on node16: preactjs/compressed-size-action@v2. For more info: https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/

timdorr:
This isn't a hard opinion, but I'm on the negative side of this. This introduces less determinism into our builds and test suite. That has the potential for breakage and extra workload for us as maintainers. 

There's nothing necessarily wrong with using older Node versions, since this isn't deployed software and the practical effects on end-users are minimal. The most likely actual breakage is with the versions of tools that we use and the supported Node versions of our various Actions. In that context, I think it's more important to keep builds and tests consistent and as close to deterministic as possible.

nickmccurdy:
I believe it's more deterministic, as at least it's using the same default Node version in each workflow. Also Mark has already agreed to my idea, but I'm curious what he thinks about this specifically.

timdorr:
That's consistency, not determinism. The build environment will change over time. A run one week under Node 20 might be different or break the next week under Node 22. It's the same reason we use a lockfile for our dependencies.

I'm all for consistency. I have issue with nondeterminism.

balramshejal:
test

PR #2068, opened 2023-09-14
Fix useSelector() in combination with lazy loaded components breaks with react v18 (#1977)

### LINKED ISSUE #1977: useSelector() in combination with lazy loaded components breaks with react v18

- React: 18.2.0
- ReactDOM/React Native: 18.2.0
- React Redux: 8.0.5

1. Open the codesandbox ( https://codesandbox.io/s/elegant-hofstadter-i2h34h?file=/src/features/counter/InnerCounter.js )
2. Open the  page in a new browser window
3. Click the "INCREMENT ME!" text
4. If the number in the first row increments -> Full reload the page, repeat 3.
5. The number does not increase although the store is updated (see console.log output) but the selector does not trigger

Additional note: This works correctly with react v17 (ReactDOM.render) and breaks in v18 (createRoot). I have also debugged the behavior and have seen that the Provider component is unmounted which removes the store subscriptions of the useSelector hooks. After the Provider mounts again, the subscriptions are not restored because the child components do not render again.

The number is increased on every click.

all

- [X] Yes

jeroenpx:
Fixes #1977

Includes new unit test which can reproduce the issue.

### DIFF
--- a/src/utils/Subscription.ts
+++ b/src/utils/Subscription.ts
@@ -99,9 +99,26 @@ export function createSubscription(store: any, parentSub?: Subscription) {
   let unsubscribe: VoidFunc | undefined
   let listeners: ListenerCollection = nullListeners
 
+  // Reasons to keep the subscription active
+  let subscriptionsAmount = 0
+
+  // Is this specific subscription subscribed (or only nested ones?)
+  let selfSubscribed = false
+
   function addNestedSub(listener: () => void) {
     trySubscribe()
-    return listeners.subscribe(listener)
+
+    const cleanupListener = listeners.subscribe(listener)
+
+    // cleanup nested sub
+    let removed = false
+    return () => {
+      if (!removed) {
+        removed = true
+        cleanupListener()
+        tryUnsubscribe()
+      }
+    }
   }
 
   function notifyNestedSubs() {
@@ -115,10 +132,11 @@ export function createSubscription(store: any, parentSub?: Subscription) {
   }
 
   function isSubscribed() {
-    return Boolean(unsubscribe)
+    return selfSubscribed
   }
 
   function trySubscribe() {
+    subscriptionsAmount++
     if (!unsubscribe) {
       unsubscribe = parentSub
         ? parentSub.addNestedSub(handleChangeWrapper)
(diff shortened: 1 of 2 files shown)


markerikson:
Oh wow, that's great!

How did you investigate and figure this out?

jeroenpx:
Well, it started with running into the issue in an application at work. The useSelectors suddenly stopped updating for no reason after doing some code changes.

I ended up finding #1977 which described exactly the situation I was in and indicated that it could be because the Provider component removed all the nested subscriptions and then doesn't re-add them again. 

I looked into the source of useSelector to see if I could re-add the subscription somehow but that didn't look like the right place. It has no knowledge of the Provider component some number of levels above in the tree that removed the subscription seemingly at random. Solution is to have the Provider not clean up subscriptions it didn't create itself.

I still didn't really get why the Provider was getting cleaned up and re-created, but while writing the unit test based on the example in #1977 and trying to simplify it to the very minimal steps needed, I realized that this is something specific to a Suspense component. It can temporary remove the items from the page (which cleans up the useEffects) and show "loading" and then re-add the *existing* items again once it is done loading the newly added subcomponent. All the existing components that were already there in the Suspense though are simply re-added without actually re-rendering anything. It doesn't run anything except the useEffects in those components again. It keeps all the useState and the useSyncExternalStoreWithSelector subscriptions as it had it before.

This confirmed for me that the Provider should not be cleaning up the nested subscriptions because those components might not get re-rendered even if the Provider component is temporary removed from the DOM to show the loading and then re-added without changes afterwards (and most importantly without re-render). It confirmed for me that the fix I did made sense from a React life-cycle point of view.

See also step by step reproduction in the unit test.

markerikson:
Okay, finally had time to look at this.

Changes seem reasonable. Still haven't fully grokked the exact sequence, but this seems to match my general expectations of what might be going on with React Suspense.

More than that: **_thank you_ for including such an _excellent_ test!** Having a test here, _and_ with fantastic comments about what behavior we expect to happen, is a _huge_ help to me as a maintainer when reading something like this! I'm very grateful that you added that.

Let's get this in!

PR #2083, opened 2023-11-13
Typo : extra `;` in code block

jdeniau:
Remove this extra `;`

image


timdorr:
Looks like something leftover from an older version of Prettier. Thanks!

Issue #2106, opened 2023-12-11
Unable to resolve peer dependency react-native on react-native@0.73.0


R3DST0RM:
### What version of React, ReactDOM/React Native, Redux, and React Redux are you using?

- React: 18.2.0
- ReactDOM/React Native: 0.73.0 (React-Native)
- Redux: 5.0.0
- React Redux: 9.0.3 (trying to install)

### What is the current behavior?

Trying to install `react-redux@9.0.3` causes NPM to throw a could not resolve dependency error.

```bash
npm ERR! While resolving: react-redux@9.0.3
npm ERR! Found: react-native@0.73.0
npm ERR! 
npm ERR! Could not resolve dependency:
npm ERR! peerOptional react-native@"^0.69.0" from react-redux@9.0.3
npm ERR! node_modules/react-redux
npm ERR!   react-redux@"*" from the root project
npm ERR! 
npm ERR! Conflicting peer dependency: react-native@0.69.12
npm ERR! node_modules/react-native
npm ERR!   peerOptional react-native@"^0.69.0" from react-redux@9.0.3
npm ERR!   node_modules/react-redux
npm ERR!     react-redux@"*" from the root project
npm ERR! 
npm ERR! Fix the upstream dependency conflict, or retry
npm ERR! this command with --force or --legacy-peer-deps
npm ERR! to accept an incorrect (and potentially broken) dependency resolution.
```

The problem is reproducible by:

```bash
npx react-native@latest init Foo
cd Foo
npm install --save redux redux-thunk reselect
npm install --save react-redux 
```

### What is the expected behavior?

Installation should succeed without any error

### Which browser and OS are affected by this issue?

macOS

### Did this work in previous versions of React Redux?

- [X] Yes

markerikson:
*sigh* that probably needs to be a `>=`, I guess.

I hate NPM's peer dep handling.

PR #2107, opened 2023-12-11
Allow react-native newer than 0.69 as peer dependency

R3DST0RM:
As described in https://github.com/reduxjs/react-redux/issues/2106, `react-redux@9.0.3` could not be installed due to conflicting peerDependency `react-native`.

PeerDependency `react-native` requires a minimum version of 0.69.0, while allowing newer versions, this pull request fixes https://github.com/reduxjs/react-redux/issues/2106 the NPM dependency handling introduced in 9.0.3.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -9415,7 +9415,7 @@ __metadata:
   peerDependencies:
     "@types/react": ^18.2.25
     react: ^18.0
-    react-native: ^0.69.0
+    react-native: ">=0.69"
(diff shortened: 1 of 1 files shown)


markerikson:
Can you also rerun `yarn` to regenerate the lockfile?

R3DST0RM:
> Can you also rerun `yarn` to regenerate the lockfile?

Of course. Done

PR #2110, opened 2024-01-02
Update hooks.md — reselect usage with multiple instances simplified

VorontsovIE:
It seems that modern reselect allows one to share the same selector instance in multiple components, so docs can be simplified.


timdorr:
Thanks!

PR #2122, opened 2024-01-24
Remove unused isProcessingDispatch

Connormiha:
ref `isProcessingDispatch` never used.


timdorr:
Yep, this looks like some vestigial code from when we did the switchover to uSES. Thanks!

PR #2123, opened 2024-01-24
Move `Equals` constraint into an intersection type.

DanielRosenwasser:
In https://github.com/microsoft/TypeScript/issues/57117, we noticed that react-redux's build is impacted by a change in in https://github.com/microsoft/TypeScript/issues/56515. What this means is that TypeScript 5.4, you'll get an error in `expectExactType`.

```ts
export function expectExactType<T>(t: T) {
  return <U extends Equals<T, U>>(u: U) => {}
//        ~
// error TS2313: Type parameter 'U' has a circular constraint.
}
```

The solution I would suggest for now is to just make `U` extend `T` and move the constraint to an intersection wherever you use `U`.

```diff
-  return <U extends Equals<T, U>>(u: U) => {}
+  return <U extends T>(u: U & Equals<T, U>) => {}
```

### DIFF
--- a/test/typeTestHelpers.ts
+++ b/test/typeTestHelpers.ts
@@ -47,7 +47,7 @@ export declare const exactType: <T, U>(
 ) => IfEquals<T, U>
 
 export function expectExactType<T>(t: T) {
-  return <U extends Equals<T, U>>(u: U) => {}
+  return <U extends T>(u: U & Equals<T, U>) => {}
 }
 
 type EnsureUnknown<T> = IsUnknown<T, any, never>


timdorr:
Thanks, Daniel! This is only in our test suite, so it shouldn't be a user-facing issue, but I very much appreciate the attention to detail on your end and the PR!

DanielRosenwasser:
@EskiMojo14 looks like this is from use of `any` in the conditional type. It seems like using `unknown` fixes the issue.

[Playground link](https://www.typescriptlang.org/play?ts=5.4.0-dev.20240126#code/KYDwDg9gTgLgBDAnmYcCSBnAggO0QHgBUAaOQqAV2FIDEBDAGw1QF44dgA3YKAPjhYAoOHAD0ohMAzwAlgDM4Ad1R0oqAOYQZOdQgAWqBsDnwsAOQAicKDPV74YOjD1xt+1AGMIOACYyYMt7CCJSoAD5wcozMcKAwwL4YcAAUhLEg8YnsXDxwAPwhVHAAXJHRwACU+WShJXD0TMCCgkgocACiAI4U0USkAKr8bJi4BMEkwRzcUMTBI3j4-aRTPKQA2oQAuumZPklr-dsFyQfbcQl7cBtHcBQ4ANY4EIo4dStQVaXvvII-gqCQWCRO4eALedIoUHtEB0UGEZDAADyDB8RF4yRgpUIVQA3sE1DAKFBXosdhckl0ekw+nBBuiKKV+lUWPwcQBfQQc-7gaDwOQgsGvAHAKEwuEIszARRojFY3H44CE4lwUnnLKEemMuAAMg63V6JFpvGZrI5XJ8IoYqk83mkcAgYEFjAAyjAbDpStJ3boIncLXJtMAfABuZriOAAARgGAAtMLQXGoFBoNzITBobCYPCUMifMkAOT5irJB1Ohiu70VVMi9NirMI3MF-NwOhJL3aH23XzGQM+YulwI4F1ujtV+O1zPZpEokuOwfDytNquCcNR2PjxPJqDV0WTiVSpezssV0c7ifilCSxRNlttkc6OC+7sBjh9o-z8v39RV1fRuPgGtNxTccMwvK9328BdRyXZogA)

Let me know if that works!

EskiMojo14:
that seems to do the trick, thanks!

in the slightly more long term I think we'll switch to vitest's type test utilities, so hopefully those will continue to work :)

PR #2124, opened 2024-01-25
Migrate from Jest to Vitest

aryaemami59:
This PR:

  - [X] Migrates test configurations from Jest to Vitest in order to stay consistent with the rest of the redux packages.
  - [X] Drop Jest related packages.
  - [X] Drop React Native related tests.


markerikson:
FYI, I ran into issues here with the RN test when I was trying to update last year, which is why I stayed on Jest at that point.  (At least now with v9 we've dropped the multiple aliased tests for React 17 and such that we had with v8.)

timdorr:
Since we're no longer testing RN, this makes a lot of sense. Plus, removing a net 3,315 lines from the yarn.lock file is great!

I'll let @markerikson merge, since he's closer to this stuff than me. But LGTM!

PR #2125, opened 2024-01-28
Format all files

aryaemami59:


docs/api/connect.md
```diff
@@ -125,7 +125,7 @@ The second parameter is normally referred to as `ownProps` by convention.
 
 ```js
 // binds on component re-rendering
-<button onClick={() => this.props.toggleTodo(this.props.todoId)} />
+;<button onClick={() => this.props.toggleTodo(this.props.todoId)} />
```
EskiMojo14, line 128:
```suggestion
<button onClick={() => this.props.toggleTodo(this.props.todoId)} />
```
any idea how to prevent this happening? i understand why it's being added but it looks bad

aryaemami59: Yeah that doesn't look good, let me see if there is any way I can make that go away.

PR #2129, opened 2024-02-08
Migrate type tests to Vitest

aryaemami59:
This PR:

  - [X] Converts `typetest.ts` file extensions to `test-d.ts` to be compatible with Vitest.
  - [X] Converts `expectType()` to either `assertType()` or `expectTypeOf().toMatchTypeOf()`.
  - [X] Converts `expectExactType` to `expectTypeOf().toEqualTypeOf()`.
  - [X] Converts lone-blocks to `test` or `describe` blocks.


.github/workflows/test.yml
```diff
@@ -42,6 +39,42 @@ jobs:
...
+      - name: Use node ${{ matrix.node }}
+        uses: actions/setup-node@v3
+        with:
+          node-version: ${{ matrix.node }}
+          cache: 'yarn'
+
+      - name: Install deps
+        run: yarn install
+
+      - uses: actions/download-artifact@v3
```
timdorr, line 63:
```suggestion
      - uses: actions/download-artifact@v4
```
Needs to be v4 to find the upload

.github/workflows/test.yml
```diff
@@ -42,6 +39,42 @@ jobs:
...
+    strategy:
+      fail-fast: false
+      matrix:
+        node: ['20.x']
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v3
+
+      - name: Use node ${{ matrix.node }}
+        uses: actions/setup-node@v3
```
timdorr:
```suggestion
        uses: actions/checkout@v4

      - name: Use node ${{ matrix.node }}
        uses: actions/setup-node@v4
```

We should also bump these to v4

.github/workflows/test.yml
aryaemami59: Yes sirrrr

markerikson:
Haven't paid any attention to this, I'm afraid :) Where's it stand? I see green.

Also, just out of curiosity, what's the net benefit of doing typetests with Vitest?

aryaemami59:
@markerikson All tests are passing.

> Also, just out of curiosity, what's the net benefit of doing typetests with Vitest?

1. Type tests become more semantic and easier to read.
2. We can take advantage of the type checking utilities Vitest provides.
3. We can use the `.skip`/`.only`/`.todo` syntax.
4. We can run both runtime and type tests by using the `--typecheck` flag.

On a side note I'm also a collaborator on the [`expect-type`](https://github.com/mmkal/expect-type) which is the library Vitest uses for type checking.

markerikson:
Okay, cloned it briefly and confirmed A) it passes locally, B) it does catch errors if I introduce them. Nice!

PR #2135, opened 2024-02-21
Upgrade to Yarn 4

aryaemami59:
**This PR**:

  - [X] Bumps Node versions in CI to 20
  - [X] Upgrades package manager to Yarn 4
  - [X] Updates `actions/checkout` usages to v4


.codesandbox/ci.json
```diff
@@ -1,6 +1,6 @@
 {
   "sandboxes": ["vanilla", "vanilla-ts"],
-  "node": "14",
+  "node": "18",
```
timdorr, line 3: This is probably outside the scope of this PR, but do we still need Codesandbox in the CI setup? It seems like unnecessary complexity.

aryaemami59: I have no idea, I would be happy to get rid of it for you if you in a separate PR if you decide it shouldn't stay.

phryneas: It gives us immediate releases to play around on every release, and people get sandboxes where those releases are already preinstalled.

phryneas: For this PR, that would be 
grafik

Unfortunately, it's a bit hidden:
grafik

.github/workflows/test.yml
```diff
@@ -82,7 +82,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node: ['18.x']
+        node: ['20.x']
     steps:
       - name: Checkout repo
         uses: actions/checkout@v3
```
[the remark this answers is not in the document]
aryaemami59: Sure thing!

timdorr (approved):
Got rid of those warnings, except for the compressed-size-action that doesn't have an upgrade path yet. LGTM!

aryaemami59:
@timdorr Were the warnings caused by `actions/setup-node`?

timdorr:
That and the artifact actions. I got everything I could.

aryaemami59:
@timdorr Thank you so much!

timdorr:
Thanks!

PR #2138, opened 2024-03-04
Update typescript.md

Stabarak:
adding a missing line for exporting the store


timdorr:
Thanks, but we're just highlighting the export of the types from this file. This is not a complete version of the file, so it's assumed you export the store instance elsewhere.

PR #2139, opened 2024-03-06
Update TypeScript to 5.4

aryaemami59:
### This PR:

  - [X] Bumps TypeScript version to 5.4.
  - [X] Adds TypeScript 5.4 to TypeScript versions to test against during CI.


.github/workflows/test.yml
```diff
@@ -51,7 +51,7 @@ jobs:
       fail-fast: false
       matrix:
         node: ['18.x']
-        ts: ['4.7', '4.8', '4.9', '5.0', '5.1', '5.2', '5.3']
+        ts: ['4.7', '4.8', '4.9', '5.0', '5.1', '5.2', '5.3', '5.4']
```
timdorr, line 54: Do we want to drop some older versions at this point? 4.7 was released May 2022, 4.8 in August 2022, and 4.9 in November 2022. 

I say we drop 4.7 and 4.8, since users that treat TS versions like semver may still be stuck on 4.9.

EskiMojo14: generally i think we shouldn't drop testing against TS versions until we release a major version, to accidental breaking changes

EskiMojo14:
thanks!

PR #2155, opened 2024-04-07
Remove `no-resolution` flag and add `--exclude-entrypoints` when running `are-the-types-wrong` during CI

aryaemami59:
`@arethetypeswrong/cli` now has an `--exclude-entrypoints` flag which we can use to exclude the `alternate-renderers` entrypoint from the check since it will fail under the `Node10` module resolution as `Node10` looks at actual files on disk and we don't have a file named `alternate-renderers`.


timdorr:
Oh good! This was bugging me as a false positive. Thanks!

Issue #1510, opened 2020-01-28
Nested `connect` updates are not batched


OliverJAsh:
**Do you want to request a _feature_ or report a _bug_?**

(If this is a _usage question_, please **do not post it here**—post it on [Stack Overflow](http://stackoverflow.com/questions/tagged/redux) instead. If this is not a “feature” or a “bug”, or the phrase “How do I...?” applies, then it's probably a usage question.)

Bug

**What is the current behavior?**

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem. Your bug will get fixed much faster if we can run your code and it doesn't have dependencies other than React. Paste the link to a CodeSandbox (https://codesandbox.io/s/new) or RN Snack (https://snack.expo.io/) example below:**

https://stackblitz.com/edit/react-redux-batching-xkhtq8?file=index.tsx

**What is the expected behavior?**

1. Open https://react-redux-batching-xkhtq8.stackblitz.io/
2. Open dev tools -> React profiler
3. Click "reload and start profiling"
4. After reload, end profiling

Expected: 2 commits:

1. First render
2. `Counter` _and_ `ReduxCounter` update, triggered via `dispatch` inside `Counter#componentDidMount`

Actual: 3 commits:

1. First render
2. `Counter` update, triggered via `dispatch` inside `Counter#componentDidMount`
3. `ReduxCounter` update, triggered via `dispatch` inside `Counter#componentDidMount`

IIUC, there is no reason why 2 and 3 can not be batched into a single React re-render?

![Screen Recording 2020-01-28 at 16 30 28](https://user-images.githubusercontent.com/921609/73283925-b51e4000-41eb-11ea-844b-7750753c42fe.gif)

**Which versions of React, ReactDOM/React Native, Redux, and React Redux are you using?  Which browser and OS are affected by this issue? Did this work in previous versions of React Redux?**

![image](https://user-images.githubusercontent.com/921609/73283414-f8c47a00-41ea-11ea-963f-3ea90c9f8c6d.png)

markerikson:
My first guess is that it's specifically because you're dispatching in `componentDidMount`.  When you queue a React state update in `cDM`, React has to _synchronously_ re-render, immediately, before the browser can re-paint.

What happens if you move the call into a single button click, instead?

OliverJAsh:
Unfortunately the same happens inside a click handler: https://stackblitz.com/edit/react-redux-batching-nodjcz?file=index.tsx

[...]

Issue #1436, opened 2019-10-29
`useIsomorphicLayoutEffect` cannot pass test cases in non-web environment


malash:
**Do you want to request a _feature_ or report a _bug_?**

Report a bug.

**What is the current behavior?**

https://github.com/reduxjs/react-redux/blob/2eac86163be2bd5627dab3e33e8b4e0926895442/src/hooks/useSelector.js#L22-L27

https://github.com/reduxjs/react-redux/blob/672540023769845d6d162b382d518945321956ab/src/components/connectAdvanced.js#L40-L45

These two files implement the `useIsomorphicLayoutEffect` hook to [prevent the React warning](https://github.com/reduxjs/react-redux/issues/1402) by using `useEffect` instead of `useLayoutEffect`. But it could break on non-web environment, like React Native, [and etc](https://github.com/chentsulin/awesome-react-renderer).

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem. Your bug will get fixed much faster if we can run your code and it doesn't have dependencies other than React. Paste the link to a CodeSandbox (https://codesandbox.io/s/new) or RN Snack (https://snack.expo.io/) example below:**

I tested by these steps:

1. Clone the `react-redux` repo
2. Modify `useIsomorphicLayoutEffect` to `const useIsomorphicLayoutEffect = useEffect`
3. Run `npx jest`

![image](https://user-images.githubusercontent.com/1812118/67777832-47079700-fa9d-11e9-898c-da9a9017bf7a.png)

**What is the expected behavior?**

1. We should add some test cases for non-web environment. For example [@testing-library/react-native](https://testing-library.com/docs/native-testing-library/intro) provides the way to test `react-native` in jest.
2. We should review the implement of `useIsomorphicLayoutEffect` and fix it if there did have some bugs.

**Which versions of React, ReactDOM/React Native, Redux, and React Redux are you using?  Which browser and OS are affected by this issue? Did this work in previous versions of React Redux?**

react-redux@7.1.1

**cc**

@byunicorn

markerikson:
No one has yet filed actual bugs reporting this doesn't work in React Native.  Do you have any actual test cases that show it fails in an RN app?

malash:
😂Bug happened here https://github.com/reduxjs/react-redux/issues/1437

markerikson:
Yep, but I wasn't sure how to try to add RN unit tests at the time.

malash:
[...]

PR #2156, opened 2024-04-10
Fix `useIsomorphicLayoutEffect` usage in React Native environments

### LINKED ISSUE #2150: Connected children components might have mapStateToProps not called

- React: 18.2.0
- ReactDOM/React Native: 0.73.5
- Redux: 5.0.1
- React Redux: 9.1.0

In my react native app, I have connected parent components which might contain connected children components (react navigation top material tabs). The child components might dispatch action to change its state, but sometimes the state of the child components is not refreshed, mapStateToProps is not called at all. It's not an issue about mutable state because if I change to another tab and return back, the state is correctly refreshed and subsequent state change after this is correctly refreshed. I moved back from 9.1.0 to 8.1.3, and the problem is solved.

mapStateToProps should be called whenever store state change.

react native ios

- [X] Yes

aryaemami59:
## __Overview__

It looks like React Native was not using `useLayoutEffect` which was causing nested component updates to not get properly batched when using the `connect` API. 

## __This PR__:

  - [X] Fixes `useIsomorphicLayoutEffect` usage in React Native environments. 
  - [X] Fixes #2150.
  
Possible related issues: #1313, #1444, #1510, #1436.
  
I did test the actual build with `yalc`, React Native, Expo and Expo web seem to be working properly. We can probably get rid of [`src/utils/useIsomorphicLayoutEffect.native.ts`](https://github.com/reduxjs/react-redux/blob/master/src/utils/useIsomorphicLayoutEffect.native.ts) file, so let me know if you want me to do that I can include it in this PR or a separate one.

### DIFF
--- a/src/utils/useIsomorphicLayoutEffect.ts
+++ b/src/utils/useIsomorphicLayoutEffect.ts
@@ -16,6 +16,18 @@ export const canUseDOM = !!(
   typeof window.document.createElement !== 'undefined'
 )
 
+// Under React Native, we know that we always want to use useLayoutEffect
+
+/**
+ * Checks if the code is running in a React Native environment.
+ *
+ * @see {@link https://github.com/facebook/react-native/issues/1331 Reference}
+ */
+export const isReactNative =
+  typeof navigator !== 'undefined' && navigator.product === 'ReactNative'
+
 export const useIsomorphicLayoutEffect = canUseDOM
   ? React.useLayoutEffect
-  : React.useEffect
+  : isReactNative
+    ? React.useLayoutEffect
+    : React.useEffect
--- a/src/utils/useIsomorphicLayoutEffect.ts
+++ b/src/utils/useIsomorphicLayoutEffect.ts
@@ -26,8 +26,5 @@ export const canUseDOM = !!(
 export const isReactNative =
   typeof navigator !== 'undefined' && navigator.product === 'ReactNative'
 
-export const useIsomorphicLayoutEffect = canUseDOM
-  ? React.useLayoutEffect
-  : isReactNative
-    ? React.useLayoutEffect
-    : React.useEffect
+export const useIsomorphicLayoutEffect =
+  canUseDOM || isReactNative ? React.useLayoutEffect : React.useEffect
--- a/src/utils/useIsomorphicLayoutEffect.native.ts
+++ b/src/utils/useIsomorphicLayoutEffect.native.ts
@@ -1,5 +1,0 @@
-import { React } from '../utils/react'
-
-// Under React Native, we know that we always want to use useLayoutEffect
-
-export const useIsomorphicLayoutEffect = React.useLayoutEffect


src/utils/useIsomorphicLayoutEffect.ts
```diff
@@ -16,6 +16,18 @@ export const canUseDOM = !!(
...
+ */
+export const isReactNative =
+  typeof navigator !== 'undefined' && navigator.product === 'ReactNative'
+
 export const useIsomorphicLayoutEffect = canUseDOM
   ? React.useLayoutEffect
-  : React.useEffect
+  : isReactNative
+    ? React.useLayoutEffect
+    : React.useEffect
```
sdg9, line 33: Not sure if there is a convention in this repo but since `canUseDOM` or `isReactNative` want `useLayoutEffect` you could avoid the second ternary in favor of just doing a `||`

```suggestion
export const useIsomorphicLayoutEffect = canUseDOM || isReactNative 
  ? React.useLayoutEffect
  : React.useEffect
```

markerikson:
*sigh* of _course_ it was the `useLayoutEffect` issue, _again_.

src/utils/useIsomorphicLayoutEffect.ts
aryaemami59: Yup, sounds good.

aryaemami59:
@markerikson hopefully this'll be the end of it :), just wondering do you think I should also switch the `target` for the main entry point from `es2020` to `esnext`?

aryaemami59:
@sdg9 Thanks for the repro, it made debugging very easy!

sdg9:
I'm rephrasing https://github.com/reduxjs/react-redux/issues/2150#issuecomment-2041518174 with the fix of this PR in my own words as I want to make sure I understand what is going on.

1. Before react-redux v9 you used a [.native.js suffix](https://github.com/reduxjs/react-redux/commit/b832f835d76ec9d1fcfbdee6c80d497df5dbca06#diff-2cc3fb4125a84f0b15c9eac8264ec8574a6ac595e566ba2c8bf104fbdd3b5ca9R1-R5) to differentiate behavior for ReactNative.  In ReactNative `useIsomorphicLayoutEffect` was explicitly set to `useLayoutEffect`, while the other file had DOM vs server logic.
2. With the react-redux upgrade to v9, the npm artifact was pre-bundled to a single js file.  This change eliminated the use of the `native.js` suffix, which inadvertently caused React Native to be handled the same way as server environments (using `useEffect`) as react-native is not DOM based.  
3. With this PR ReactNative once again uses `useLayoutEffect`.  

Is this an accurate understanding?

Just curious, diving more into # 2, it seems like the impetus of having a conditional driving `useIsomorphicLayoutEffect` is to differentiate server environments vs all others (based on comments in the modified file on how `useLayoutEffect` triggers warnings on servers).  
Would it make more sense to structure the conditional logic to something like the following?
```js
export const useIsomorphicLayoutEffect = isServer ? useEffect : useLayoutEffect
```
  I ask because had it been written this way this issue would never have ocurred when the `native.js` file was dropped, React-Native would have defaulted to the client behavior rather than the server behavior.

I can understand if this is easier said than done, for example I imagine it's easier/more consistent to check if we're running in a DOM vs react-native environment than to check if it's Node.js/server but I'm not certain, hence the question.

markerikson:
Yep, that's a good summary.  (And ironically, that warning is going away in React 19.)

Yeah, tweaking the conditional that way would make sense, I think.

aryaemami59:
@markerikson Do you want me to include further tweaks in this PR or separate?

sdg9:
What about a two pronged approach? Move forward with this PR as-is since we know it resolves the issue. This will unblock RN consumer like myself immediately and really keeps behavior identical to how things were in v8.x
  Then if desired discuss long term whether the conditional makes sense to keep as-is or tweak it based on my question which is more of a forward looking potential preventative measure.

markerikson:
Okay, out in https://github.com/reduxjs/react-redux/releases/tag/v9.1.1 !

PR #2164, opened 2024-04-30
Fix `useRef` usages to be called with an explicit argument of `undefined`.

aryaemami59:
## This PR:

  - [X] Runs `npx types-react-codemod useRef-required-initial .` to fix `useRef` usages to be called with an explicit argument of `undefined` in preparation for React 19.
  
  ### Relevant Links:
  
  - [`useRef` requires an argument](https://react.dev/blog/2024/04/25/react-19-upgrade-guide#useref-requires-argument)
  - [`useRef-required-initial` codemod](https://github.com/eps1lon/types-react-codemod/?tab=readme-ov-file#useref-required-initial)

### DIFF
--- a/src/components/connect.tsx
+++ b/src/components/connect.tsx
@@ -647,7 +647,11 @@ function connect<
       const renderIsScheduled = React.useRef(false)
       const isMounted = React.useRef(false)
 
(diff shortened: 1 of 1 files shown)


timdorr:
Thanks!

PR #2167, opened 2024-05-02
Drop now-unneeded RN peer dep

markerikson:
We listed React Native as a peer dep because we needed to import `unstable_batchedUpdates`.  

We no longer depend on that at all since React batches natively, and we don't have a single `"react-native"` reference in the source. 

This is ultimately the reason why NPM is blowing up with peer dep mismatches, so dropping that peer dep listing _should_ hopefully fix things.

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,14 +52,10 @@
   "peerDependencies": {
     "@types/react": "^18.2.25",
     "react": "^18.0",
-    "react-native": ">=0.69",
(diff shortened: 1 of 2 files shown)


markerikson:
Confirmed on a fresh local Next app that we get the failure with `"react-redux": "^9"` and `npm i @reduxjs/toolkit"`, and that `npm i` works with this PR's build.

PR #2173, opened 2024-05-15
chore: add `size-limit` config

aryaemami59:
## Summary

- Add [**`size-limit`**](https://github.com/ai/size-limit) config


timdorr:
Honestly, I'm not a huge fan of these sorts of tools. They tend to be more annoying than helpful. Maybe we can limit it to just a `react19` branch, so we don't have to tend to it after we complete that work?

aryaemami59:
@timdorr

> Honestly, I'm not a huge fan of these sorts of tools. They tend to be more annoying than helpful.

That's fair, but `size-limit` does give us an idea of how the changes that we make impact our bundle sizes. And with the way I setup the config, it can keep track of every single named export, in every single package entry point, in both development and production builds. And the config is simple enough that I hope we'd be able to insert into any one of the redux packages and it _should_ just work.

> Edit: Wait, don't we already have this? https://github.com/reduxjs/react-redux/blob/master/.github/workflows/size.yaml

While `preactjs/compressed-size-action` does give us some info about bundle size changes, it looks like it's only looking at `dist/cjs/index.js` and `dist/react-redux.legacy-esm.js`. It also doesn't look at named exports, or different environments.

The idea is to have this here so that if we were ever to make a set of changes, we could use it to keep an eye on bundle sizes, it doesn't necessarily have to put a "limit" on anything.

timdorr:
Could this supplant our usage of compressed-size-action? That one hasn't been updated in a long time and causes some warnings in CI that will eventually turn into errors, so it's been on my radar to replace it.

aryaemami59:
I guess we could use https://github.com/EskiMojo14/size-limit-action, same thing we're doing with RTK.

markerikson:
Yeah, I think we want to switch to the other size package for consistency.

aryaemami59:
The other size package?

markerikson:
https://github.com/EskiMojo14/size-limit-action , per above

scottpolo (commented):
LGTM

timdorr:
Note: The action is failing because we don't have size-limit on `master`. Once this lands, that error goes away.

timdorr:
Can we just merge this at this point? I don't need another resolution for 2025 😂

markerikson:
@aryaemami59 same question as with the RTK PR - what does the actual output look like atm?

aryaemami59:
> @aryaemami59 same question as with the RTK PR - what does the actual output look like atm?

It should look something like [**this**](https://github.com/reduxjs/react-redux/actions/runs/28326129722/job/83916340829?pr=2173#step:5:106). 

<details>

<summary><strong>Here is what the output looks like when run locally:</strong></summary>

<br />

react-redux-size-limit-local-output

</details>

PR #2178, opened 2024-06-09
Fix lint error

sharefe-demo:
**Lint Problem**

When you run `yarn lint`, it will throw error:

yarn lint problem

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -23,7 +23,9 @@
   },
   "settings": {
     "import/ignore": ["react-native"],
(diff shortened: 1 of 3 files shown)


timdorr:
Oh crap, it looks like ESLint has been busted for a bit and we haven't been running it in CI. Looking into this in some more detail, but it looks like we were relying on a `pretest` script and support for that was dropped some time ago.

timdorr:
Thank you for looking into this and helping to surface the issue we had in CI. I've got it fixed now on `master`, but thank you for pointing me in the right direction!

PR #2179, opened 2024-06-09
Reduce the exclamation mark and make store swapping simply

sharefe-demo:
**Problem**
Too much exclamation mark expresstions.

**Change**
Keep the exclamation mark away as soon as you can, just Make the code look better.


markerikson:
Hi! Appreciate the PR, but the existing logic works as-is, and we're not planning on making any real changes to the internals of `connect` at this point.

PR #2181, opened 2024-06-11
Fix misleading context + state line

markerikson:
The line that said "store _and store state_ in context" was written in Dec 2018, when we briefly _did_ do that for React-Redux v6.  It hasn't been true since April 2019, when we released v7.


timdorr:
LGTM. Any hold up?

markerikson:
I just forgot to merge this because I was on a plane :)

PR #2225, opened 2025-03-16
chore: update TypeScript version to 5.8

aryaemami59:
## **This PR**:

- [X] Bumps TypeScript to version 5.8.
- [X] Adds TypeScript 5.8 to TS versions to test against during CI.
- [X] Remove TypeScript versions lower than 5.0 from the CI matrix.


EskiMojo14:
noice

PR #2247, opened 2025-07-11
Updated  initialization to avoid using  syntax for better compatibility.

shenzhenjinma:
**Problem:**
React Native 0.80, when running without enabling Hermes or the New Architecture, does not support the `??` (nullish coalescing assignment) or `?.` (optional chaining) operators. This causes compatibility issues and can lead to application crashes in environments where these features are not supported.

**Solution:**
Updated the code to replace `??=` and `?.` operators with more compatible traditional syntax (`if` checks). This ensures the code works seamlessly in React Native 0.80 environments without requiring Hermes or the New Architecture to be enabled.

**Changes:**
- Replaced `??=` in Context.ts with `if` checks for nullish values.
- Replaced `?.` in `connect.tsx` with traditional `if` checks for property existence.

**Impact:**
This change improves compatibility and prevents crashes in React Native 0.80 environments that do not use Hermes or the New Architecture. It ensures the codebase remains stable across different configurations.

**Testing:**
- Verified the updated code works correctly in environments with and without Hermes/New Architecture enabled.
- Ensured no functionality was broken by the changes.

This PR makes the codebase more robust and compatible with a wider range of React Native setups.

### DIFF
--- a/src/components/Context.ts
+++ b/src/components/Context.ts
@@ -28,10 +28,14 @@ const gT: {
 function getContext(): Context<ReactReduxContextValue | null> {
   if (!React.createContext) return {} as any
 
-  const contextMap = (gT[ContextKey] ??= new Map<
-    typeof React.createContext,
-    Context<ReactReduxContextValue | null>
-  >())
+  let contextMap = gT[ContextKey];
+  if (!contextMap) {
+    contextMap = new Map<
+      typeof React.createContext,
+      Context<ReactReduxContextValue | null>
+    >();
+    gT[ContextKey] = contextMap;
+  }
   let realContext = contextMap.get(React.createContext)
   if (!realContext) {
     realContext = React.createContext<ReactReduxContextValue | null>(
--- a/src/components/connect.tsx
+++ b/src/components/connect.tsx
@@ -535,7 +535,7 @@ function connect<
         // Users may optionally pass in a custom context instance to use instead of our ReactReduxContext.
         // Memoize the check that determines which context instance we should use.
         let ResultContext = Context
-        if (propsContext?.Consumer) {
+        if (propsContext && propsContext.Consumer) {
           if (process.env.NODE_ENV !== 'production') {
             const isValid = /*#__PURE__*/ isContextConsumer(
               // @ts-ignore


timdorr:
Yeah, I'm in agreement. This is similar to arguments about supporting older versions of web browsers. If you truly need this stuff handled, you can either transpile the code with something like oxc or swc, or remove it with something like `patch-package`.

shenzhenjinma:
> These features are widely available for about 5 years at this point and were added to the language with ES2020 - tbh., I'm not entirely sure it's a good idea that we push additional bundle size onto all modern consumers because you're running with an engine that's behind on features by half a decade.
> 
> Is there a possibility that you can transpile your dependencies to your required target? `react-redux` is not going to be the only package that uses modern Syntax, so this will become an uphill battle if you don't do that.

Thank you for pointing this out. While it's true that ES2020 features like ??= and ?. have been widely supported for years, React Native (RN) projects often face compatibility challenges due to varying configurations. Specifically, many RN developers do not enable Hermes or the New Architecture, which are required for full ES2020 support. This results in runtime crashes when encountering modern syntax.

To address this, we've carefully reviewed the codebase and identified only two instances where these features were used (??= and ?.). These have been replaced with more compatible alternatives to ensure stability across all RN setups, including those without Hermes enabled.

Transpiling dependencies is indeed a good practice, but in this case, modifying these two occurrences directly avoids introducing additional complexity or bundle size. It ensures compatibility without requiring developers to adjust their build pipelines or configurations.

This change is minimal yet impactful, allowing the project to work seamlessly across all RN environments.

phryneas:
This change adds 90 bytes to every single person consuming the library - React Native or web - but it only serves a very small subset of users that 
* runs React Native without Hermes or new Architecture
* doesn't transpile their dependencies
* keeps dependencies very up-to-date
* for some reason doesn't have a single other dependency using ES2020 syntax

That is a Venn Diagram with a *very* small overlap, while causing a bundle size increase for everyone else - and while this syntax is the one that's important for you, other similar edge cases might come up with different required changes - and suddently those 90 bytes become hundreds of bytes over all of these small additions summed up.

Please, look into either patch-package or transpiling your dependencies.

PR #2248, opened 2025-07-11
Updated  initialization to avoid using  syntax for better compatibility.

shenzhenjinma:


shenzhenjinma:
These features are widely available for about 5 years at this point and were added to the language with ES2020 - tbh., I'm not entirely sure it's a good idea that we push additional bundle size onto all modern consumers because you're running with an engine that's behind on features by half a decade.

Is there a possibility that you can transpile your dependencies to your required target? react-redux is not going to be the only package that uses modern Syntax, so this will become an uphill battle if you don't do that.

Thank you for pointing this out. While it's true that ES2020 features like ??= and ?. have been widely supported for years, React Native (RN) projects often face compatibility challenges due to varying configurations. Specifically, many RN developers do not enable Hermes or the New Architecture, which are required for full ES2020 support. This results in runtime crashes when encountering modern syntax.

To address this, we've carefully reviewed the codebase and identified only two instances where these features were used (??= and ?.). These have been replaced with more compatible alternatives to ensure stability across all RN setups, including those without Hermes enabled.

Transpiling dependencies is indeed a good practice, but in this case, modifying these two occurrences directly avoids introducing additional complexity or bundle size. It ensures compatibility without requiring developers to adjust their build pipelines or configurations.

This change is minimal yet impactful, allowing the project to work seamlessly across all RN environments.

shenzhenjinma:
Thank you for your input! While I understand the argument for supporting modern syntax, the issue here is that the bundled code in the dist folder automatically includes ??= and ?.. This creates compatibility problems for environments where these features are not supported, such as React Native projects without Hermes enabled.

Given that the entire project only has two instances of these modern operators, I believe replacing them with more compatible syntax is a reasonable solution. It avoids the need for additional transpilation steps and ensures broader compatibility without significantly impacting maintainability.

shenzhenjinma:
@phryneas @

PR #2252, opened 2025-08-21
Don't override ref property when not configured to.

### LINKED ISSUE #2251: External property ref is always overridden with value undefined

- React: 19.1.1
- Redux: 5.0.1
- React Redux: 9.2.0

Whe wrapping a component with `connect` without passing the option `forwardRef` set to true, the wrapped component always receives a prop ref with value `undefined` overriding any other prop named `ref`:

```tsx
function WrappedComponent(props) {
   const {ref} = props
   // ref will always be undefined  when wrapped within connect like below.
   return <pan ref={ref} />
}

function HocComponent(props) {
   const [wrapperRef, setRef] = useState(null)
   // The current behavior of connect would override the ref!
   return <WrappedComponent ref={setRef} {...props} />
}
export default connect(...)(HocComponent)
```

A passed property `ref` is passed to the wrapped component, when the option `fowardRef` is not present or falsy.

_No response_

- [ ] Yes

PSpSynedra:
The current behavior of `connect` is to always pass along a ref property, event when the option `forwardRef` was not truthy or preset.

This pull request 
* corrects this error by only forwarding a ref property when actually configured to.
* avoids problems, when accessing the forwarded but always `undefined` ref.
* fixes #2251

The error in the following use case is also fixed with this pull-request:
```
import WrappedComponent from '...'
function HocComponent(props) {
   const [wrapperRef, setRef] = useState(null)
   // The current behavior of connect would override the ref!
   return <WrappedComponent ref={setRef} {...props} />
}
export default connect(...)(HocComponent)
```

_NOTE:_ 
My initial intend was to remove the use of `React.forwardRef(...)` completely, but this would have broken the compatibility with React prior to version 19.

### DIFF
--- a/src/components/connect.tsx
+++ b/src/components/connect.tsx
@@ -747,11 +747,14 @@ function connect<
       // Now that all that's done, we can finally try to actually render the child component.
       // We memoize the elements for the rendered child component as an optimization.
       const renderedWrappedComponent = React.useMemo(() => {
+        const effectiveProps = {...actualChildProps};
+        if (forwardRef) {
+          effectiveProps.ref = reactReduxForwardedRef
+        }
         return (
           // @ts-ignore
           <WrappedComponent
-            {...actualChildProps}
-            ref={reactReduxForwardedRef}
+            {...effectiveProps}
           />
         )
       }, [reactReduxForwardedRef, WrappedComponent, actualChildProps])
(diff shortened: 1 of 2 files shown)


PSpSynedra:
Unfortunately I cannot get the "Test Published Artifact expo (pull_request)" to pass.
When I run `yarn install` with node-22.18.0 as suggested by some internet search results, my local `yarn.lock` does not get changed.

Can I run the failing test locally on my machine to fix the issue?
Anny suggestions on how to fix this?

PR #2263, opened 2025-10-31
[DRAFT] Initial React "concurrent stores" compat prototype

markerikson:
This PR:

- Is a first prototype to see if we can use the WIP React "concurrent stores" API as a replacement for `useSyncExternalStore`
  - Added a Yalc-built version of https://github.com/thejustinwalsh/react-concurrent-store that exposes a couple additional types to satisfy TS
  - Adds a copy of the `reactStoreEnhancer` from the polyfill tests
  - Updates `<Provider>` and `ReactReduxContext` to render the polyfill's `<StoreProvider>` component and the `reactStore` instance in context
  - Rewrites `useSelector` to call the new `useStoreSelector` hook instead of `useSyncExternalStore`, and adds wrapper logic that tries to backfill support for equality checks and unstable selector references (hopefully to be handled by the hook itself later on)
  - Adds a `createTestStore` util that configures the store with the React store enhancer
  - Updates `useSelector.spec.tsx` to use that `createTestStore` util for all of its stores
  - Updates `useSelector.spec.tsx` to _mostly_ pass:
    - commented out now-broken assertions that accessed the React-Redux internal subscription counts, as the current implementation is now relying on the polyfill's `StoreManager` class to manage subscriptions
    - Disabled the "O(1) subscription removal" test, as `StoreManager` is using an array and not a linked list like our `Subscription` class
    - Updated all places where we get different selector call counts, which increased noticeably (1 -> 3, 2 -> 4, 4 -> 8, 3 -> 5, etc).  Clearly the new implementation is calling selectors a lot more.

## Background

@captbaritone and @thejustinwalsh have been working in https://github.com/thejustinwalsh/react-concurrent-store to prototype the upcoming React "concurrent stores" API.  As a loose description, what we're hoping for is "`useSyncExternalStore`, minus the `Sync`" - aka a concurrent/transition-compatible way to integrate external state into React.

They've got a first POC polyfill up on NPM as `react-concurrent-store`, and the repo has a couple tiny demo integrations for Relay and Redux.

This is my attempt to do an experimental integration and rewrite React-Redux to use it and see how far we get, a la the alpha integration of `useSyncExternalStore` in #1808 .

## Results

Impressively, especially for just a couple hours of hacking:

**All but 3 `useSelector` tests pass!**

The caveat is that I did change pretty much all places where we assert the number of selector calls (which all went up 2+ calls), and disabled a couple places where we asserted that we had N active internal subscriptions (as this mechanism bypasses our own Subscription implementation).

The test failures are:

- A test that asserts we always use the latest selector
- 2 tests that expect we handle errors thrown from selectors in a specific way

```
 FAIL  test/hooks/useSelector.spec.tsx > React > hooks > useSelector > uses the latest selector
AssertionError: expected [ +0, +0 ] to deeply equal [ +0, 1 ]

- Expected
+ Received

  Array [
    0,
-   1,
+   0,
  ]

 ❯ test/hooks/useSelector.spec.tsx:481:31
    479|           forceRender()
    480|         })
    481|         expect(renderedItems).toEqual([0, 1])
       |                               ^
    482|
    483|         rtl.act(() => {

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯[1/3]⎯

 FAIL  test/hooks/useSelector.spec.tsx > React > hooks > useSelector > edge cases > ignores transient errors in selector (e.g. due to stale props)
AssertionError: expected [Function doDispatch] to not throw an error but 'Error' was thrown

- Expected:
undefined

+ Received:
"Error"

 ❯ test/hooks/useSelector.spec.tsx:528:34
    526|             })
    527|           }
    528|           expect(doDispatch).not.toThrowError()
       |                                  ^
    529|
    530|           spy.mockRestore()

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯[2/3]⎯

 FAIL  test/hooks/useSelector.spec.tsx > React > hooks > useSelector > edge cases > re-throws errors from the selector that only occur during rendering
AssertionError: expected [Function] to throw an error
 ❯ test/hooks/useSelector.spec.tsx:602:14
    600|               normalStore.dispatch({ type: '' })
    601|             })
    602|           }).toThrowError()
       |              ^
    603|
    604|           spy.mockRestore()
```

## Analysis

I believe our own v7 `useSelector` implementation would catch errors, swallow them, and force a re-render (under the assumption that if you had a stale props / "zombie child" scenario, then the problems would all go away by the time the component was done re-rendering).  I don't know what the current `uSES` semantics are, but here `useStoreSelector` is clearly doing something different.

The selector stability test is obviously failing due to some combo of the actual `useStoreSelector` implementation not actually supporting unstable selector references and the attempt to make them work in userland here.

Overall, though, this is a pretty solid indication that we're on the right track already.  Those are all pretty edge-case-y things, and the core functionality is working.

### `connect`?

This _does_ bring up a big question about if and how we'd be able to convert `connect` over.  Personally I'd rather _not_ :)  `connect` has always been a big ugly beast of an implementation, and as it is I'm pretty sure the hacks I've got in place right now are an abuse of `uSES`'s semantics.  I don't even want to think about making that work with `useStoreSelector`.  I'd be perfectly fine saying "NOPE, `connect` works as-is, with all its limitations, No Concurrent Behavior For You, go migrate to `useSelector` if you want that".

### Store Enhancers?

You'll note this POC requires adding a Redux store enhancer to let us capture the action create the "React-compatible store":

```ts
export const addReactStore: StoreEnhancer<{
  reactStore: ReactStore
}> = (createStore) => {
  return (reducer, preloadedState) => {
    const store = createStore(reducer, preloadedState)

    // Create concurrent-safe store wrapper
    const reactStore = createStoreFromSource({
      getState: store.getState,
      reducer: reducer,
    })

    // Intercept dispatch to notify reactStore
    const originalDispatch = store.dispatch
    store.dispatch = (action: any) => {
      const result = originalDispatch(action)
      reactStore.handleUpdate(action)
      return result
    }

    // Attach reactStore to Redux store
    return Object.assign(store, { reactStore })
  }
}
```

It's hypothetically possible we might be able to find a different solution as we go, but as a first attempt this seems like the right integration point to me.  This is something that would presumably be added _after_ the middleware enhancer (and before the DevTools?), so that it only sees actions that are about to reach the reducer.  (Actually, now that I say that... I don't know what happens when you try time-traveling in the DevTools, so that's something we should investigate.)

### Subscriptions?

I realized this bypasses our own `Subscription` class entirely.  That class has historically served 3 purposes:

- It enforced React batching pre-React-18
- It allowed us to do cascading updates in the commit phase, which was necessary to enforce `connect`'s invariants that "`mapState` always sees the latest props from its parent"
- By using a linked list for subscribers instead of an array, we get O(1) performance for removing a subscriber ( #1523 , #1870 ).  The polyfill's `StoreManager` class is using an array, so it presumably has that same perf problem.  (I looked at `uSES`, and technically [I don't think it actually tracks any subscribers itself](https://github.com/facebook/react/blob/488d88b018ee8fd1fac56cab22dfa8796ebce30b/packages/react-reconciler/src/ReactFiberHooks.js#L1874)? 

Some of the tests failed because they asserted there were N active subscriptions, and now we don't have any subscriptions ourselves at all.

I'm not sure what happens if you were to mix-and-match at this point.  I assume that a `connect` in the tree would still get triggered from `Subscription`. 

To some extent `StoreManager` _is_ an alternative to `Subscription`.  Probably ought to use a linked list there if it's going to track subscribers, though.

### `useStoreSelector` Changes

Per https://github.com/thejustinwalsh/react-concurrent-store/issues/13 we really need this to allow unstable selector references, because this is a standard ecosystem pattern: 

```ts
const todo = useSelector(state => selectTodoById(state, id))
```

We also need to handle customizable equality checks.  Currently we rely on `useSyncExternalStoreWithSelector` supporting that:

- [`useSyncExternalStoreWithSelector.js`: `isEqual`](https://github.com/facebook/react/blob/488d88b018ee8fd1fac56cab22dfa8796ebce30b/packages/use-sync-external-store/src/useSyncExternalStoreWithSelector.js#L24)

`useStoreSelector` will probably need to do something similar.

There's also the error-handling semantics question.

### Testing Scenarios

Beyond that, though, we'd want to start adding some tests with transition behaviors and see what happens.

We'd also want to do some testing with Redux middleware and the Redux DevTools and see what happens and what expectations might break - certainly in normal sync mode, then with transitions added in.


einarq:
Great to see this progressing. 
Re the eventual migration story, are you hoping to allow some type of gradual migration? Or will it most likely have to be all at once?
If the latter I guess we should start converting all “connect” usages to useSelector.

markerikson:
@einarq Not sure yet. My assumption is this would end up being a new major version of React-Redux (v10), similar to how we shipped v8 with the switch from our own `useSelector` implementation to `useSyncExternalStore` (polyfill), and then v9 was the switch from the polyfill to the built-in `uSES`.

That in turn would also require waiting for a new version of React containing the method and a bump to match that version of React as the required minimum (unless there's a similar polyfill).

I wouldn't want to have flags to flip behavior, or add a new hook that's `useSelectorConcurrentCompat`, or anything like that.

And no, I am _not_ spending time on `connect` for this :)  Heh. As bad as the recent Star Wars movies were, the phrase "Let the past die. Kill it if you have to." comes to mind :)

thejustinwalsh:
Hey, thanks for putting this together! Reviewing now to see if there is anything further I can push on to unblock on the concurrent store side.

PR #2269, opened 2025-12-24
Mark `connect` as deprecated

### LINKED ISSUE #2264: Mark `connect` as deprecated
### What is the new or updated feature that you are suggesting?

We're doing early collaboration with the React team on the design and behavior of the upcoming "concurrent stores" API.  Loosely put, it's meant to be a concurrent compatible replacement for `useSyncExternalStore`.

This would likely ship as React-Redux v10, and also require a newer minimum version of React (unless they ship a polyfill for backwards compat with older versions).

Per discussion in  https://github.com/reduxjs/react-redux/pull/2263#issuecomment-3566978744 , I don't intend to spend any time upgrading `connect` to use the new API, which means that it would still be using `useSyncExternalStore` and have its limitations with concurrent behavior like transitions.

We've said for years that `useSelector` ought to be the default, and that while we're not _removing_ `connect`, we also recommend against using it.

Given that, we ought to do the [same thing we did for `createStore`](https://github.com/reduxjs/redux/releases/tag/v4.2.0), and officially mark `connect` as `@deprecated`.  As with `createStore`, we still have no plans to remove `connect`, but we really don't think anyone should be writing new code that uses it directly.

As with `createStore`, I'd probably ship an aliased version that doesn't have the deprecation, for folks who _reallly_ don't want those deprecated strikethroughs in their codebase (or get flagged by enterprise linting requirements).

Once again, to be very clear: **I do not intend to remove `connect` any time in the near future!**  It's possible we might remove it _someday_, but I have no plans to actually do so, and _if_ we ever do it'd be a few years / major versions down the road.

Given that `connect` has a bigger bundle size, much more complex TS usage, worse re-rendering performance due to the tiered update cascading in the commit phase, and presumably won't have upcoming concurrent compat like `useSelector`, there's no good reason for people to still use it.  Even the auto-wrapping-with-`React.memo` behavior is irrelevant now that the React Compiler exists.

I'll probably ship this deprecation in a new minor release in the next few weeks when I have time to do the work, and update the docs to cover this.  (I might even rip out the ancient `connect`-based tutorial page while I'm at it.)

### Why should this feature be included?

Because `connect` is obsolete :)

### What docs changes are needed to explain this?

Release notes, removing some `connect` references, etc.

markerikson:
- Restructured the `connect` export to preserve the variable name
- Marked `connect` as `@deprecated`
- Added a `legacy_connect` re-export without the deprecation
- Fixed a hydration test assertion for an error message that changed in React 19.2

Fixes #2264


aryaemami59 (approved):

markerikson:
_cracks knuckles_ awright it's past time!

timdorr:
~Mark connect as deprecated~

Mark, connect _has_ deprecated

PR #2306, opened 2026-05-20
fix(`connect`): log `mapStateToProps` errors to `console.error` in dev (#1942)

### LINKED ISSUE #1942: connect swallows errors in mapStateToProps

- React: 18.2
- ReactDOM/React Native: react-dom 18.2
- Redux: 4.2.0
- React Redux: 8.0.2

https://codesandbox.io/s/gifted-architecture-wr4f86?file=/src/App.js

1. Have a code that can cause an error in mapStateToProps, e.g.
```js
  return {
    foo: state.foo.length, // when foo is undefined, causes an error
  };
};
```
2. Dispatch an action that changes the state so that it causes the error above
```js
dispatch(updateFoo(undefined));
```
3. Rerender doesn't happen, no errors emitted in console, componentDidCatch catches nothing. The error is swallowed.

This issue makes debugging code that's using react-redux really hard.
The expected behaviour is that the error is visible in console.

Windows, latest Chrome here. Probably affects all browsers.

- [ ] Yes

webdevelopersrinu:
## Summary

Logs errors thrown in `mapStateToProps` (or selectors) to `console.error`
in non-production environments, so they surface during development
instead of being silently swallowed.

Fixes #1942

## Background

Since v8.0.0, errors thrown inside `mapStateToProps` are caught in the
subscription callback ([connect.tsx#L133-L136][1]) and stored in
`lastThrownError`. In v7, those errors surfaced because the component
unmounted on selector errors (the unmount path re-throws
`lastThrownError`). In v8+, the component no longer unmounts — it just
re-renders in its pre-error state — so the error is silently swallowed.

This was diagnosed by @alexanderchr in
[#1942 (Dec 12, 2024)][2], and @markerikson approved the dev-only
logging approach in the [same thread][3]:

> "logging this in dev seems like a passable fallback"

## Changes

- `src/components/connect.tsx` — log the caught error to `console.error`
  when `process.env.NODE_ENV !== 'production'`, with a brief message and
  a link back to this issue
- `test/components/connect.spec.tsx` — add a unit test that dispatches
  an action which causes `mapStateToProps` to throw, then asserts the
  error was logged via `console.error`

## Behavior

- ✅ **Production**: unchanged (the new branch is dead-code-eliminated)
- ✅ **Development**: error is now logged to the console with the
  original error object, making it discoverable in DevTools

## Test plan

- [x] `yarn test` — all `connect` tests pass (65/65), new test passes,
  no regressions in code I touched
- [x] `yarn lint` runs as part of `pretest` and passes
- [x] Type-check passes (`vitest --typecheck` reports no errors)

[1]: https://github.com/reduxjs/react-redux/blob/master/src/components/connect.tsx#L133-L136
[2]: https://github.com/reduxjs/react-redux/issues/1942#issuecomment-2540022906
[3]: https://github.com/reduxjs/react-redux/issues/1942#issuecomment-2540097832

### DIFF
--- a/src/components/connect.tsx
+++ b/src/components/connect.tsx
@@ -103,6 +103,7 @@ function subscribeUpdates(
   notifyNestedSubs: () => void,
   // forceComponentUpdateDispatch: React.Dispatch<any>,
   additionalSubscribeListener: () => void,
+  displayName: string,
 ) {
   // If we're not subscribed to the store, nothing to do here
   if (!shouldHandleStateChanges) return () => {}
@@ -135,7 +136,7 @@ function subscribeUpdates(
       lastThrownError = e as Error | null
       if (process.env.NODE_ENV !== 'production') {
         console.error(
-          'An error occurred in `mapStateToProps` (or a selector) of a `connect()`-ed component. ' +
+          `An error occurred in \`mapStateToProps\` (or a selector) of the \`connect()\`-ed component "${displayName}". ` +
             'See https://github.com/reduxjs/react-redux/issues/1942 for details.',
           e,
         )
@@ -719,6 +720,7 @@ function _connect<
             childPropsFromStoreUpdate,
             notifyNestedSubs,
             reactListener,
+            displayName,
           )
         }
 
--- a/src/components/connect.tsx
+++ b/src/components/connect.tsx
@@ -133,6 +133,13 @@ function subscribeUpdates(
     } catch (e) {
       error = e
       lastThrownError = e as Error | null
+      if (process.env.NODE_ENV !== 'production') {
+        console.error(
+          'An error occurred in `mapStateToProps` (or a selector) of a `connect()`-ed component. ' +
+            'See https://github.com/reduxjs/react-redux/issues/1942 for details.',
+          e,
+        )
+      }
     }
 
     if (!error) {
--- a/test/components/connect.spec.tsx
+++ b/test/components/connect.spec.tsx
@@ -287,7 +287,8 @@ describe('React', () => {
         })
 
         consoleSpy.mockRestore()
-        process.env.NODE_ENV = originalNodeEnv
+
+        vi.unstubAllEnvs()
       })
 
       it("should retain the store's context", () => {
--- a/test/components/connect.spec.tsx
+++ b/test/components/connect.spec.tsx
@@ -221,6 +221,12 @@ describe('React', () => {
       })
 
       it('should log errors thrown in mapStateToProps to console.error in development (#1942)', () => {
+        // The dev log is gated behind `process.env.NODE_ENV !== 'production'`.
+        // Pin it to 'development' so the test exercises the dev branch explicitly
+        // rather than relying on the test runner's ambient env.
+        const originalNodeEnv = process.env.NODE_ENV
+        process.env.NODE_ENV = 'development'
+
         const store: Store = createStore(stringBuilder)
 
         // Throw on the first call only. The subscription-update path catches
@@ -268,6 +274,16 @@ describe('React', () => {
         )
         expect(loggedAnError).toBe(true)
 
+        // The logged message should name the offending component so the
+        // developer can tell which `connect()`-ed component threw.
+        const namedTheComponent = newCalls.some((args) =>
+          args.some(
+            (arg) =>
+              typeof arg === 'string' && arg.includes('Connect(Container)'),
+          ),
+        )
+        expect(namedTheComponent).toBe(true)
+
         // Dispatch again so the subscription path runs a successful update,
         // which clears `lastThrownError` and prevents the unmount-time rethrow.
         rtl.act(() => {
@@ -275,6 +291,7 @@ describe('React', () => {
         })
 
         consoleSpy.mockRestore()
+        process.env.NODE_ENV = originalNodeEnv
       })
 
       it("should retain the store's context", () => {
--- a/test/components/connect.spec.tsx
+++ b/test/components/connect.spec.tsx
@@ -220,6 +220,63 @@ describe('React', () => {
         expect(tester.getByTestId('string')).toHaveTextContent('ab')
       })
 
+      it('should log errors thrown in mapStateToProps to console.error in development (#1942)', () => {
+        const store: Store = createStore(stringBuilder)
+
+        // Throw on the first call only. The subscription-update path catches
+        // and logs the error; the subsequent re-render then succeeds, so the
+        // error does not propagate up to the test runner.
+        let shouldThrow = false
+        const mapState = (state: string) => {
+          if (shouldThrow) {
+            shouldThrow = false
+            throw new Error('mapStateToProps failed')
+          }
+          return { value: state }
+        }
+
+        class Container extends Component<{ value: string }> {
+          render() {
+            return <Passthrough {...this.props} />
+          }
+        }
+        const ConnectedContainer = connect(mapState)(Container)
+
+        const consoleSpy = vi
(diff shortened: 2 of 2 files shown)


aryaemami59:
The current CI failure is related to [**arethetypeswrong/arethetypeswrong.github.io#258**](https://github.com/arethetypeswrong/arethetypeswrong.github.io/issues/258) which is caused by [**101arrowz/fflate#258**](https://github.com/101arrowz/fflate/issues/258).

As a temporary workaround, we can pin **`fflate`** to **`0.8.2`** via **`resolutions`** in **`package.json`**:

```json
  "resolutions": {
    "fflate": "0.8.2"
  },
```

This should unblock CI until the upstream issue is resolved.

webdevelopersrinu:
Thanks @aryaemami59 — totally missed your earlier comment, sorry about
that. Happy to add the `fflate` resolutions pin to this PR if you'd
prefer, or wait if you're patching master directly. Just let me know.

aryaemami59:
@webdevelopersrinu

> Thanks @aryaemami59 — totally missed your earlier comment, sorry about that. Happy to add the `fflate` resolutions pin to this PR if you'd prefer, or wait if you're patching master directly. Just let me know.

Thanks. I'd prefer to wait and see whether this gets resolved upstream first. That also gives the rest of the team more time to review the PR.

**Update:** This has since been fixed upstream in [**`@arethetypeswrong/core@0.18.3`**](https://github.com/arethetypeswrong/arethetypeswrong.github.io/releases/tag/%40arethetypeswrong%2Fcore%400.18.3), so the **`fflate`** resolutions pin isn't needed here. I've opened **#2307** to bump **`@arethetypeswrong/cli`** to **`0.18.3`** and unblock CI.

src/components/connect.tsx
aryaemami59, line 105: Add a `displayName` param to `subscribeUpdates`:

```suggestion
  additionalSubscribeListener: () => void,
  displayName: string,
```

src/components/connect.tsx
```diff
@@ -133,6 +133,13 @@ function subscribeUpdates(
     } catch (e) {
       error = e
       lastThrownError = e as Error | null
+      if (process.env.NODE_ENV !== 'production') {
+        console.error(
+          'An error occurred in `mapStateToProps` (or a selector) of a `connect()`-ed component. ' +
```
aryaemami59: Name the offending component in the message:

```suggestion
          `An error occurred in \`mapStateToProps\` (or a selector) of ${displayName}. ` +
```

src/components/connect.tsx
aryaemami59, line 722: Pass `displayName` at the call site (it's already in scope here):

```suggestion
            reactListener,
            displayName,
```

test/components/connect.spec.tsx
```diff
@@ -220,6 +220,63 @@ describe('React', () => {
         expect(tester.getByTestId('string')).toHaveTextContent('ab')
       })
 
+      it('should log errors thrown in mapStateToProps to console.error in development (#1942)', () => {
```
aryaemami59, line 223: Explicitly set `NODE_ENV` to `'development'` so the test pins the dev branch.

```suggestion
      it('should log errors thrown in mapStateToProps to console.error in development (#1942)', () => {
        vi.stubEnv('NODE_ENV', 'development')

```

test/components/connect.spec.tsx
```diff
@@ -220,6 +220,63 @@ describe('React', () => {
...
+        })
+
+        const newCalls = consoleSpy.mock.calls.slice(callsBeforeDispatch)
+        const loggedAnError = newCalls.some((args) =>
+          args.some(
+            (arg) =>
+              arg instanceof Error && arg.message === 'mapStateToProps failed',
+          ),
+        )
+        expect(loggedAnError).toBe(true)
```
aryaemami59, line 271: Assert the logged message names the offending component.

```suggestion
        expect(loggedAnError).toBe(true)

        // The message should name the offending connected component so the
        // developer knows where to look.
        const loggedComponentName = newCalls.some((args) =>
          args.some(
            (arg) =>
              typeof arg === 'string' && arg.includes('Connect(Container)'),
          ),
        )

        expect(loggedComponentName).toBe(true)
```

aryaemami59 (commented):
Two small follow-ups on top of the dev-logging change:

1. **Name the component that threw** — the message currently says "a `connect()`-ed component", which isn't actionable. `displayName` (`Connect(X)`) is already computed in `_connect` and in scope at the call site, so threading it through is cheap and matches the existing `mapStateToProps() in Connect(X) …` style.
2. **Make the test pin `NODE_ENV`** — it's titled "…in development" but never sets the env; it passes only because the runner isn't `production`.

webdevelopersrinu:
@aryaemami59 thanks for the review! Addressed all the feedback in 048188c:

- Threaded `displayName` through `subscribeUpdates` and included it in the
  dev `console.error` message, so it now names the offending component
  (e.g. `"Connect(Container)"`). Used the double-quoted style to match the
  existing "Could not find store" error in this file.
- Updated the test to explicitly set `NODE_ENV='development'` (and restore
  it), and to assert the logged message names the component.

Let me know if you'd like anything else.

test/components/connect.spec.tsx
```diff
@@ -220,6 +220,80 @@ describe('React', () => {
         expect(tester.getByTestId('string')).toHaveTextContent('ab')
       })
 
+      it('should log errors thrown in mapStateToProps to console.error in development (#1942)', () => {
+        // The dev log is gated behind `process.env.NODE_ENV !== 'production'`.
+        // Pin it to 'development' so the test exercises the dev branch explicitly
+        // rather than relying on the test runner's ambient env.
+        const originalNodeEnv = process.env.NODE_ENV
+        process.env.NODE_ENV = 'development'
```
aryaemami59: I'd much prefer we use Vitest's [`vi.stubEnv()`](https://vitest.dev/api/vi.html#vi-stubenv)
here instead of mutating `process.env.NODE_ENV` directly. Two reasons:

- **It can't leak.** The manual save/restore only restores `NODE_ENV` if execution actually
  reaches the cleanup line at the bottom of the test. If any assertion before it throws,
  `process.env.NODE_ENV` stays `'development'` and bleeds into every subsequent test in this
  file. `vi.stubEnv()` records the original value for us.
- **Less bookkeeping.** No `originalNodeEnv` variable to thread through the test.

```suggestion
        vi.stubEnv('NODE_ENV', 'development')
```

Then the `process.env.NODE_ENV = originalNodeEnv` line at the end becomes a single [`vi.unstubAllEnvs()`](https://vitest.dev/api/vi.html#vi-unstuballenvs) (per my other comment).

test/components/connect.spec.tsx
```diff
@@ -220,6 +220,80 @@ describe('React', () => {
...
+        expect(namedTheComponent).toBe(true)
+
+        // Dispatch again so the subscription path runs a successful update,
+        // which clears `lastThrownError` and prevents the unmount-time rethrow.
+        rtl.act(() => {
+          store.dispatch({ type: 'APPEND', body: 'b' })
+        })
+
+        consoleSpy.mockRestore()
+        process.env.NODE_ENV = originalNodeEnv
```
aryaemami59: Restore the stubbed env at the end of the test:

```suggestion

        vi.unstubAllEnvs()
```

aryaemami59 (changes requested):

aryaemami59 (approved):

webdevelopersrinu:
Thanks for the review @aryaemami59

timdorr:
Thanks!

PR #2310, opened 2026-06-07
feat: add dev stability checks to connect

louzhedong:
Adds an opt-in `stabilityCheck` option to `connect` and warns in development when `mapStateToProps` returns a different result for identical inputs.

Verification:
- `node_modules\.bin\vitest.cmd --run test/components/connect.spec.tsx`
- `node_modules\.bin\tsc.cmd --noEmit -p tsconfig.test.json`

### DIFF
--- a/src/components/connect.tsx
+++ b/src/components/connect.tsx
@@ -38,6 +38,7 @@ import type {
   ReactReduxContextInstance,
 } from './Context'
 import { ReactReduxContext } from './Context'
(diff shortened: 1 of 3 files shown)


louzhedong:
I’ve kept the branch up to date and verified the change locally with the repo’s test suite and typecheck. If there are any concerns about the stabilityCheck behavior or naming, I can adjust quickly.

PR #2311, opened 2026-06-07
fix: surface `mapStateToProps` errors in `connect`

louzhedong:
This makes `connect` rethrow subscription callback errors during render so `mapStateToProps` failures can reach an error boundary instead of being swallowed.

Validation:
- ='C:\Users\Administrator\AppData\Local\npm-cache\_npx\ebaba8b9e55fd0a9\node_modules\node\bin\node.exe'; &  'C:\Program Files\nodejs\node_modules\corepack\dist\yarn.js' vitest test/components/connect.spec.tsx --run

### DIFF
--- a/src/components/connect.tsx
+++ b/src/components/connect.tsx
@@ -100,6 +100,7 @@ function subscribeUpdates(
   renderIsScheduled: React.MutableRefObject<boolean>,
   isMounted: React.MutableRefObject<boolean>,
   childPropsFromStoreUpdate: React.MutableRefObject<unknown>,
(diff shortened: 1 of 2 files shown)


louzhedong:
Quick follow-up in case this is easier to review now than the other `connect` PR:

This change is intentionally narrow. It only re-surfaces `mapStateToProps` subscription errors during render so existing error boundaries can catch them, without changing the successful update path.

If you'd prefer this handled in a different layer of `connect`, or want the regression coverage trimmed to a smaller repro, I'm happy to adjust.

PR #2340, opened 2026-07-20
fix(`connect`): split `mapStateToProps`/`mapDispatchToProps` unions into ordered overloads (#2244)

### LINKED ISSUE #2244: Typescript 7 won’t infer container types correctly

- React: latest
- ReactDOM/React Native: na
- Redux: latest
- React Redux: latest

See bug report here:
https://github.com/microsoft/typescript-go/issues/977#issuecomment-2981187745

The types now fail with typescript go due to different type ordering.

The type ordering will change with typescript go and there afaik is not going to be any way round that.

In that comment Anders suggested a fix:

> Ideally, the React declaration files would break the union parameter into separate overloads and list them in order of preference.

Na 

- [ ] Yes

veksa:
Fixes #2244.

`connect`'s `mapStateToProps` and `mapDispatchToProps` are typed as unions that include both a "factory" form (`() => (state) => props`) and a plain form. A factory is structurally assignable to its plain counterpart, so the compiler has to pick a "first" candidate to infer `TStateProps` / `TDispatchProps` from — and `tsc` and the new native compiler (`tsgo` / TS 7) order union members differently. One of them ends up inferring the wrong candidate and the connected component's props collapse to `{}` (a required prop shows up as "missing" at the call site).

As Anders suggested in the upstream analysis (microsoft/typescript-go#977), the robust fix is to break the union into separate overloads listed in order of preference rather than relying on union ordering. So I've split the affected overloads with the factory form listed first, followed by the plain form. A plain object-returning function isn't assignable to the factory signature, so a factory matches the first overload and everything else falls through — deterministically, and independent of the compiler.

A couple of notes on scope:

- The object form of `mapDispatchToProps` doesn't get its own factory overload: factories and dispatch functions are already matched by the function-form overloads that precede it, so the object overloads only need the plain (`MapDispatchToProps`) parameter. This keeps the overload count from exploding.
- The `mergeProps` overloads deliberately keep the union parameters. There `mergeProps` receives `stateProps` / `dispatchProps` as contextually-typed (non-inferring) parameters, so those types can only come from the map functions — and a leading factory overload would swallow a plain map function and collapse the inferred props to `{}`. Keeping both forms as candidates in a single overload preserves the original behavior. It's documented inline.

Types only, no runtime changes.

I added type tests covering the repro cases from the issue (plain/factory `mapStateToProps`, the function/factory/object forms of `mapDispatchToProps`, a factory + factory combination, and the wrong-return-type cases). They pass under both `tsc` and the native compiler, and — this is the important part — they stay green even if the union members are reordered, which is the actual failure mode here: reordering the unions on plain `tsc` reproduces the exact errors from the report before this change, and no longer does after it.

One caveat worth mentioning: current `tsgo` dev builds may already infer these cases correctly, but the union ordering is still load-bearing and fragile, and the compiler-side issue was closed as "not planned" — so this makes the types stable regardless of compiler or union order.

### DIFF
--- a/src/components/connect.tsx
+++ b/src/components/connect.tsx
@@ -15,10 +15,14 @@ import type {
 } from '../types'
 
 import type {
+  MapStateToProps,
+  MapStateToPropsFactory,
   MapStateToPropsParam,
+  MapDispatchToProps,
+  MapDispatchToPropsFactory,
+  MapDispatchToPropsFunction,
   MapDispatchToPropsParam,
   MergeProps,
-  MapDispatchToPropsNonObject,
   SelectorFactoryOptions,
 } from '../connect/selectorFactory'
 import defaultSelectorFactory from '../connect/selectorFactory'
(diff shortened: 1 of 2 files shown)


markerikson:
Ran into this while testing out TS7 readiness.  Thanks!

PR #2355, opened 2026-07-30
lazy deep tracking via a coarse segment tier (mount ≈ stock)

veksa:
Builds on #2354 (defer tracking off the mount render path) and takes it further. Adds a coarse top-level "segment" tier in front of the deep path tracking, so the per-selector proxy/signal graph is built **lazily** — only for hooks whose top-level slice actually changed — instead of eagerly for every hook at mount.

Net effect: the signals render/dispatch wins are kept, and the one remaining downside — modestly slower mount — goes away, with mount back at stock levels.

## How it works

- At mount a selector records only which top-level state keys it reads (one shallow proxy, cheap) — no deep graph is built.
- On dispatch the changed top-level segments are computed (reference compare, O(number of top-level keys)); only hooks reading a changed segment build their deep graph, on first touch. After that they're driven by the existing deep tracking exactly as before, so steady-state pruning is unchanged.
- The first-touch burst is time-sliced so it never blocks a frame (tunable via `sliceThreshold`).
- No tearing: an unbuilt hook's `getSnapshot` recomputes from the latest committed state, so it can't show a stale value against synchronously-updated siblings (test included).

## Scenario table (medians of 5 — stock → signals → this PR)

**Renders** (the "total work" metric):

| scenario | stock | signals | this PR |
| --- | --: | --: | --: |
| derived-selectors | 506 | 38 | 40 |
| price-ticker | 966 | 1106 | 1133 |
| entity-list / selective-update / deeptree | — | ≈ | ≈ |

**Dispatch avg (ms):**

| scenario | stock | signals | this PR |
| --- | --: | --: | --: |
| derived-selectors | 0.82 | 0.28 | 0.25 |
| entity-list-array | 0.95 | 0.45 | 0.42 |
| multi-selector-component | 0.83 | 0.47 | 0.34 |
| price-ticker | 5.96 | 2.48 | 2.48 |

**Mount (ms)** — the downside, now gone:

| scenario | stock | signals | this PR |
| --- | --: | --: | --: |
| derived-selectors | 5.9 | 29.0 | 6.2 |
| price-ticker | 27.8 | 158 | 29.5 |
| entity-list-array | 45 | 62 | 47 |

Renders and dispatch stay at signals levels; mount comes back to stock.

## At scale — price-ticker, 200k rows

| build | mount | dispatch avg |
| --- | --: | --: |
| stock 9.2 | 454 | 93.6 |
| signals (this PR's base) | 5905 | 12.9 |
| this PR | 553 | 23.9 |

Mount is ~10× better than signals; dispatch stays far under stock. Dispatch sits a bit above signals here only because signals front-loads all its work at mount (the ~5.9s) while this PR spreads it across the first seconds of updates — net it saves ~5.3s at load.

## Tradeoffs

- `sliceThreshold` (default 64): above it, a large first-touch burst is time-sliced (responsive, but those updates stagger during the burst); `Infinity` runs it inline (atomic, can block on a huge first touch). Steady-state updates are always synchronous, so this only affects the first touch of a slice.
- At extreme scale the warmup spreads the build work across the first seconds of dispatches; once warm, it equals signals.

### DIFF
--- a/src/signals/SignalProvider.tsx
+++ b/src/signals/SignalProvider.tsx
@@ -13,18 +13,30 @@ import { changedSegments } from './coarseSegments'
 import type { CoarseSub } from './coarseSegments'
 import { scheduleCallback, shouldYield } from './scheduler'
 
-// Above this many candidates in a single dispatch, the coarse pass (each
-// candidate builds its deferred deep graph + notifies) is time-sliced so it
-// never blocks one frame; below it, running inline avoids scheduler overhead.
-const COARSE_SLICE_THRESHOLD = 64
 import { alienEngine } from './engine'
 import type { SignalEngine } from './types'
+
+// Default for SignalProvider's `sliceThreshold`: above this many candidates in
+// a single dispatch, the coarse first-touch pass is time-sliced (responsive,
+// but updates stagger); below it, it runs inline (atomic — all candidates in
+// one synchronous pass). Override via the prop.
+const DEFAULT_SLICE_THRESHOLD = 64
 
 export interface SignalProviderProps<
   A extends Action<string> = UnknownAction,
   S = unknown,
 > extends ProviderProps<A, S> {
   engine?: SignalEngine
+  /**
+   * Candidate count above which the coarse first-touch build+notify burst is
+   * time-sliced across macrotasks (keeps a huge burst from blocking a frame,
+   * at the cost of updates staggering during that burst). Below it the pass
+   * runs inline in one synchronous, atomic step. Default `64`. Set to
+   * `Infinity` to always run inline (atomic, may block on a large first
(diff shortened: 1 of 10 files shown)


markerikson:
The slicing idea is neat, but it's not something I want to pursue due to inconsistency concerns as well as not actually saving on total work.

_This_ idea is excellent :)  I'm probably going to close this and implement a modified version, but thanks for coming up with this suggestion!  Not something that had crossed my mind for the signals effort yet, and ironically I'd even seen a previous variation of that as a _general_ solution and then played with it myself here ( https://github.com/reduxjs/react-redux/pull/2246 ).  so I probably _should_ have thought of something like this myself :)  Thanks!

markerikson:
Integrated into the signals PR, thanks!

PR #2356, opened 2026-07-30
chore: update caniuse-lite and bump yarn to 4.18.0

veksa:
Fixes the `Browserslist: caniuse-lite is outdated` warning in CI.

- `caniuse-lite` → `1.0.30001806` in `website/` and `examples/publish-ci/rr-rsc-context/`.
- `BROWSERSLIST_IGNORE_OLD_DATA` on the `Build example` step: `cra4` and `next` come from
  the RTK repo pinned to a commit, their data is frozen and can't be updated from here.
- Yarn `4.4.1` → `4.18.0` along the way.


veksa:
Follow-up: CI was failing because Yarn merges every `.yarnrc.yml` up the tree —
the RTK clone lives inside our checkout and pins Yarn 4.4.1, so the old Yarn was
the one running and choked on `npmMinimalAgeGate`.

Moved the clone to `$RUNNER_TEMP`, dropped that setting (a no-op here anyway),
and swapped the deprecated `--frozen-lockfile` for `--immutable`.

timdorr:
Looks good. Thanks!

PR #2358, opened 2026-07-31
docs: update `Provider` props to match the current types

dfedoryshchev:
The `<Provider>` props reference on this page predates v9:

- `stabilityCheck` is typed as `StabilityCheck`, a name that no longer exists in the codebase. The current type is `DevModeCheckFrequency`.
- `identityFunctionCheck` is missing, even though `Provider` accepts it and the hooks page shows it in use.
- The generic parameters read `<A extends Action = AnyAction, S = any>`, while the source declares `<A extends Action<string> = UnknownAction, S = unknown>`.

Docs-only change.

### DIFF
--- a/docs/api/Provider.md
+++ b/docs/api/Provider.md
@@ -21,7 +21,7 @@ The [Hooks](./hooks.md) and [`connect`](./connect.md) APIs can then access the p
 ### Props
 
 ```ts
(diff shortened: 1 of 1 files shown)


EskiMojo14 (approved):

EskiMojo14:
thanks!

Issue #1633, opened 2020-09-24
Docs: accessing store page useStore Hook Link seem to be broken


makuzaverite:
In docs when you try to go to http://react-redux.js.org/api/hooks.md#useStore nothing works it is 404 page 

this needs to be fixed

timdorr:
File is here: https://github.com/reduxjs/react-redux/blob/master/docs/using-react-redux/accessing-store.md

It appears to be the correct link, so I'm not sure what's going on there.

makuzaverite:
Ok got it

spencer741:
This issue is still persisting in a particular part of the docs.
https://react-redux.js.org/using-react-redux/accessing-store#using-the-usestore-hook

The link in that section is likely incorrect. Showing 404...
https://react-redux.js.org/api/hooks.md#useStore

markerikson:
This may be related to the whole lowercasing of anchors thing that DSv2 fixed at some point.

As I've said numerous times, the docs need to be updated to use DS v2, but I don't have time to do it myself right now.

PR #2372, opened 2026-08-05
docs: fix two broken internal doc links

dfedoryshchev:
`docs/` has exactly two broken internal links, and both are fixed here.

- `troubleshooting.md` pointed `<Provider>` at `#provider-store`, an anchor that exists on no page. It now points at the Provider API page.
- `using-react-redux/accessing-store.md` pointed at `../api/hooks.md#useStore`, but `## useStore()` generates the lowercase id `usestore`, so the link landed at the top of the Hooks page. Reported back in 2020 in #1633. The other link on that page already uses the lowercase form.

Docs-only change.


markerikson:
Thanks!

PR #2389, opened 2026-08-27
fix: respect falsy server snapshots and production context overrides

OskarEichler:
## Fixes

- Honor `Provider serverState={0}`, `false`, `''`, and `null` instead of silently using the client store's state during server rendering and hydration.
- Honor a connected component's `context` prop in production. Currently the context assignment is inside the development-only validation block: production either throws when only a custom provider exists or silently selects the default provider's store when both providers exist.

## Implementation

Use an explicit undefined check for the optional server snapshot, and move the custom-context assignment outside the development validation guard. The validation itself remains development-only. The diff changes three lines across two files; subscription and selector logic are unchanged.

## Verification

- Existing upstream suite: 186 tests across 20 files, with no type errors.
- ESLint, standalone type tests, full package build matrix, targeted Prettier, and `git diff --check` pass.
- Inline checks against built development and production packages: 38 scenarios covering hooks and connect, falsy/truthy/omitted snapshots, hydration without recoverable errors, post-hydration store updates, custom-only and nested providers, and changing the context prop on a mounted component.
- Before the change, all four falsy snapshots used client state; production custom-only context threw and nested contexts read the wrong store.
- No test files were added or modified. React Doctor flags the existing `useSyncExternalStore` call inside try/catch; this patch does not alter that error-handling path. Full CI OS/React-version matrices and device tests were not run locally.

## Compatibility / breaking changes

No public API, types, dependencies, or supported-platform changes. Explicit falsy server snapshots now take effect as documented; production context overrides now match development behavior. Apps relying on either previous bug will see the corrected state selection. Omitted/undefined snapshots still fall back to `store.getState`.

### DIFF
--- a/test/components/connect.spec.tsx
+++ b/test/components/connect.spec.tsx
@@ -2236,7 +2236,7 @@ describe('React', () => {
         expect(actualState).toEqual(expectedState)
       })
 
-      it.each(['development', 'production'])(
+      it.for(['development', 'production'] as const)(
         'should use a custom context prop in %s',
         (nodeEnv) => {
           vi.stubEnv('NODE_ENV', nodeEnv)
--- a/test/components/connect.spec.tsx
+++ b/test/components/connect.spec.tsx
@@ -2236,41 +2236,46 @@ describe('React', () => {
         expect(actualState).toEqual(expectedState)
       })
 
-      it('should use a custom context provider and consumer if passed as a prop to the component', () => {
-        class Container extends Component {
-          render() {
-            return <Passthrough />
+      it.each(['development', 'production'])(
+        'should use a custom context prop in %s',
+        (nodeEnv) => {
+          vi.stubEnv('NODE_ENV', nodeEnv)
+          class Container extends Component {
+            render() {
+              return <Passthrough />
+            }
           }
-        }
 
-        const context = React.createContext<ReactReduxContextValue<
-          any,
-          AnyAction
-        > | null>(null)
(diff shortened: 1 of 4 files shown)


timdorr:
Can you add some tests to both verify the behavior and ensure this doesn't regress in the future?

test/components/connect.spec.tsx
```diff
@@ -2236,41 +2236,46 @@ describe('React', () => {
         expect(actualState).toEqual(expectedState)
       })
 
-      it('should use a custom context provider and consumer if passed as a prop to the component', () => {
-        class Container extends Component {
-          render() {
-            return <Passthrough />
+      it.each(['development', 'production'])(
```
aryaemami59:
```suggestion
      it.for(['development', 'production'] as const)(
```

test/integration/ssr.spec.tsx
```diff
@@ -124,6 +124,25 @@ describe('New v8 serverState behavior', () => {
     vi.restoreAllMocks()
   })
 
+  it.each([0, false, '', null])(
+    'uses a falsy serverState value of %p',
```
aryaemami59:
```suggestion
  it.for([0, false, '', null] as const)(
    'uses a falsy serverState value of %s',
```

OskarEichler:
Applied both suggestions: the production-mode context test and falsy serverState test now use Vitest it.for parameterization. The current check suite is green.

